Issue #1, opened 2017-10-20
Repo setup


mattklein123:
Please use this issue to discuss general setup. I want to agree on a bunch of things before anyone starts pushing code:

- [x]  CI setup. (CircleCI)
- [x]  Build system. Bazel? 
- [x]  Dependency tracking. Glide/godep/something else?
- [x]  Outline of project scope. What this is and what this isn't in README.

mattklein123:
@kyessenov @junr03 @rshriram 

@rodaine @twoism can you please advise/help with Go project setup.

cc @envoyproxy/maintainers

kyessenov:
From Istio experience: we went with Bazel for go for build & dependency management, but it's rough, and everyone ends up using regular go tools anyways. However, it's fairly easy to convert Bazel workspace into a full-vendorized flat go build by symlinking stuff into `vendor`. I think we want Bazel for protobuf code generation, and we can either commit the generated code or not, based on the preferences.

htuch:
Please setup with https://github.com/envoyproxy/data-plane-api/pull/201 and native `proto_library` in mind.

rshriram:
I suggest that we stick with Circle CI.. ITs pretty sleek.
Second, for build, I suggest we stick with standard go tooling, if we want this repo to be used by wide variety of projects. Keep in mind that not everyone builds Envoy from scratch. Forcing them to install bazel just to convert this into a go project is a non starter for adoption.

We also want to be able to take advantage of go standard tooling in CircleCI.. So, my suggestion would be to commit the pb.go files as most projects do. This entire on the fly generation merely saves a few key strokes of a simple make command, but adds a tremendous overhead.

kyessenov:
Go tools are pretty nice, but dependency management is a zoo at the moment. `go dep` seems to be the future. Either way, I can maintain a Bazel build just to be able to import it elsewhere as Bazel infected lots of projects around Envoy already.

rshriram:
Dependency tracking: Glide/godep is fine. Given that @junr03 is the one with existing implementation of a v2 library, I suggest we go with whatever he is already using (which happens to be glide).

We can have a Bazel build, but lets not make it as the canonical build system. We can have scripts that ensure that the bazel files are updated always, so that its easier to import in other bazel projects.

9len:
Big +1 on standard Go tool chain, if you want broad adoption. 

[...]

PR #3, opened 2017-10-23
glide, makefile, and circleCI

rshriram:
Makefile with some useful stuff, glide.yaml to initialize the code base.

Signed-off-by: Shriram Rajagopalan <shriram@us.ibm.com>


mattklein123:
@rshriram FYI I'm going to make Circle a required check right now, pending the rest of the reviews you all are doing.

rshriram:
Yep thanks.

rshriram:
@kyessenov / @junr03 / @bndw PTAL. Glide setup, automatic version / SHA tagging, and circle ci setup. I also have some bits in there for proto generation using makefile, but its unused. I will leave that to @kyessenov as he is already working on bazel based pb.go generation.

rshriram:
completes all the tasks in #1

Makefile
```diff
@@ -0,0 +1,120 @@
...
+compile:
+	@go build -ldflags '$(LDFLAGS)' -o $(BINDIR)/$(OUTPUT_NAME)
+
+.PHONY: proto
+proto:
+	@go get -u github.com/golang/protobuf/{proto,protoc-gen-go}
+	@protoc --go_out=plugins=grpc:pkg/ vendor/github.com/envoyproxy/data-plane-api/api/*.proto \
+					--proto_path=vendor/github.com/envoyproxy/data-plane-api \
+					--proto_path=vendor/github.com/googleapis/googleapis/
+	@protoc --go_out=plugins=grpc:pkg/ vendor/github.com/envoyproxy/data-plane-api/api/filter/*.proto \
```
davecheney: I don't think this is right. When I run this command I get a file with imports like this
```
package envoy_api_v2_filter

import proto "github.com/golang/protobuf/proto"
import fmt "fmt"
import math "math"
import envoy_api_v21 "api"
import envoy_api_v22 "api"
import envoy_api_v24 "api"
```
These should be `github.com/envoyproxy/go-control-plane/pkg/envoy/api`

bndw: Looks like the same issue as https://github.com/golang/protobuf/issues/63

Makefile
```diff
@@ -0,0 +1,120 @@
+.DEFAULT_GOAL	:= compile
+
+#------------------------------------------------------------------------------
+# Variables
+#------------------------------------------------------------------------------
+
+SHELL 		:= /bin/bash
+BINDIR		:= bin
+BUILDDIR    := build
```
bndw: Nit -- indentation

Makefile
rshriram: Thanks for catching this @davecheney .. I think I will take this line out for now, since @kyessenov is generating protos using bazel (using same proto version as envoy repo), and our plan is to commit these protos into the repo, so that others can just import this library.

Makefile
rshriram: Its GH doing the weird thing.. IT looks correct on the terminal.

rshriram:
@bndw updated.

.circleci/config.yml
```diff
@@ -0,0 +1,12 @@
+version: 2
+jobs:
+  build:
+    working_directory: /go/src/github.com/envoyproxy/go-control-plane
+    docker:
+      - image: rshriram/envoy-go-build:956131ed
```
junr03: can we pin to a generic image?

.circleci/custom-image/install-deps.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+VERSION=3.4.0
```
junr03: Why are we setting this version var here? It's not used anywhere. Am I missing something?

Makefile
```diff
@@ -0,0 +1,126 @@
...
+DOCKERDIR	:= docker
+RELEASEDIR  := release
+OUTPUT_NAME := envoyctl
```
junr03: should this be `go-control-plane`?

Makefile
kyessenov: it's tabs

glide.lock
```diff
@@ -0,0 +1,31 @@
+hash: d6fc1ffda5336bc94cc564496911ac0f6d812481c65b06e2a7833116cf3d127c
+updated: 2017-10-24T09:32:55.826752182-04:00
+imports:
+- name: github.com/envoyproxy/data-plane-api
```
kyessenov: not needed (there's no code there)

glide.yaml
```diff
@@ -0,0 +1,25 @@
+package: github.com/envoyproxy/go-control-plane
+import:
+- package: golang.org/x/net
+  subpackages:
+  - context
+- package: github.com/envoyproxy/data-plane-api
```
kyessenov: yup, not needed in glide

Makefile
kyessenov: ref https://github.com/envoyproxy/data-plane-api/pull/204

.circleci/config.yml
rshriram: So, the standard go image from circle CI does not have protoc and glide installed. So, i had to create this. I thought that once we land this, we can (a) try to reuse lyft/envoy-build (with some improvements), or push this as a standard build image to dockerhub.

.circleci/custom-image/install-deps.sh
rshriram: Duh.. I hardcoded it everywhere else.. good catch

Makefile
rshriram: envoyctl is so cool! :( .. Either way, I thought that this would change rapidly, when we start putting in CLIs in cmd/ folder.. We need one per CLI tool.

glide.lock
rshriram: I just copied this from whatever @junr03 gave me. @junr03 any thoughts on this?

rshriram:
Yes. That way its simpler in this PR.

PR #4, opened 2017-10-24
Initial import of generated files

kyessenov:


kyessenov:
Waiting for the API PR to be merged...

bazel/update.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
rshriram: Just thinking.. does it make sense to fold this into the main envoy repo, and we can reuse the lyft/envoy-build bazel image? That image has all the stuff we need. Basically, we reduce this command down to one liner (docker run -it lyft/envoy-build -c ... )

kyessenov:
I can move them to top-level. Would need to rewrite the import for "api".
Is referring by `github.com/envoyproxy/go-control-plane/vendor/api` not good?

bazel/update.sh
kyessenov: This script just bulids the go stuff from API repo skipping cc and py code.

davecheney:
Bad things will happen when dev or glide try’s to “collapse” vendored dependencies. Can you just move it up one level to github.com/envoyproxy/go-control-plane/api .

kyessenov:
Done moving. The build passes since `api` is skipped over (and not pulled indirectly since there is no other code)

kyessenov:
I'll follow up with another update (the SHA references my PR, not master)

PR #11, opened 2017-10-30
Use nonce per xDS stream

kyessenov:
Nonce should be maintained per xDS stream within ADS stream.

### DIFF
--- a/pkg/grpc/server.go
+++ b/pkg/grpc/server.go
@@ -72,6 +72,11 @@ type watches struct {
 	clusters  cache.Watch
 	routes    cache.Watch
 	listeners cache.Watch
+
+	endpointNonce int64
+	clusterNonce  int64
+	routeNonce    int64
+	listenerNonce int64
 }
 
 // cancel all watches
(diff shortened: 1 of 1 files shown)


pkg/grpc/server.go
```diff
@@ -83,77 +88,89 @@ func (values watches) cancel() {
...
+	var values watches
+	defer func() {
+		values.cancel()
+	}()
 
 	// sends a response by serializing to protobuf Any
-	send := func(resp cache.Response, typeURL string) error {
+	send := func(resp cache.Response, typeURL string) (int64, error) {
 		resources := make([]*any.Any, len(resp.Resources))
+		streamNonce = streamNonce + 1
```
rshriram: for clarity, I suggest prefixing nonce with the type. Like endpoint-x, cluster-y, etc.

rshriram: bit more work, but easier to debug

kyessenov: This is the nonce generator, to make sure nonces are not shared between responses. Are you suggesting having four generators that don't overlap?

Issue #7, opened 2017-10-25
Caching xDS responses


kyessenov:
The management server has to maintain a cache of the desired configurations for all proxies in the mesh. Since there could be a lot more proxies that different configurations, it seems we want to be able to group proxy nodes into buckets.

Let's start with a generic bucketing function:
```go
function nodeGroup(proxy api.Node): string
```

For example, we can use the cluster field in the node to group API responses if the cluster corresponds to the service cluster. Alternatively, we can use the generic metadata field in the node message to assign configuration to proxies (e.g. in kubernetes, we may use pod labels to uniequely identify sidecars).

The cache is a map from the node group and response type to the desired response:
```go
var cache map[NodeGroup][ResponseType]proto.Message
```

mattklein123:
For Lyft I think we basically need a bucket for a group, and then a bucket for canary also (which is a node concept). @junr03 can you work with @kyessenov to make sure we can accomplish this in the first version.

PR #12, opened 2017-10-31
Simple cache implementation

kyessenov:
Implement a cache of responses with a callback function.
The cache is protected by a single lock, keeps track of watches and versions.
The cache responds to open watches on every update, and increments a version per request type.

### DIFF
--- a/pkg/cache/simple.go
+++ b/pkg/cache/simple.go
@@ -14,25 +14,151 @@
 
 package cache
 
-import "github.com/golang/protobuf/proto"
+import (
+	"fmt"
+	"sync"
+
+	"github.com/envoyproxy/go-control-plane/api"
+	"github.com/golang/protobuf/proto"
+)
 
 // SimpleCache is a snapshot-based cache that maintains a single version per
-// xDS response, node group tuple.
+// xDS response, node group tuple, with no canary updates.
 type SimpleCache struct {
-	responses map[CacheKey][ResponseType]proto.Message
-	versions  map[CacheKey][ResponseType]int
+	// responses are cached resources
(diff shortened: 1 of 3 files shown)


kyessenov:
contributes to #7

pkg/cache/simple.go
```diff
@@ -0,0 +1,164 @@
...
+	// callback requests missing responses
+	callback func(Key, ResponseType)
+
+	// watchCount is the ID generator for watches
+	watchCount int64
+
+	// groups is the hashing function for proxy nodes
+	groups NodeGroup
+
+	mu sync.Mutex
```
rshriram: can we shift to a RW lock? a bit more scalable, given that we have 90% reads by a lot of clients.

pkg/cache/simple.go
```diff
@@ -0,0 +1,164 @@
...
+			for _, watch := range cache.watches[group][typ] {
+				watch.Value <- Response{
+					Version:   fmt.Sprintf("%d", cache.versions[group][typ]),
+					Resources: resources,
+				}
+				// remove clean-up as the watch is discarded immediately
+				watch.stop = nil
+			}
+			// discard all watches; the client must request a new watch to receive updates after ACK/NACK
+			cache.watches[group][typ] = make(map[int64]Watch)
```
rshriram: It would be good to have stats, as we have in Istio. At some point, this will enable us to profile as well.
This might be too much to ask. So feel free to defer it for follow up PR.

rshriram:
per offline conversation, we can move to finer grained locks once we have this working e2e.

PR #15, opened 2017-10-31
Update snapshot cache and validate manually

kyessenov:
Working manual test of Envoy consuming ADS

Instructions to run:
```bash
term1> go run pkg/test/main/main.go
term2> envoy -c pkg/test/main/server_ads.yaml  --service-cluster "service-cluster" --service-node "service-node"
term3> while sleep 1; do curl localhost:9000/test; done
```

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -93,6 +93,21 @@ const (
 	ListenerResponse
 )
 
+func (typ ResponseType) String() string {
+	switch typ {
+	case EndpointResponse:
+		return "endpoints"
+	case ClusterResponse:
+		return "clusters"
+	case RouteResponse:
+		return "routes"
+	case ListenerResponse:
+		return "listeners"
+	default:
+		return "unknown"
+	}
+}
+
 // Key is the node group identifier
 type Key string
 
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -62,9 +62,9 @@ type Watch struct {
 	// Type is the response type.
 	Type ResponseType
 
-	// Names for monitored resources. LDS/CDS expect empty list since they are
-	// discovered globally. RDS/EDS expect explicitly enumerated resource names
-	// as dictated by LDS/CDS.
+	// Names of requested resources (or empty to mean all resources). Resources
+	// that are not explicitly mentioned in this field are ignored by the proxy
+	// node.
 	Names []string
 
 	stop func()
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -57,6 +57,11 @@ type Watch struct {
 	// close the corresponding stream.
 	Value chan Response
 
+	// Names for monitored resources. LDS/CDS expect empty list since they are
+	// discovered globally. RDS/EDS expect explicitly enumerated resource names
+	// as dictated by LDS/CDS.
+	Names []string
+
 	stop func()
 }
 
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -43,8 +43,10 @@ type ConfigWatcher interface {
 type Cache interface {
 	ConfigWatcher
 
-	// SetResource sets a response for a node group and a type.
-	SetResource(Key, ResponseType, []proto.Message)
+	// SetSnapshot sets a response snapshot for a node group. The snapshots
+	// should have distinct versions and be internally consistent (e.g. all
+	// referenced resources must be included in the snapshot).
+	SetSnapshot(Key, Snapshot) error
 }
 
 // Watch is a dedicated stream of configuration resources produced by the
@@ -56,6 +58,9 @@ type Watch struct {
 	// unrecoverable error has occurred in the producer, and the consumer should
 	// close the corresponding stream.
 	Value chan Response
+
+	// Type is the response type.
+	Type ResponseType
 
 	// Names for monitored resources. LDS/CDS expect empty list since they are
 	// discovered globally. RDS/EDS expect explicitly enumerated resource names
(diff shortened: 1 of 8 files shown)


pkg/cache/simple.go
```diff
@@ -46,119 +49,169 @@ type SimpleCache struct {
 	mu sync.Mutex
 }
 
-// MakeSimpleCache initializes a simple cache.
+// Snapshot is an internally consistent snapshot of xDS resources.
+// Snapshots should have distinct versions per node group.
```
rshriram: Instead of saying per node group, why not just say per service cluster?

pkg/test/main/server_ads.yaml
```diff
@@ -0,0 +1,23 @@
...
+        address: 127.0.0.1
+        port_value: 18000
+    lb_policy: ROUND_ROBIN
+    http2_protocol_options: {}
+admin:
+  access_log_path: /dev/null
+  address:
+    socket_address:
+      address: 127.0.0.1
+      port_value: 19000
```
rshriram: Where is the bit to start Envoy? If it’s manual can we just script those steps and dump script here? Helps others reproduce the test.

pkg/cache/cache.go
```diff
@@ -57,6 +59,14 @@ type Watch struct {
 	// close the corresponding stream.
 	Value chan Response
 
+	// Type is the response type.
+	Type ResponseType
+
+	// Names for monitored resources. LDS/CDS expect empty list since they are
+	// discovered globally. RDS/EDS expect explicitly enumerated resource names
+	// as dictated by LDS/CDS.
```
rshriram: Don’t understand. More explanation..

pkg/cache/cache.go
```diff
@@ -82,23 +92,12 @@ type Response struct {
 	Canary bool
```
rshriram: I think we can get rid of this until we have that feature implemented

pkg/cache/simple.go
```diff
@@ -46,119 +49,169 @@ type SimpleCache struct {
...
-		cache.versions[group] = make(map[ResponseType]int)
+	cache.snapshots[group] = snapshot
+
+	// trigger existing watches
+	if _, ok := cache.watches[group]; ok {
+		for _, watch := range cache.watches[group] {
+			respond(watch, snapshot, group)
+		}
+		// discard all watches; the client must request a new watch to receive updates and ACK/NACK
+		delete(cache.watches, group)
```
rshriram: Why trigger and then delete ?

pkg/cache/cache.go
kyessenov: See the discussion on slack...

pkg/cache/cache.go
kyessenov: This was explicitly requested by @junr03, so making provision to implement it.

pkg/cache/simple.go
kyessenov: Node is not just a service cluster. There are labels and metadata inside. We may want to group by a combination of a cluster and envoy build for example.

pkg/cache/simple.go
kyessenov: A trigger should make envoy request a new watch, no need to wait for cancellation.

pkg/test/main/server_ads.yaml
kyessenov: see PR description.

pkg/cache/cache.go
rshriram: Could you summarize it and dump it here ?

kyessenov: It's pretty much what comment says. LDS/CDS request all resources. EDS/RDS request resources by name. We only reply to requested names since the rest gets ignored.

rshriram: And put this explanation above (which is clearer) in the comment :)

pkg/cache/cache.go
rshriram: They are not using the discovery canary field. They are using EDS canary field, which is already covered by the fact that we are using the eds proto. cc @junr03

junr03: Sorry @kyessenov I don't remember talking to you about this? We don't currently use this field in the response.

kyessenov: Awesome, then I'll drop it for now. I remember on slack that RDS + canary is something that we want to support.

pkg/cache/cache.go
kyessenov: Done

pkg/cache/resource.go
```diff
@@ -0,0 +1,78 @@
...
+	case *api.ClusterLoadAssignment:
+		return v.GetClusterName()
+	case *api.Cluster:
+		return v.GetName()
+	case *api.RouteConfiguration:
+		return v.GetName()
+	case *api.Listener:
+		return v.GetName()
+	default:
+		return ""
```
bndw: Should this return "unknown" for consistency with `ResponseType.String()`?

pkg/cache/simple.go
```diff
@@ -46,119 +49,169 @@ type SimpleCache struct {
...
 	defer cache.mu.Unlock()
 
 	// update the existing entry
-	if _, ok := cache.responses[group]; !ok {
-		cache.responses[group] = make(map[ResponseType][]proto.Message)
-		cache.versions[group] = make(map[ResponseType]int)
+	cache.snapshots[group] = snapshot
+
+	// trigger existing watches
+	if _, ok := cache.watches[group]; ok {
```
bndw:
```go
if watches, ok := cache.watches[group]; ok {
    for _, watch := range watches {
    }
}
```

pkg/cache/simple.go
```diff
@@ -46,119 +49,169 @@ type SimpleCache struct {
...
-	cache.responses[group][typ] = resources
+	return nil
+}
 
-	// increment version
-	cache.versions[group][typ]++
+// Respond to a watch with the snapshot value. The value channel should have capacity not to block.
+// TODO(kuat) this responds immediately and can cause the remote node to spin if it consistently fails to ACK the update
+func respond(watch Watch, snapshot Snapshot, group Key) {
+	typ := watch.Type
```
bndw: nit: I'm a fan of grouping declaration/assignments, but not sure if this is recommended. cc: @davecheney 

```go
var (
    typ        = watch.Type
    resources  = snapshot.resources[typ]
    version    = snapshot.version
)
```

bndw:
Looks like I was too slow 😛

pkg/cache/simple.go
kyessenov: thanks will follow up with this.

pkg/cache/resource.go
kyessenov: an empty name is an invalid name (most likely) but a name `unknown` could be a legitimate resource name.

pkg/cache/simple.go
kyessenov: let me know what the stylistic recommendation is :)

PR #18, opened 2017-11-10
Unit tests

kyessenov:
Added unit tests with little functional change. Some code shuffling for testing purposes.
```
ok      github.com/envoyproxy/go-control-plane/pkg/cache        0.257s  coverage: 98.5% of statements
ok      github.com/envoyproxy/go-control-plane/pkg/server       0.009s  coverage: 92.9% of statements
?       github.com/envoyproxy/go-control-plane/pkg/test [no test files]
?       github.com/envoyproxy/go-control-plane/pkg/test/main    [no test files]
?       github.com/envoyproxy/go-control-plane/pkg/test/resource        [no test files]
ok      github.com/envoyproxy/go-control-plane/pkg/util 0.007s  coverage: 77.8% of statements
?       github.com/envoyproxy/go-control-plane/pkg/version      [no test files]
```

Modified `Cache` interface to have a singular `Watch` declaration instead of multiple `WatchEndpoints` and so on. The implementation is likely to share code between the different stubs, so this should reduce clutter.


Makefile
```diff
@@ -29,18 +29,9 @@ GOPKGS		= $(shell go list ./... \
...
 
-# linker flags to set build info variables
-BUILD_SYM	:= github.com/envoyproxy/go-control-plane/pkg/version
-LDFLAGS		+= -X $(BUILD_SYM).version=$(APP_VER)
-LDFLAGS		+= -X $(BUILD_SYM).gitRevision=$(shell git rev-parse --short HEAD 2> /dev/null  || echo unknown)
-LDFLAGS		+= -X $(BUILD_SYM).branch=$(shell git rev-parse --abbrev-ref HEAD 2> /dev/null  || echo unknown)
-LDFLAGS		+= -X $(BUILD_SYM).buildUser=$(shell whoami || echo nobody)@$(shell hostname -f || echo builder)
-LDFLAGS		+= -X $(BUILD_SYM).buildDate=$(shell date +%Y-%m-%dT%H:%M:%S%:z)
-LDFLAGS		+= -X $(BUILD_SYM).goVersion=$(word 3,$(shell go version))
-
```
rshriram: why are you removing this? This is needed to build the pkg/version that allows us to see the SHA associated with the binary.

rshriram:
I cannot comment on the Go-ey ness of the code. It looks sensible. Please undo the changes related to the version related stuff in Makefile. They are required to embed the git commit sha in the binary (built by the version package). Thats the reason I included the version package as well in main.go

Makefile
kyessenov: This is entirely unnecessary in a library. There is no main function, why do we need ldflags?

kyessenov: We can add those, but if this is to be used as part of another project, LD flags make little sense. We can them back with the CLI.

kyessenov:
cc @zackbutcher for some go-ey ness.

Makefile
rshriram: in that case, fine.

kyessenov: CLIs take a lot of time. And noone likes yet another CLI. It's easy to build a main with this library, where LD flags can be added.

9len: 👍 Turbine Labs will consume this as a library, and we have our won way of managing injection of this sort of thing. Thanks!

PR #27, opened 2018-01-03
Fix loop over resources in cache:simple:respond

hsmade:
This would break when having, for example, multiple clusters and requesting ClusterLoadAssignment for the second in resources.

I used this repo to build an XDS that uses consul as backend (opinionated towards our in-house setup). I have 114 services, and envoy only has endpoints for one. This fixes that.


bndw:
Hi @hsmade, could you please sign the DCO? There's more information available [here](https://github.com/probot/dco#how-it-works).

PR #32, opened 2018-01-10
Support multiple resources in cache:simple

hsmade:
This would break when having, for example, multiple clusters and requesting ClusterLoadAssignment for the second in resources.

I used this repo to build an XDS that uses consul as backend (opinionated towards our in-house setup). I have 114 services, and envoy only has endpoints for one. This fixes that.

### DIFF
--- a/pkg/cache/simple.go
+++ b/pkg/cache/simple.go
@@ -133,13 +133,18 @@ func respond(watch Watch, snapshot Snapshot, group Key) {
 		}
 
 		// check that every snapshot resource name is present in the request
+		found := false
 		for _, resource := range resources {
 			resourceName := GetResourceName(resource)
-			if _, exists := names[resourceName]; !exists {
-				glog.V(10).Infof("not responding for %s from %q at %q since %q not requested %v",
-					typ.String(), group, version, resourceName, watch.Names)
-				return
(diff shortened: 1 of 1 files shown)


pkg/cache/simple.go
```diff
@@ -133,14 +133,19 @@ func respond(watch Watch, snapshot Snapshot, group Key) {
...
+		found := false
 		for _, resource := range resources {
 			resourceName := GetResourceName(resource)
-			if _, exists := names[resourceName]; !exists {
-				glog.V(10).Infof("not responding for %s from %q at %q since %q not requested %v",
-					typ.String(), group, version, resourceName, watch.Names)
-				return
+			if _, exists := names[resourceName]; exists {
+				found = true
+				break
```
kyessenov: This seems to check that there exists at least on resource in the snapshot that was requested. For example,
if the snapshot contains resources `a` and `b`; and the request asks for `a` and `c` then we return `a` and `b`.
Is this intentional?

kyessenov:
I have an example usage of this library for k8s here: https://github.com/kyessenov/envoymesh/blob/master/envoy/generator.go#L116.
It's rather inefficient, but it seems to be logically correct.

kyessenov:
This should be superseded by #43

PR #35, opened 2018-01-17
Update makefile

kyessenov:
In preparation of protoc-gen-validate support.
There are issues with PGV that need to be addressed first:
- `api/auth.proto` is in a distinct package that prevents PGV from working at the moment.
- gogo-optimized annotations are not understood by PGV.


generate_protos.log
```diff
@@ -12,11 +10,11 @@ Date:   Tue Jan 16 13:30:08 2018 -0800
...
-Generating gogofast protos api/filter ...
-Generating gogofast protos api/filter/accesslog ...
-Generating gogofast protos api/filter/http ...
-Generating gogofast protos api/filter/network ...
+Generating protos api ...
+Generating protos api/auth ...
+Generating protos api/filter ...
+Generating protos api/filter/accesslog ...
+Generating protos api/filter/http ...
+Generating protos api/filter/network ...
```
rshriram: Why do we commit this log file?

kyessenov: To record the version of data-plane-api and versions of the tools.

PR #36, opened 2018-01-22
Fix gogo-genproto dependency

rshriram:
use istio.io/gogo-genproto instead of github.com/istio/gogo-genproto, as the current form is preventing the use of this library in istio code base.


rshriram:
@9len I believe it should work in the same way (istio or non istio), if you are using dep/glide. We have made sure that istio.io/gogo-genproto is accessible for all the go tools (all our go packages are rooted at istio.io). @douglas-reid might be able to provide more go expertise here.

kyessenov:
Thanks for fixing this. Did not realize the vanity URL was mandatory :)

kyessenov:
@9len there is no dependency on istio, we use `istio.io` as a placeholder for Google APIs (which are brought in by gRPC services) until the official googleapis packages support gogo (per @wora).

gnirodi:
Thanks guys for pushing this through.

9len:
Thanks. Misread the diff, I understand now.

PR #39, opened 2018-02-01
Upgrade data-plane-api

kyessenov:
Updates data-plane-api to take in package re-organization in data-plane-api.

_IMPORTANT_: this is a breaking change since package paths have changed, and hopefully stabilized. I have verified that the latest Envoy should work end-to-end with this change. It should be an easy update to fix the clients since the proto content is stable, and only import paths need to be modified.

Requires the following PRs to be merged:
- https://github.com/envoyproxy/data-plane-api/pull/452
- https://github.com/envoyproxy/envoy/pull/2495

Minor improvements to code generation using gogoproto annotations.

cc @davecheney @crazytan @bndw @joyeb @junr03 @rshriram  (any other folks that I missed)


build/generate_protos.sh
```diff
@@ -2,33 +2,22 @@
 set -o errexit
 set -o nounset
 set -o pipefail
+shopt -s nullglob
 
-root="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+root="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )/.."
+xds=${root}/vendor/github.com/envoyproxy/data-plane-api
```
crazytan: perhaps you forgot to add data-plane-api as a dependency to glide.yaml?

kyessenov:
Yes, there's no functional change. Only alignment with data-plane-api packaging hierarchy.

PR #43, opened 2018-02-10
Implement ADS/xDS/REST server modes; integration tests

kyessenov:
- [x] add code coverage reporting
- [x] get rid of glog dependency
- [x] implement snapshot consistency check
- [x] simplify some types around the cache (request/response/nodegroup)
- [x] add server callbacks
- [x] implement xDS mode by optionally replying for subsets
- [x] implement REST mode using a custom gRPC gateway
- [x] add a docker-based integration against master Envoy: the test driver can scale number of listeners/routes/clusters independently and frequency of the checks
- [x] try out access log service in the integration test
- [x] separate snapshot versions per resource type 
- [x] add status info per node


pkg/cache/snapshot.go
```diff
@@ -0,0 +1,297 @@
...
+	log log.Logger
+
+	// snapshots are cached resources
+	snapshots map[string]Snapshot
+
+	// status information for all nodes
+	status map[string]StatusInfo
+
+	// callback requests missing responses
+	callback func(string)
```
9len: is this used anymore? It seems like you need it in `CreateWatch` to tell whatever the snapshot provider is that you need a snapshot for a given key.

(also, I think you need some way to signal to the snapshot provider that you no longer need the key, or it will keep providing snapshots forever)

pkg/test/main/main.go
```diff
@@ -16,80 +16,205 @@
...
+func (cb *callbacks) OnStreamClosed(id int64) {
+	log.Debugf("stream %d closed", id)
+}
+func (cb *callbacks) OnStreamRequest(int64, *v2.DiscoveryRequest) {
+	if cb.signal != nil {
+		close(cb.signal)
+		cb.signal = nil
+	}
+}
+func (cb *callbacks) OnStreamResponse(int64, *v2.DiscoveryResponse) {}
```
9len: Does it makes sense to pass the `Node` here? It can be obtained from the `DiscoveryRequest` in `OnStreamRequest` and `OnFetch`, but without it we have to keep a `map[int64]Node`, which is cumbersome.

pkg/cache/snapshot.go
9len: ah, I see. thanks.

pkg/test/main/main.go
kyessenov: Yes, it does make sense. This is a quirk of the protocol - typically it's the same node on the stream, but each request comes with a copy of the node.

kyessenov:
cc @joeyb @davecheney - this version should work with non-ADS and basic REST fetch method.

cetanu:
Are you building this on a Linux OS? When building it on a OSX, the bin/test file has the wrong format. Any chance of building that file inside the container?

kyessenov:
Thanks @cetanu -- added `$(BINDIR)/test-linux` target specifically for docker. The regular binary test should run fine on Mac with Mac-built envoy.

kyessenov:
Split into three PRs

PR #49, opened 2018-03-02
gRPC gateway for Fetch

kyessenov:

### DIFF
--- a/pkg/server/gateway.go
+++ b/pkg/server/gateway.go
@@ -1,0 +1,96 @@
+// Copyright 2018 Envoyproxy Authors
+//
+//   Licensed under the Apache License, Version 2.0 (the "License");
+//   you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


pkg/server/gateway.go
```diff
@@ -0,0 +1,96 @@
...
+	default:
+		http.Error(resp, "no endpoint", http.StatusNotFound)
+		return
+	}
+
+	if req.Body == nil {
+		http.Error(resp, "empty body", http.StatusBadRequest)
+		return
+	}
+
```
bndw: Do we need to `defer req.Body.Close()` here?

kyessenov: Seems like not necessary (from godoc):
```go
        // For server requests the Request Body is always non-nil
        // but will return EOF immediately when no body is present.
        // The Server will close the request body. The ServeHTTP
        // Handler does not need to.
```

PR #61, opened 2018-03-15
Update data-plane-api

kyessenov:
- introduce a few well-known names for filters


pkg/cache/resource.go
```diff
@@ -30,10 +30,51 @@ type Resource interface {
...
+
+	// Echo network filter
+	Echo = "envoy.echo"
+	// HTTPConnectionManager network filter
 	HTTPConnectionManager = "envoy.http_connection_manager"
-	TCPProxy              = "envoy.tcp_proxy"
+	// TCPProxy network filter
+	TCPProxy = "envoy.tcp_proxy"
+	// RateLimit network filter
+	RateLimit = "envoy.ratelimit"
```
kyessenov: Can you please double-check this list?

pkg/cache/resource.go
```diff
@@ -30,10 +30,51 @@ type Resource interface {
 
 // Common names for Envoy filters.
 const (
-	CORS                  = "envoy.cors"
-	Router                = "envoy.router"
+	// Buffer HTTP filter
+	Buffer = "envoy.buffer"
```
9len: is `cache` the right package for these?

kyessenov: It's not ideal. We have "cache", "server", "util", "test", "log" packages. Which one would be better? Should we create a new one "filter", maybe?

kyessenov: cc @ostromart since he recently touched this code

9len: I would say either util or filter.

kyessenov: In the interest of package economy, I'll move these well-known names to util.

PR #74, opened 2018-04-24
fix golint and HTTPGateway logging

dunjut:
Roughly logging without testing errors is annoying, so here is a minor fix on the `HTTPGateway`'s response log.


Makefile
```diff
@@ -140,7 +140,9 @@ tools.govet:
 tools.golint:
 	@command -v golint >/dev/null ; if [ $$? -ne 0 ]; then \
 		echo "--> installing golint"; \
-		go get github.com/golang/lint/golint; \
+		mkdir -p $$GOPATH/src/golang.org/x \
```
taion809: did you mean to use `$$`?

PR #80, opened 2018-06-04
Close request.Body in Gateway

pims:
Closes request.Body to avoid memory leak.


kyessenov:
I think ServeHTTP does not have to close the body. See the request documentation:
```
        // For client requests a nil body means the request has no
        // body, such as a GET request. The HTTP Client's Transport
        // is responsible for calling the Close method.
        //
        // For server requests the Request Body is always non-nil
        // but will return EOF immediately when no body is present.
        // The Server will close the request body. The ServeHTTP
        // Handler does not need to.
```

pims:
Thanks for pointing this out, I learned something today. I’ll close this PR as it’s no longer relevant.

PR #111, opened 2018-09-06
Adds access to context in the server callbacks

wozz:
This change adds the stream/request context to the server callbacks.  This allows functionality such as an authentication middleware that extracts an identity and adds it to the context to have authorization checks in the server callbacks.  The context can be checked during the OnStreamOpen or OnFetchRequest callbacks.  If for some reason the request should not be allowed to proceed, then the callback can return an error to end processing.

Since this is a breaking change for consumers of this package as a library, the version number should be bumped.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -48,6 +48,7 @@ type Server interface {
 // The callbacks are invoked synchronously.
 type Callbacks interface {
 	// OnStreamOpen is called once an xDS stream is open with a stream ID and the type URL (or "" for ADS).
+	// Returning an error will end processing and close the stream. OnStreamClosed will still be called.
 	OnStreamOpen(context.Context, int64, string) error
 	// OnStreamClosed is called immediately prior to closing an xDS stream with a stream ID.
 	OnStreamClosed(int64)
@@ -55,7 +56,8 @@ type Callbacks interface {
 	OnStreamRequest(int64, *v2.DiscoveryRequest)
 	// OnStreamResponse is called immediately prior to sending a response on a stream.
 	OnStreamResponse(int64, *v2.DiscoveryRequest, *v2.DiscoveryResponse)
-	// OnFetchRequest is called for each Fetch request
+	// OnFetchRequest is called for each Fetch request. Returning an error will end processing of the
+	// request and respond with an error.
 	OnFetchRequest(context.Context, *v2.DiscoveryRequest) error
 	// OnFetchResponse is called immediately prior to sending a response.
 	OnFetchResponse(*v2.DiscoveryRequest, *v2.DiscoveryResponse)
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/gogo/protobuf/proto"
 	"github.com/gogo/protobuf/types"
+	"google.golang.org/grpc"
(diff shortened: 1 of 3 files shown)


pkg/server/server.go
```diff
@@ -170,7 +173,9 @@ func (s *server) process(stream stream, reqCh <-chan *v2.DiscoveryRequest, defau
 	}
 
 	if s.callbacks != nil {
-		s.callbacks.OnStreamOpen(streamID, defaultTypeURL)
+		if err := s.callbacks.OnStreamOpen(stream.Context(), streamID, defaultTypeURL); err != nil {
+			return err
```
kyessenov: I think it would be good to add a test case for the error condition (to maintain coverage).

pkg/server/server.go
```diff
@@ -47,20 +48,20 @@ type Server interface {
 // The callbacks are invoked synchronously.
 type Callbacks interface {
 	// OnStreamOpen is called once an xDS stream is open with a stream ID and the type URL (or "" for ADS).
-	OnStreamOpen(int64, string)
+	OnStreamOpen(context.Context, int64, string) error
```
9len: Could do this in a backwards-compatible way by calling this `OnStreamOpenWithContext`, (and `OnFetchRequestWithContext` below). That said, this is a pretty easy change to take, so probably not worth it.

pkg/server/server.go
```diff
@@ -47,20 +48,20 @@ type Server interface {
...
 	// OnStreamClosed is called immediately prior to closing an xDS stream with a stream ID.
 	OnStreamClosed(int64)
 	// OnStreamRequest is called once a request is received on a stream.
 	OnStreamRequest(int64, *v2.DiscoveryRequest)
 	// OnStreamResponse is called immediately prior to sending a response on a stream.
 	OnStreamResponse(int64, *v2.DiscoveryRequest, *v2.DiscoveryResponse)
 	// OnFetchRequest is called for each Fetch request
-	OnFetchRequest(*v2.DiscoveryRequest)
+	OnFetchRequest(context.Context, *v2.DiscoveryRequest) error
```
9len: It's a little awkward that two of these functions now return errors, but the rest do not. In particular, if `OnFetchRequest` can return errors, `OnStreamRequest` is likely to encounter the same errors. Does it make sense to convert all of these to return errors?

wozz: I was thinking about this just as a way to do auth/pre-processing for a request.  Would it make more sense to add a new Fetch callback like `OnFetchBegin` to separate it and keep the semantics more similar between stream/fetch handling?

I believe any other errors could already be handled by the cache Fetch or Watcher functions.  If there's a good use case for each of the callbacks to return errors, I can add them though.

pkg/server/server.go
wozz: Any change to the Callbacks interface would require an update. So even adding new functions like this would require implementations to add new no-op functions, which seems about the same amount of effort.

pkg/server/server.go
9len: ah that makes sense. I think it's fine as-is, but it's worth documenting why these two functions return an error (eg document explicitly that these would be where you'd stick auth or connection pre-processing).

A separate callback might also make sense, but leaving it this way allows for different behaviors on fetch vs stream open, which might better for some implementations.

Thanks for the quick response!

Issue #114, opened 2018-09-17
Integration test does not wait before closing HTTP Servers


klarose:
I want to make some changes to go-control-plane locally. Prior to making any modifications, I decided to run the integration tests before doing anything else. Surprisingly, they failed:
```
make integration.docker 
--> building Linux test binary
--> building test docker image
Sending build context to Docker daemon  182.3MB
Step 1/5 : FROM envoyproxy/envoy:latest
 ---> cf2442a79328
Step 2/5 : ADD sample /sample
 ---> Using cache
 ---> 901328cc36a6
Step 3/5 : ADD build/integration.sh build/integration.sh
 ---> Using cache
 ---> e9f5d2931ab0
Step 4/5 : ADD bin/test-linux /bin/test
 ---> Using cache
 ---> 21abea4c5a7c
Step 5/5 : ENTRYPOINT build/integration.sh
 ---> Using cache
 ---> c430b556d974
Successfully built c430b556d974
Successfully tagged test:latest
docker run -it -e "XDS=ads" test -debug
INFO[0000] upstream listening HTTP/1.1                   port=18080
INFO[0000] waiting for the first request...             
INFO[0000] gateway listening HTTP/1.1                    port=18001
INFO[0000] access log server listening                   port=18090
INFO[0000] management server listening                   port=18000
ERRO[0000] http: Server closed                          
ERRO[0000] http: Server closed                          
DEBU[0000] stream 1 open for                            
DEBU[0000] open watch 1 for type.googleapis.com/envoy.api.v2.Cluster[] from nodeID "test-id", version "" 
INFO[0000] initial snapshot {Xds:ads Version: UpstreamPort:18080 BasePort:9000 NumClusters:4 NumHTTPListeners:2 NumTCPListeners:2} 
INFO[0000] executing sequence                            requests=5 updates=3
INFO[0000] update snapshot                               version=v0
DEBU[0000] respond open watch 1[] with new version "v0" 
DEBU[0000] respond type.googleapis.com/envoy.api.v2.Cluster[] version "" with version "v0" 
DEBU[0000] respond type.googleapis.com/envoy.api.v2.ClusterLoadAssignment[cluster-v0-3 cluster-v0-2 cluster-v0-1 cluster-v0-0] version "" with version "v0" 
DEBU[0000] open watch 2 for type.googleapis.com/envoy.api.v2.Cluster[] from nodeID "test-id", version "v0" 
[...]

PR #115, opened 2018-09-17
Wait for http server to exit prior to closing (#114)

klarose:
We are running the test http servers in a goroutine. Consequently,
executing returns to the thread starting the server. This caller
assumes that the server is done, so it closes the server for graceful
shutdown. However, the server may not be done since it's running on a
separate thread and we're not waiting for that to finish. This leads to
the integration failing randomly.

Fix the problem by waiting for ctx->Done, like we're doing with the
grpc servers.

### DIFF
--- a/pkg/test/server.go
+++ b/pkg/test/server.go
@@ -66,10 +66,6 @@ func RunHTTP(ctx context.Context, upstreamPort uint) {
 			log.Error(err)
 		}
 	}()
-	<-ctx.Done()
-	if err := server.Shutdown(ctx); err != nil {
-		log.Error(err)
-	}
 }
 
 // RunAccessLogServer starts an accesslog service.
@@ -137,8 +133,4 @@ func RunManagementGateway(ctx context.Context, srv xds.Server, port uint) {
 			log.Error(err)
 		}
 	}()
-	<-ctx.Done()
-	if err := server.Shutdown(ctx); err != nil {
-		log.Error(err)
-	}
 }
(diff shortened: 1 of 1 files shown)


rshriram:
@9len or @kyessenov ?

pkg/test/server.go
```diff
@@ -66,6 +66,7 @@ func RunHTTP(ctx context.Context, upstreamPort uint) {
 			log.Error(err)
 		}
 	}()
+	<-ctx.Done()
 	if err := server.Shutdown(ctx); err != nil {
```
wozz: isn't this using an expired context now? will the clean shutdown actually do anything?

klarose: I'm admittedly very inexperienced with go, so I may be wrong, but from reading the documentation on Server.Shutdown (https://golang.org/pkg/net/http/?#Server.Shutdown), I think you are correct. My understanding is that it is intended for when you're about to shut down a server, so that you finish processing any outstanding work, without accepting new connections.

I actually can't think of why it was being called in the first place in this test. We know exactly how much work to do, so there should be no surprises awaiting clients when we exit at the end of the test. Do you have any idea why?

As for a better fix, do I even need the <-ctx.Done(), or should I just remove it and the shutdown all-together? The test seems to pass without it, but I guess this could be used as a way to keep the outer goroutine from exiting prior to the test being finished? Is that good practice?

wozz: I guess the way I would do it is create a sub context with `context.WithCancel`, so that the context can be cancelled, which will actually end the sub context and continue to the server shutdown with the parent context.  I'm not sure how you would ensure the cancel function is called when the tests finish.  Just leaving out the clean shutdown might be fine too.

Another option might be to switch to using https://golang.org/pkg/net/http/httptest/ since this is a test package.

klarose: I just removed it all-together.

PR #118, opened 2018-09-27
bugfix:test HTTPServer exits immedidately after startup

maplebeats:
#114 #115


rshriram:
please fix dco

kyessenov:
Seems like this is no longer necessary since `server.Shutdown` has been removed.

PR #124, opened 2018-10-15
Update instructions related to proto files

venilnoronha:
This PR adds instructions about generating proto files if there's an update to the `data-plane-api` dependency.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,22 +51,29 @@ make tools
 make depend.install
 ```
 
-2. Edit the code in your favorite IDE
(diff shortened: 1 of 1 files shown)


kyessenov:
Does code not build automatically?
We can adjust the generation script is there is a specific problem.

venilnoronha:
I've had to do this manually every time. I don't see any references to the `generate` make target in the vendor targets.

venilnoronha:
ping

PR #140, opened 2019-01-03
add Version tagging section

vadimeisenbergibm:
@ggreenway @dio For your review.


taion809:
I agree this would be pretty hard to apply consistently and asking for a suggested tag is also a bit onerous.

Would it be possible to create a series of yaml files related to common use cases that can be used for testing the generated protobuf and if these break or require updates we know to apply a major version update and if not we can carry on business as usual?

Issue #153, opened 2019-01-25
Allow Callbacks.OnStreamRequest() to return an error to close streams


biefy:
When trying to add a simple rate limiting into an implementation of Callbacks, it came to me that OnStreamRequest() cannot close stream processing as it does not return an error.  It could be a common case to allow server close stream on specific requests.

alecholmez:
I think we can close. Right @biefy ?

PR #155, opened 2019-01-29
Allow Callbacks.OnStreamRequest() to return an error to close stream

biefy:
As mentioned #153, this PR allows Callbacks.OnStreamRequest() to return an error to close streams.  In certain circumstances, rate limiting, for example, server should be able to make a decision to terminate a stream after it receives a stream request.


kyessenov:
Note: will need a release note in the next release.

PR #156, opened 2019-01-30
test: use typed configs

kyessenov:
Envoy switched to using protobuf Any in place of Struct when the config type is well known.


kyessenov:
Ping @rshriram . This is just a sample in the test.

PR #163, opened 2019-03-29
Add a GetSnapshot method to SnapshotCache interface

jaricftw:
Currently `SnapshotCache` doesn't have a way to get the snapshot for a `node`, not even checking whether a snapshot exist for a `node`.  

This adds a `GetSnapshot` method to the `SnapshotCache` interface. Example use case: when implementing the OnStreamRequest callback, it needs to know whether the snapshot for a `node` has already been generated or not, and generate snapshot only if it doesn't

### DIFF
--- a/pkg/cache/simple.go
+++ b/pkg/cache/simple.go
@@ -47,6 +47,9 @@ type SnapshotCache interface {
 	// the version differs from the snapshot version.
 	SetSnapshot(node string, snapshot Snapshot) error
 
+	// GetSnapshots gets the snapshot for a node.
+	GetSnapshot(node string) (Snapshot, error)
+
 	// ClearSnapshot removes all status and snapshot information associated with a node.
 	ClearSnapshot(node string)
 }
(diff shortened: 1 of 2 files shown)


pkg/cache/simple.go
```diff
@@ -122,6 +125,18 @@ func (cache *snapshotCache) SetSnapshot(node string, snapshot Snapshot) error {
 	return nil
 }
 
+// GetSnapshots gets the snapshot for a node, and returns an error if not found.
+func (cache *snapshotCache) GetSnapshot(node string) (Snapshot, error) {
+	cache.mu.Lock()
```
kyessenov: `RLock()` is more appropriate here?

jaricftw: Oh sure, missed that.

jaricftw:
> Looks fine. Can you add a note about lack of lock protection for the content of the cache itself? E.g. do not modify the snapshot directly since it's being read concurrently.

What do you mean "lack of lock protection"? We do have the `RWMutex` protecting the cache content. Writes is locked exclusively so there will not be any concurrent writes or reads.

PR #179, opened 2019-05-08
introduce go modules

kyessenov:


davecheney:
LGTM. In my module project I also used a proxy as it sped up fetching dependencies and seemed to make them more reliable -- go modules seemed to hit github harder than dep or glide.

https://github.com/heptio/contour/blob/master/Dockerfile#L4

kyessenov:
Interesting. Thanks for the suggestion about GOPROXY. I'm trying to solve the issue with non-go dependencies (like data-plane-api).  I don't see how to use go modules to import them, so perhaps, need another way to fetch that repo.

davecheney:
> I don't see how to use go modules to import them, so perhaps, need another way to fetch that repo.

I know its probably swearing in church, but `git submodule?`

davecheney:
Or I'd try to solve this with a Makefile and some bash, perhaps someone else wants to argue that bazel can do this. (I don't know, i'm living a bazel free life)

kyessenov:
After much pain, I think I've simplified the build to avoid using docker during presubmit.
Generation is the next step, but this is ready to go as-is.

Dockerfile.ci
```diff
@@ -0,0 +1,26 @@
...
+RUN go get github.com/envoyproxy/protoc-gen-validate@v0.0.14
+
+# Install protoc 3.6.1
+RUN apt-get update && apt-get install unzip
+RUN wget -O /tmp/protoc.zip https://github.com/protocolbuffers/protobuf/releases/download/v3.6.1/protoc-3.6.1-linux-x86_64.zip  \
+  && unzip /tmp/protoc.zip -d /usr/local/ \
+  && rm /tmp/protoc.zip
+
+# circle does not like files in the checkout directory
+RUN rm go.mod
```
davecheney: nit: you can rm both of these on the same line, it doesn't really matter, I just get annoyed with how many intermediary containers docker build creates

Dockerfile.ci
```diff
@@ -0,0 +1,26 @@
...
+RUN rm go.sum
+
+# add envoy
+# Due to https://github.com/envoyproxy/envoy/issues/6237, we have to pin envoy temporarily
+COPY --from=envoyproxy/envoy:0634e7bcbfea7317eaf805a657d048ffa27f519a /usr/local/bin/envoy /usr/local/bin/envoy
```
davecheney: wow, i didn't know you could copy from an arbitrary hash; TIL

Dockerfile.ci
kyessenov: I'll fix it up later. Still need to wire code generation through this container.

Dockerfile.ci
kyessenov: yeah, this is a side effect of staged builds. pretty neat.

PR #185, opened 2019-05-22
Several nodes may look to single snapshot

WIgor:
Hi

Want to propose this change. 

The main idea - in many cases, several nodes will have the same set of resources. And especially we don't need to know in advance exact names of nodes that we want to configure. But obviously, watchers must be tracked separately depending on the exact node name.


kyessenov:
Hey, sorry for the lack of reviews. I can only spend a limited amount of my time here. In general, this idea makes sense, but I didn't have time to look at the technicalities.

WIgor:
Actually, as I realized later, no much sense. I thought that can't be several watchers for the single "node". It can - so no reason to overcomplicate this.

PR #187, opened 2019-05-23
Remove deprecated cluster.hosts in sample

mpitt:
Substitute `cluster.hosts` with `cluster.load_assignment` in the sample configs, as per [deprecation](https://www.envoyproxy.io/docs/envoy/latest/intro/deprecated#version-1-8-0-oct-4-2018).

### DIFF
--- a/sample/bootstrap-ads.yaml
+++ b/sample/bootstrap-ads.yaml
@@ -21,9 +21,14 @@ node:
 static_resources:
   clusters:
   - connect_timeout: 1s
-    hosts:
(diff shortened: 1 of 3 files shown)


kyessenov:
Thanks! Does `make integration` pass on your machine? We might be using an older envoy in the CI here which would explain the failing test.

kyessenov:
It got "un-deprecated" but we should still merge this.

kyessenov:
Redone wit #243

PR #202, opened 2019-08-01
Replace gogo/googleapis with istio.io/gogo-genproto

rshriram:
googleapis are already in istio.io/gogo-genproto [which is already a dependency for this repo]. So, its better to use the googleapis in the gogo-genproto dependency so that everything is consistent rather than directly importing from gogo/googleapis which could go out of sync with whats in gogo-genproto/googleapis
Signed-off-by: Shriram Rajagopalan <rshriram@tetrate.io>


geeknoid:
@kyessenov @yangminzhu This is blocking a huge PR in istio/istio, I'd appreciate  speedy approval.

Thanks.

kyessenov:
Istio should not be hosting those APIs but given that no-one else is willing to do that, this is fine.

kyessenov:
Discussion in https://github.com/istio/api/pull/1014#issuecomment-517399623

jyotimahapatra:
@rshriram I have to update the data-plane-api in go-control-plane and the generate proto command fails with `google/api/expr/v1alpha1/syntax.proto: File not found.
envoy/config/rbac/v2/rbac.proto: Import "google/api/expr/v1alpha1/syntax.proto" was not found or had errors.
envoy/config/rbac/v2/rbac.proto:99:3: "google.api.expr.v1alpha1.Expr" is not defined.` because the mod version https://github.com/envoyproxy/go-control-plane/blob/master/go.mod#L13 does not have syntax.proto. I am not too versatile with protos. Can you help me?

kyessenov:
I think this was a mistake. We should revert to use gogoprotobuf definitions and add there what's missing.

jyotimahapatra:
I'm not too sure about the protos and might be doing something wrong, but if we could update the data-plane-api in go-control-plane, it would help me.

jyotimahapatra:
Also, looks like envoy uses an archived version of googleapis? https://github.com/envoyproxy/envoy/blob/master/api/bazel/repository_locations.bzl#L37

kyessenov:
Can someone explain why this was needed? I let it get merged since there was some urgency but now I regret this.

lita:
This is hindering Lyft's ability to upgrade to the new configs as now we are required to vendor istio which is not something we want to do. 

Can we discuss a workaround as @kyessenov was mentioning?

lita:
Nevermind. I see https://github.com/gogo/googleapis/pull/7 and this is getting fixed. Thank you so much @kyessenov!

PR #209, opened 2019-08-13
Properly reply to envoy clients when the REST fetch is skipped.

NickBOA:
When an envoy client initiates a REST fetch, `go-control-plane` compares the client's current version against the proper current version. If these versions match then the fetch is short-circuited and "errors". This erroring culminates in a HTTP status code 500 being sent to the client (just like with any generic error). Due to the change [here](https://github.com/envoyproxy/envoy/pull/6948) and the documentation [here](https://www.envoyproxy.io/docs/envoy/latest/configuration/overview/v2_overview#rest-endpoints) in `envoyproxy`, skipped fetches should be signaled to the client as HTTP status code 304.

This PR gives the skipped fetch path a specific error type (in `simple.go`) and then checks for that specific error type in `gateway.go`. A 304 is specially returned when the error is of the `SkipFetchError` type. This will allow the `envoy` client to properly understand that the fetch was skipped and it will no longer log the following warning:

```
[28][warning][config] [bazel-out/k8-opt/bin/source/common/config/_virtual_includes/http_subscription_lib/common/config/http_subscription_impl.h:115] REST config update failed: fetch failure
```

### DIFF
--- a/pkg/util/wellknown.go
+++ b/pkg/util/wellknown.go
@@ -113,3 +113,12 @@ const (
 	// HTTPGRPCAccessLog sink for the HTTP gRPC access log service
 	HTTPGRPCAccessLog = "envoy.http_grpc_access_log"
 )
+
+// SkipFetchError is the error returned when the cache fetch is short
+// circuited due to the client's version already being up-to-date.
+type SkipFetchError struct{}
+
+// Error satisfies the error interface
+func (e SkipFetchError) Error() string {
+	return "skip fetch: version up to date"
+}
--- a/pkg/cache/simple.go
+++ b/pkg/cache/simple.go
@@ -16,13 +16,13 @@ package cache
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"sync"
 	"sync/atomic"
 	"time"
 
 	"github.com/envoyproxy/go-control-plane/pkg/log"
+	"github.com/envoyproxy/go-control-plane/pkg/util"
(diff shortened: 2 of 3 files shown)


pkg/util/wellknown.go
```diff
@@ -113,3 +113,12 @@ const (
 	// HTTPGRPCAccessLog sink for the HTTP gRPC access log service
 	HTTPGRPCAccessLog = "envoy.http_grpc_access_log"
 )
+
+// SkipFetchError is the error returned when the cache fetch is short
+// circuited due to the client's version already being up-to-date.
+type SkipFetchError struct{}
```
yangminzhu: I feel a little strange to put this here, is there anyone else other than `pkg/server/gateway.go` going to need this? Should it be put somewhere in `pkg/cache` actually? cc @kyessenov

NickBOA: I too wasn't sure about where this should land. I am not super versed in this codebase, but I did glance around for a spot I thought was fitting. I am certainly open to moving it wherever you all think best. I don't **think** this exact error type would be used anywhere else; but again, I'm deferring to your all's judgement.

kyessenov: I think it's fine here given it's an internal implementation detail. This util package needs to be broken down later but it should not be coupled with this change.

NickBOA:
I only found out because I ran into. No problem at all!

PR #216, opened 2019-08-21
fix accesslog example

### LINKED ISSUE #215: Access log example causes CPU 100%
Deployed to production something similar to to the example to find out that it can cause 100% CPU load.

Reproduce:
1. Run example
2. Stop envoy
3. Look at the CPU goes to 100% 

**Issue**: 
https://github.com/envoyproxy/go-control-plane/blob/9abdeff70b1fcebc846b2518ccb73478fa10164e/pkg/test/accesslog.go#L37-L41

**Possible Solution**:
```go
 for { 
 	msg, err := stream.Recv() 
 	if err != nil { 
 		return err
 	} 
```

I'm not entirely this is the right solution, but it seems to work.

kyessenov:
Fixes #215 
Thanks for reporting!

### DIFF
--- a/pkg/test/accesslog.go
+++ b/pkg/test/accesslog.go
@@ -37,7 +37,7 @@ func (svc *AccessLogService) StreamAccessLogs(stream als.AccessLogService_Stream
 	for {
 		msg, err := stream.Recv()
 		if err != nil {
-			continue
+			return
 		}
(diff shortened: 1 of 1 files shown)


kyessenov:
Can you please re-approve? Made a typo in the original version.

PR #217, opened 2019-08-22
Update data-plane-api to 93a6ddd168406d28f4a720511777e48976f8f62e

thomasco:


kyessenov:
If you're trying to pick up CEL changes, you are likely to hit a problem with lack of gogo generated files for it...

thomasco:
Hm, looks like that's the next commit after 507755f, the current submodule ref. I'm trying to pick up 93a6ddd, the ext_authz change.

thomasco:
generate_protos.sh can't find `google/api/expr/v1alpha1/syntax.proto`, which I'm assuming should come from `istio.io/gogo-genproto`. I updated the mapping in `generate_protos.sh` and changed `go.mod` to pull in latest, but https://github.com/istio/gogo-genproto/commit/929161cd91197d0ba0eab67b083a886db13295cd removed the proto files from `gogo-genproto`, so I'm stuck again.

thomasco:
Opened istio/gogo-genproto#19 about the proto files, but I'm not sure if that's the right way to go here

kyessenov:
We switched to using golang/protobuf. It should be up to date now.

PR #221, opened 2019-09-03
data-plane-api 7818642b2294cee404137a8fb97b65cba89d1b2d

dschaller:


kyessenov:
There's some work done to clean up dependency on istio/gogo-genproto. That dependency moved to use an unreleased version of gogoprotobuf and causing issues.

dschaller:
Yeah I noticed that this is an issue. We should probably figure out a longer term solution for things like this since we are making breaking changes into an upstream project before the downstream projects can build properly.

kyessenov:
Longer term we should use golang/protobuf. See #202 .

kyessenov:
Ref https://github.com/gogo/googleapis/pull/7

jyotimahapatra:
@mattklein123 fyi. In order to update the data-plane-api, we need https://github.com/gogo/googleapis/pull/7 and a subsequent fix to use the googleapis in go-control-plane.

jyotimahapatra:
@kyessenov Looks like the data-plane-api submodule is now removed. and the `generate_proto.sh` has been removed too.
We want to use the lates go-control-plane in our control plane to deal with few new deprecated fields. 
Is there a new version of go-control-plane now that we can use?

kyessenov:
Master is using golang/protobuf, which I strongly suggest to start using if you plan to build a new control plane.

jyotimahapatra:
Thanks @kyessenov . I will move our control plane to use master.

PR #236, opened 2019-10-12
Fix the order of cancel call, or it will never run.

beautytiger:


beautytiger:
> I think this is intentional. Watch is cancelled as part of the loop iteration. Go loop variables are not scoped to the loop body, unlike Java/C/C++.

I think `cancel` func is defined in loop body, and only used inside its test func, if it is not called in its own func, it will never be called.
it is not a nonlocal var or a global var, is it?

Issue #237, opened 2019-10-31
Graceful shutdown for the streaming handlers


jyotimahapatra:
We recently had a change for graceful shutdown of the control plane and used `grpc.GracefulStop()` .
`GracefulStop` sends goaway for new connections, but blocks until the current open connections are open. We need a way to add a signal in the handler https://github.com/envoyproxy/go-control-plane/blob/8e1e5558642ae54a35eca889ba37bc550e131216/pkg/server/server.go#L165 to recognize a shutdown and return from the handler.

kyessenov:
Fixed by the PR.

PR #239, opened 2019-11-05
Adding support to shutdown handler

jyotima:
We are missing a way in the go-control-plane to cleanly shutdown the handler. 

Scenario: We are implementing a scenario to gracefully shutdown the grpc server running the go-control-plane. The grpc server shutdown will stop accepting new connections, but will block until the existing connections are closed. 

With this PR, we are trying to add support for shutting down the stream by depending on an additional `done` channel.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -69,8 +69,8 @@ type Callbacks interface {
 }
 
 // NewServer creates handlers from a config watcher and callbacks.
-func NewServer(config cache.Cache, callbacks Callbacks) Server {
-	return &server{cache: config, callbacks: callbacks}
+func NewServer(config cache.Cache, callbacks Callbacks, done chan bool) Server {
+	return &server{cache: config, callbacks: callbacks, done: done}
 }
 
 type server struct {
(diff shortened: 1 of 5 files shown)


jyotima:
#237

jyotima:
@kyessenov Does this approach looks reasonable?

pkg/server/server.go
```diff
@@ -69,8 +69,8 @@ type Callbacks interface {
 }
 
 // NewServer creates handlers from a config watcher and callbacks.
-func NewServer(config cache.Cache, callbacks Callbacks) Server {
-	return &server{cache: config, callbacks: callbacks}
+func NewServer(config cache.Cache, callbacks Callbacks, done chan bool) Server {
+	return &server{cache: config, callbacks: callbacks, done: done}
```
kyessenov: This looks fine, but I think it's more idiomatic to use `context.Context` and then select on `context.Done()`. WDYT? See https://golang.org/pkg/context/#Context

jyotimahapatra: Thank you for the suggestion. That completely makes sense.
I have corrected the usage now. LMK if you have more suggestions.

jyotima:
Thank you!!

Issue #238, opened 2019-11-01
Add the support for caching serialized protobuf bytes of xDS response


freedomljc:
Currently, go-control-plane stores the protobuf format of xDS response in the cache, and [marshals](https://github.com/envoyproxy/go-control-plane/blob/8e1e5558642ae54a35eca889ba37bc550e131216/pkg/server/server.go#L147) it when sending out to envoy. 

If thousands of downstream machines watching the xDS change of one upstream service, it will marshal the same data thousands of times when sending the xDS response to envoy. The issue will get amplified if the xDS of the upstream service gets change very quickly(Especially EDS part).

We noticed that when the upstream services scale up quickly, it causes the cpu spike of go-control-plane server.

To solve the issue, we'll need to add the support of caching serialized xDS response.

kyessenov:
Sounds reasonable. We should probably keep serialized protobufs in the cache plus little metadata.

freedomljc:
@kyessenov , I just sent out the PR https://github.com/envoyproxy/go-control-plane/pull/240, and it is to allow users to marshal the resource while implementing the cache interface. 
Pls let me you your thoughts.

lita:
@freedomljc is it possible to just store marshall responses in the cache vs storing both?

freedomljc:
@lita , from the implementation of go-control-plane, the caching logic is optional, and the built-in caching logic is for the the entry-level usage. We only need to update the interface `CreateWatch` and `Fetch`, and there exist some discussions around it: https://github.com/envoyproxy/go-control-plane/pull/240#discussion_r343329534. pls let me know if your have further questions.

PR #240, opened 2019-11-05
Support marshaling resource

freedomljc:
The PR is about the issue https://github.com/envoyproxy/go-control-plane/issues/238.
In this PR, it updates the interface of Response https://github.com/envoyproxy/go-control-plane/blob/e78985eb6c01876c02fb837c077253d03b5e6425/pkg/cache/cache.go#L60-L70, and allow users to specify whether the resource is marshaled, when implementing the interface of Cache: https://github.com/envoyproxy/go-control-plane/blob/e78985eb6c01876c02fb837c077253d03b5e6425/pkg/cache/cache.go#L32-L57

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -65,8 +65,19 @@ type Response struct {
 	// Proxy responds with this version as an acknowledgement.
 	Version string
 
+	// The value indicating whether the resource is marshaled, and only one of `Resources` and `MarshaledResources` is available.
+	ResourceMarshaled bool
+
 	// Resources to be included in the response.
 	Resources []Resource
+
+	// Marshaled Resources to be included in the response.
+	MarshaledResources []MarshaledResource
+}
+
+type MarshaledResource struct {
+	Bytes []byte
+	ResourceType string
 }
 
 // SkipFetchError is the error returned when the cache fetch is short
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -75,10 +75,7 @@ type Response struct {
 	MarshaledResources []MarshaledResource
 }
 
-type MarshaledResource struct {
-	Bytes []byte
-	ResourceType string
-}
+type MarshaledResource = []byte
 
 // SkipFetchError is the error returned when the cache fetch is short
 // circuited due to the client's version already being up-to-date.
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -75,6 +75,7 @@ type Response struct {
 	MarshaledResources []MarshaledResource
 }
 
+// MarshaledResource is an alias for the serialized binary array.
 type MarshaledResource = []byte
 
 // SkipFetchError is the error returned when the cache fetch is short
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -21,7 +21,6 @@ import (
 	"strconv"
 	"sync/atomic"
 
-	"github.com/golang/protobuf/proto"
 	"github.com/golang/protobuf/ptypes/any"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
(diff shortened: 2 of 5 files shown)


pkg/cache/cache.go
```diff
@@ -65,8 +65,19 @@ type Response struct {
...
+	ResourceMarshaled bool
+
 	// Resources to be included in the response.
 	Resources []Resource
+
+	// Marshaled Resources to be included in the response.
+	MarshaledResources []MarshaledResource
+}
+
+type MarshaledResource struct {
```
jessicayuen: Perhaps we should put this in the resource file: https://github.com/envoyproxy/go-control-plane/blob/master/pkg/cache/resource.go

pkg/server/server.go
```diff
@@ -138,19 +137,32 @@ func createResponse(resp *cache.Response, typeURL string) (*v2.DiscoveryResponse
 	if resp == nil {
 		return nil, errors.New("missing response")
 	}
-	resources := make([]*any.Any, len(resp.Resources))
-	for i := 0; i < len(resp.Resources); i++ {
+
+	var resources []*any.Any
+	if resp.ResourceMarshaled {
+		resources = make([]*any.Any, len(resp.MarshaledResources))
+	} else{
```
jessicayuen: nit: space between `else` and `{`

pkg/cache/cache.go
```diff
@@ -65,8 +65,19 @@ type Response struct {
 	// Proxy responds with this version as an acknowledgement.
 	Version string
 
+	// The value indicating whether the resource is marshaled, and only one of `Resources` and `MarshaledResources` is available.
+	ResourceMarshaled bool
```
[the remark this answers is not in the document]
jessicayuen: Who specifies this value? If this is for the go-control-plane library consumer to determine whether to use marshaling, should it be a flag in the Request object / Fetch function instead?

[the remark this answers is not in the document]
freedomljc: For the length check, i think there potentially exist two issues:
- If the resource is unavailable for the specific request, either Resources  or MashaledResources will be empty, should we throw the exception?
- There isn't a good way to add the length check from current design. If adding the length check into the Response, we will need to enforce updating the Resources/MashaledResources via some method, and adding the length check in that method. Another option is to add length check when setting the value to Response, but it happens when implementing the interface https://github.com/envoyproxy/go-control-plane/blob/e78985eb6c01876c02fb837c077253d03b5e6425/pkg/cache/cache.go#L42 and https://github.com/envoyproxy/go-control-plane/blob/e78985eb6c01876c02fb837c077253d03b5e6425/pkg/cache/cache.go#L50, then it is inappropriate since it doesn't enforce the check.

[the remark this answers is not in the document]
freedomljc: Regarding who specifies the value, it happens when go-control-plane library consumers implement the interface https://github.com/envoyproxy/go-control-plane/blob/e78985eb6c01876c02fb837c077253d03b5e6425/pkg/cache/cache.go#L42 and https://github.com/envoyproxy/go-control-plane/blob/e78985eb6c01876c02fb837c077253d03b5e6425/pkg/cache/cache.go#L50.

There shouldn't a flag in the request object, since the actual xDS response sending back to envoy doesn't get changed. 
And the flag will be set when implementing the interface CreateWatch/Fetch.

pkg/cache/cache.go
freedomljc: I can see it in both ways, keeping MarshaledResource in the cache.go makes the file resource.go more isolated. But I'm open to it, and it depends on the maintainers' preference.

pkg/server/server.go
freedomljc: good catch, will update later.

pkg/cache/cache.go
```diff
@@ -65,8 +65,19 @@ type Response struct {
...
+
 	// Resources to be included in the response.
 	Resources []Resource
+
+	// Marshaled Resources to be included in the response.
+	MarshaledResources []MarshaledResource
```
kyessenov: Maybe we should use `*any.Any` instead of creating a new struct? It fulfills the same purpose, and it's a bit easier to respond after caching serialized resource.

freedomljc: Just realized that there is no need to add `ResourceType` into the `MarshaledResources`, and declare it as the binary array. It looks better now.

pkg/cache/cache.go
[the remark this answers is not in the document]
jyotimahapatra: @freedomljc Can you also demonstrate the way it will be set in https://github.com/envoyproxy/go-control-plane/blob/master/pkg/cache/simple.go#L89 so that it will serve as an example?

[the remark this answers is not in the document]
freedomljc: @jyotima , simple.go has implemented the interface Fetch/CreateWatch in: https://github.com/envoyproxy/go-control-plane/blob/2e02e0ea3899d0e6075d62a08fb9647fde055b4d/pkg/cache/simple.go#L168 and https://github.com/envoyproxy/go-control-plane/blob/2e02e0ea3899d0e6075d62a08fb9647fde055b4d/pkg/cache/simple.go#L277.

To support the flag `ResourceMarshaled` in simple.go, we'll need to add the flag in the method Fetch/CreateWatch, and also refactor the logic, then check the flag in https://github.com/envoyproxy/go-control-plane/blob/2e02e0ea3899d0e6075d62a08fb9647fde055b4d/pkg/cache/simple.go#L249 to see whether to marshal the resource before sending back to envoy.

However, just as the filename `simple` indicates, the file is mainly for the entry-level usage of go-control-plane, I'm hesitating to make the change there.

freedomljc:
@kyessenov , when you get the chance, can you take a look again? thanks.

jyotimahapatra:
@freedomljc since there are no tests around this change, im afraid of regression in future changes. We probably need a way to add some tests. WDYT?

freedomljc:
@jyotimahapatra , thanks for raising the question. It would be nice to add the integration test.
From the current end_to_end test(https://github.com/envoyproxy/go-control-plane/tree/master/pkg/test/main), it seems that we only cover the simple case(https://github.com/envoyproxy/go-control-plane/blob/master/pkg/cache/simple.go), and doesn't cover the case if we are implementing the interface by our own. 
I'll be happy to review the proposals/PRs.

PR #244, opened 2019-11-16
cache: use array instead of struct

kyessenov:
This should make it easier to extend the snapshot with other types.


kyessenov:
@yangminzhu Can you take a quick look? This is in preparation for UDPA/xDS v3 changes.

pkg/cache/resource.go
```diff
@@ -46,18 +46,38 @@ const (
...
+type ResponseType int
+
+const (
+	Endpoint ResponseType = iota
+	Cluster
+	Route
+	Listener
+	Secret
+	Runtime
+	UnknownType // token to count the total number of supported types
```
yangminzhu: nit: The `UnknownType` looks a little strange to represent the `total number of supported types`, especially when looking in the `type Snapshot struct` definition below.

kyessenov: Yeah... It's also used as a hole for unknowns though. So not sure if it's better to call a hole TotalNumberTypes or the number UnknownType.

PR #261, opened 2020-01-28
Fix Integration Test Using Docker

alecholmez:
This creates a startup wrapper script that will eliminate the need for users have artifacts locally first. It will build and use the pre-existing Envoy binaries in the released containers.

The only new dependency is Docker but if need be that can be stated in the README.


alecholmez:
Let me clean up these commits

PR #262, opened 2020-01-29
Fixed Integration Tests With Docker Wrapper

alecholmez:
This fixes our integration tests using a docker wrapper which enables developers to no longer have a local envoy binary on their machine.


alecholmez:
```bash
echo "Building test linux binary"
Building test linux binary
build github.com/cncf/udpa/go/udpa/annotations: cannot load github.com/cncf/udpa/go/udpa/annotations: open /go-control-plane/vendor/github.com/cncf/udpa/go/udpa/annotations: no such file or directory
Makefile:33: recipe for target 'bin/test' failed
```
I think the go-control-plane CI container might need to pull in this new dependency or I can change the CI file to run `go mod vendor` every time it builds

Makefile
```diff
@@ -30,18 +30,19 @@ format:
 .PHONY: $(BINDIR)/test integration integration.ads integration.xds integration.rest integration.ads.tls
 
 $(BINDIR)/test:
-	@go build -race -o $@ pkg/test/main/main.go
+	echo "Building test linux binary"
+	@CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -mod=vendor -a -tags netgo -ldflags '-w -extldflags "-static"' -o $@ pkg/test/main/main.go
```
kyessenov: can you enable the race detector with `-race`?

Makefile
```diff
@@ -30,18 +30,19 @@ format:
...
 
 integration: integration.xds integration.ads integration.rest integration.ads.tls
 
 integration.ads: $(BINDIR)/test
-	env XDS=ads build/integration.sh
+	env XDS=ads build/run_integration.sh
```
kyessenov: I think you can't run docker in circle CI? When I tried it was really painful so I just made the CI image be based on envoyproxy/envoy:<release>.

Makefile
kyessenov: `-mod=vendor` is breaking the build since we should be using modules.

alecholmez: > can you enable the race detector with `-race`?

I'll try and get that back in. Turning CGO off disables the `-race` flag but I can enable it again.

Makefile
alecholmez: > I think you can't run docker in circle CI? When I tried it was really painful so I just made the CI image be based on envoyproxy/envoy:.

Ah gotcha, this fixed the local test but yea not CI is broken. I'll try and work something clever but I would l like to keep the local test using docker

justincely: so,  we can run docker in Circle, but it's a bit of a pain.  Circle won't let you mount any volumes or hit any exposed ports from the circle machine for security reasons.   You can sometimes get past this by running any needed tests from ANOTHER image, but it's a lot of pain with `docker cp` to move scripts and tests around. 

The other way to get docker to work straight in circle is to use the older `machine` executor type.  The problem is you don't get to choose your image anymore; you're dumped into an ubuntu image from which you need to install everything you need (and hope it's supported).  We go this route for many of our projects; but to keep the usage of the `machine` executor as limited as possible we split out those tests into a separate circle step.  May be worth doing here if you do want to run these tests.

kyessenov: I tried using docker in the past, but IMHO it's not worth it. Using a base image with a _binary_ is simpler and faster. The only drawback is that we can't use latest produced envoy binaries, which may be a plus since Envoy does introduce breaking changes sometimes.

alecholmez:
Closing... will re-open shortly

PR #274, opened 2020-03-18
Remove the use of a deprecated field

jyotimahapatra:
The change is only in a test file. Moving dependency of the test from deprecated config into typed_config

### DIFF
--- a/pkg/test/resource/resource.go
+++ b/pkg/test/resource/resource.go
@@ -336,7 +336,7 @@ func (ts TestSnapshot) Generate() cache.Snapshot {
 
 		if ts.TLS {
 			for i, chain := range listener.FilterChains {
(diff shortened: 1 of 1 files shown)


jyotimahapatra:
@kyessenov PTAL

This will help remove deprecated fields from go-control-plane and help me consolidate a plan to move v2->v3

pkg/test/resource/resource.go
```diff
@@ -350,6 +350,13 @@ func (ts TestSnapshot) Generate() cache.Snapshot {
 						},
 					},
 				}
+				mt, _ := ptypes.MarshalAny(tlsc)
+				chain.TransportSocket = &core.TransportSocket{
+					Name: "envoy.transport_sockets.tls",
```
kyessenov: FYI: name can be anything now. Type takes priority over name.

PR #275, opened 2020-03-18
Create v2 folder structure for cache

jyotimahapatra:
This is the first PR in the series of PRs to enable v3 api support.
This PR is creating a v2 folder structure for cache, so that subsequent PRs can use a a script and go generate to create a parallel v3 folder structure.


jyotimahapatra:
PTAL @kyessenov

PR #276, opened 2020-03-21
Create v2 separation for cache interface

jyotimahapatra:
This PR does refactoring such that a mapping is created between the import of v2 and v3.
The PR only does a refactoring of module paths.

There;s a script that can be used to create a copy of v2 and `sed` the module paths into v3.
The new v3 files are not created to avoid creating a massive PR.


jyotimahapatra:
PTAL @kyessenov

jyotimahapatra:
closing this one because integration tests are failing. I will use this PR as a reference to the idea of the approach

jyotimahapatra:
> * it is not clear what the authoring model should be? Do you want to switch over to using v3 as the main version eventually and freeze v2? I think we need to document it somewhere in README.
I'm intending to let both versions stay active together. Further PRs will also extend the same pattern to some more folders in the code, like `server` and also add integration tests.
> * package names such as `common` and `util` are generally an anti-pattern. It's better to come up with a semantic name (maybe `types` or something), even if the package is small.
Let me attempt that.
> * I wonder how this plays with go module versioning. I think you can drop suffix `vX` from sibling imports (e.g. `x/y/v2` can use `x/y` as import), for example, but the rules are kind of complicated, so not sure if you want to rely upon them.
I can look at this aspect in a separate PR. will that work? Want to keep the PR from growing too big.

jyotimahapatra:
@kyessenov i have rename `common` and `utils` to `types` and `resource`.

PR #277, opened 2020-03-21
Create common and utils modules

jyotimahapatra:
This is a broken down version of [PR](https://github.com/envoyproxy/go-control-plane/pull/276).

This PR isolates a few common fields into `common` module and few distinctive types into `utils` module.

The rest of the PR is string replace to get the compilation working. After this separation is done, it will be easier to just copy v2 into a separate folder and replace the module paths.


jyotimahapatra:
@kyessenov PTAL

PR #279, opened 2020-03-24
Create v2 separation for server 

jyotimahapatra:
Following the same pattern as [this](https://github.com/envoyproxy/go-control-plane/pull/276), this PR creates a v2 folder for server interface. 

Next PRs will:
1. Create the v3 files using the script.
2. modify integration tests to include v3 apis
2. integrate script with CI


pkg/server/v2/server.go
```diff
@@ -26,26 +26,31 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
-	v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
-	v2grpc "github.com/envoyproxy/go-control-plane/envoy/api/v2"
+	clusterservice "github.com/envoyproxy/go-control-plane/envoy/api/v2"
```
kyessenov: Unfortunately, due to how google-internal bazel generates protobufs, we have to keep `v2` protos separate from `v2grpc` services. Do you think we can keep that if it is double imported?

jyotimahapatra: the `v2` has been renamed to `discovery` and each service is denoted by its own explicit name with a `service` prefix. The protos are not being mixed with service and should work as before. Do you suggest a change in the naming mechanism to separate them better?

jyotimahapatra: @kyessenov Can you describe more about the problem. AFAICT the modules are still double imported and protos and services are separate albeit with different names than before. Are you suggesting keeping the older names `v2` and `v2grpc` ?

kyessenov: Yeah, v2 has the protos and v2grpc has the services from the same package. It's just how google-internal bazel generates them and places them into distinct packages instead of the same one for OSS bazel.

kyessenov: awesome, I think if it's the case, then this takes care of my comment.

PR #282, opened 2020-03-26
Fix create_version CI

jyotimahapatra:
The ci logs showed that the v2 folder was landing inside v3. This PR makes sure that copy is working correctly and any edits to v3 folder cause CI failures.


jyotimahapatra:
@kyessenov PTAL

PR #284, opened 2020-03-29
Deprecate non-mandatory cache interface contracts

jessicayuen:
Deprecate from the cache interface:
* GetStatusInfo(string) StatusInfo
* GetStatusKeys() []string

These functions are not referenced anywhere in the [server
implementation](https://github.com/envoyproxy/go-control-plane/blob/e933a9f2ccd2c350a7d7c67317fceb79b3cd8ff6/pkg/server/server.go#L77).
and thus are not critical in the contract.

Oftentimes, consumers of the gcp library have had to implement the
contract but do little except `return nil` in the implementation. Rather
than make the functions mandatory, I've moved them over to the
SnapshotCache interface, where they are used.

This will not break existing implementations since users that have
implemented these functions remain operable; they're just no longer
required. Will reduce bloat in implementations that aren't using it.

Signed-off-by: Jess Yuen <jyuen@lyft.com>

### DIFF
--- a/pkg/cache/v2/cache.go
+++ b/pkg/cache/v2/cache.go
@@ -49,12 +49,6 @@ type Cache interface {
 
 	// Fetch implements the polling method of the config cache using a non-empty request.
 	Fetch(context.Context, Request) (*Response, error)
-
-	// GetStatusInfo retrieves status information for a node ID.
-	GetStatusInfo(string) StatusInfo
-
-	// GetStatusKeys retrieves node IDs for all statuses.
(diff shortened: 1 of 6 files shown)


jyotimahapatra:
Looks good to me. I don't see the api usage expect in `simple.go`.

jessicayuen:
cc @kyessenov Could you take a look?

kyessenov:
SGTM. I think it's used, but not critical for v3.

Issue #299, opened 2020-05-14
Support passthrough of Envoy DiscoveryResponses


jessicayuen:
In [xds-relay](https://github.com/envoyproxy/xds-relay), we have a use case of relaying DiscoveryResponses received upstream directly to the downstream Envoy clients.

We're utilizing go-control-plane's `CreateWatch` functionality to achieve this passthrough. A problem exists where the returned Discovery Response upstream already contains the marshaled resources. `CreateWatch` expects the responses to be of type [Response](https://github.com/envoyproxy/go-control-plane/blob/v0.9.5/pkg/cache/v2/cache.go#L54-L71) and then converts the response type to type DiscoveryResponse [later in the chain of events](https://github.com/envoyproxy/go-control-plane/blob/v0.9.5/pkg/server/v2/server.go#L143-L181).

We would like to add support to not do a DiscoveryResponse -> gcp.Response -> DiscoveryResponse conversion, and instead add an interface to support a direct passthrough of the final DiscoveryResponse.

Please refer to the discussion here for more details: https://envoyproxy.slack.com/archives/CTULFMN91/p1589474718031700

jessicayuen:
cc @kyessenov - I plan to open a PR to support this. Please let me know your thoughts.

jessicayuen:
Been thinking of ways we can approach this with pros/cons to each. Let me know your thoughts @jyotimahapatra @kyessenov 

1. Adding a new server and interface `PassthroughWatcher`, which contains a single contract `CreateWatch(Request) (value chan discovery.DiscoveryResponse, cancel func())`
- The existing server implementation is tightly coupled with the gcp Cache. This would allow us to add new support without coupling with the current contracts and adding bloat where implementations are not utilizing this new functionality. This would also allow us to create a implementation that doesn't rely on [Fetch](). This approach would involve the most work and require us to support two server types.

2. Adding a `discovery.DiscoveryResponse` variable to the [Response](https://github.com/envoyproxy/go-control-plane/blob/v0.9.5/pkg/cache/v2/cache.go#L54-L71) object.
This one is straightforward in that if this discovery response is set, we simply ignore the resource variables and do not to the conversion of `Response` -> `DiscoveryResponse` during processing. The downsides of this approach is that it creates confusion for those looking at the `Response` struct for why there exists both a DiscoveryResponse and separate resource arrays.

[...]

PR #300, opened 2020-05-15
Introduce Passthrough Response type

jyotimahapatra:
https://github.com/envoyproxy/go-control-plane/issues/299

While implementing https://github.com/envoyproxy/xds-relay/pull/85 , we realized that go-control-plane always marshals the response. As a result, we were not able to send an existing DiscoveryResponse as a passthrough. 
This PR introduces a new PassthroughResponse struct , which is not subjected to any marshalling while sending the response on the watches.


jyotimahapatra:
@jessicayuen This is a POC for conveying my thoughts on #299 . Let me know what you think

pkg/cache/v2/cache.go
```diff
@@ -51,21 +52,66 @@ type Cache interface {
 	Fetch(context.Context, Request) (*Response, error)
 }
 
+type ResponseType interface {
```
jessicayuen, line 55: How about:
```suggestion
type Serializable interface {
```

pkg/cache/v2/cache.go
```diff
@@ -51,21 +52,66 @@ type Cache interface {
...
+	GetMarshalled() ([]*any.Any, error)
```
jessicayuen, line 56: `GetMarshaledResources` would probably be more accurate. This could also be made `GetDiscoveryResponse` IMO, which might even be better for the pass through.

pkg/cache/v2/cache.go
```diff
@@ -51,21 +52,66 @@ type Cache interface {
...
+	ResponseType
+	// Request is the original request.
+	Request discovery.DiscoveryRequest
+
+	// Version of the resources as tracked by the cache for the given type.
+	// Proxy responds with this version as an acknowledgement.
+	Version string
+
+	// Resources to be included in the response.
+	Resources []*any.Any
```
jessicayuen, line 89: Mentioned in the issue, but how about storing the `DiscoveryResponse` instead?

pkg/cache/v2/cache.go
jyotimahapatra: makes sense.

pkg/cache/v2/cache.go
jyotimahapatra: how about `Marshallable` because that's a already recognizable term in golang?

jessicayuen: Since we're changing the contract to be `GetDiscoveryResponse`, maybe `ResponseType` is ok or `ResponseIface`

pkg/cache/v2/cache.go
jyotimahapatra: yes, that makes sense, will do

pkg/cache/v2/cache.go
```diff
@@ -51,21 +52,90 @@ type Cache interface {
...
+}
+
+type PassthroughResponse struct {
+	ResponseIface
+	// Request is the original request.
+	Request discovery.DiscoveryRequest
+
+	// Version of the resources as tracked by the cache for the given type.
+	// Proxy responds with this version as an acknowledgement.
+	Version string
```
jessicayuen, line 90: I'm wondering whether we can drop these two fields since they're already embedded in the `DiscoveryResponse`. Keeping them may just increase the odds of bugs where the values don't match the `DiscoveryResponse` fields.

jyotimahapatra: you're right. I'll get rid of these.

jyotimahapatra:
@jessicayuen PTAL. Once it looks good to you, we can ask Kuat to take a look next.

pkg/cache/v2/cache.go
```diff
@@ -40,32 +41,97 @@ type ConfigWatcher interface {
...
-	MarshaledResources []types.MarshaledResource
+	marshaledResponse *discovery.DiscoveryResponse
+}
+
+type PassthroughResponse struct {
+	ResponseIface
+	// Request is the original request.
+	Request discovery.DiscoveryRequest
+
+	// Resources to be included in the response.
```
jessicayuen, line 88: Comment needs updating.

pkg/cache/v2/cache.go
```diff
@@ -40,32 +41,97 @@ type ConfigWatcher interface {
 	//
 	// Cancel is an optional function to release resources in the producer. If
 	// provided, the consumer may call this function multiple times.
-	CreateWatch(Request) (value chan Response, cancel func())
+	CreateWatch(Request) (value chan ResponseIface, cancel func())
```
jessicayuen, line 44: After more thought, what do you think of keeping this named `Response`, but changing the `Response` struct to be named `DeconstructedResponse` or something? I just think passing `Response` around looks cleaner than `ResponseIface`, but I don't feel too strongly about this.

pkg/cache/v2/cache.go
jessicayuen, line 88: Comment needs updating.

pkg/cache/v2/cache.go
```diff
@@ -40,32 +41,97 @@ type ConfigWatcher interface {
...
+	// Resources to be included in the response.
+	Response *discovery.DiscoveryResponse
+}
+
+func (r Response) GetDiscoveryResponse() (*discovery.DiscoveryResponse, error) {
+	if r.isResourceMarshaled {
+		return r.marshaledResponse, nil
+	}
+
+	marshaledResources := make([]*any.Any, len(r.Resources))
```
jessicayuen, line 97: Might be worth adding a comment here to mention we marshal once so that subsequent calls don't require re-marshaling.

jyotimahapatra:
@kyessenov Please take a look.

pkg/cache/v2/cache.go
jyotimahapatra: I think I like ResponseIFace because it represents an interface.

jyotimahapatra:
@kyessenov Let us know what you feel about the change.

pkg/cache/v2/cache.go
```diff
@@ -40,32 +41,104 @@ type ConfigWatcher interface {
...
 type Cache interface {
 	ConfigWatcher
 
 	// Fetch implements the polling method of the config cache using a non-empty request.
-	Fetch(context.Context, Request) (*Response, error)
+	Fetch(context.Context, Request) (ResponseIface, error)
+}
+
+type ResponseIface interface {
+	// Get the Ccontructed DiscoveryResponse
```
mattklein123, line 56: typo Ccontructed

PR #304, opened 2020-06-02
update docs code to new version

maplebeats:
the cache demo code use an older version code, update to new version


kyessenov:
Thanks!

PR #305, opened 2020-06-03
Add CallbackFuncs helper type.

jpeach:
Add a CallbackFuncs helper type to make it simpler to implement the
`server.Callbacks` interface.

Signed-off-by: James Peach <jpeach@vmware.com>

PR #316, opened 2020-06-24
Add LoggerFuncs helper type.

jpeach:
Add a LoggerFuncs helper type to reduce the amount of boilerplate
code needed to adapt a logging API to the log.Logger interface.

Signed-off-by: James Peach <jpeach@vmware.com>


jpeach:
@jessicayuen This is just like #305 but for the Logger interface. Could you take a look?

PR #318, opened 2020-06-24
make isPerXdsConfig usable

therealmitchconnors:
This is a private interface with only one member, which is also private.  The effect is that it is impossible to set members of this type from outside the package.  It appears that the intent is that the five PerXdsConfig_* types would be usable on these members, but this is not the case.  

I am aware that this is a generated file, but I'm struggling to find data on how the file is generated, and to understand how we could modify generation to not create unusable interfaces.

For reference, the following code does not work before this PR, unless it is placed in the status package directly:

```
x := status.PerXdsConfig {
    PerXdsConfig: status.PerXdsConfig_ClusterConfig{},
}
```

The above currently fails to compile.

### DIFF
--- a/envoy/service/status/v3/csds.pb.go
+++ b/envoy/service/status/v3/csds.pb.go
@@ -149,7 +149,7 @@ type PerXdsConfig struct {
 	//	*PerXdsConfig_RouteConfig
 	//	*PerXdsConfig_ScopedRouteConfig
 	//	*PerXdsConfig_EndpointConfig
-	PerXdsConfig isPerXdsConfig_PerXdsConfig `protobuf_oneof:"per_xds_config"`
+	PerXdsConfig IsPerXdsConfig_PerXdsConfig `protobuf_oneof:"per_xds_config"`
 }
 
 func (x *PerXdsConfig) Reset() {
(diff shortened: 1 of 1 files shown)


kyessenov:
This is not going to work since you are editing an auto-generated file, which will be overwritten next time envoy syncs its protos.

therealmitchconnors:
@kyessenov I understand that.  I am looking for guidance on how to accomplish this the 'right' way.  As it is today, you cannot use go-control-plane to write a CSDS server.

kyessenov:
The proto says the interface is implemented by config dump protos: https://github.com/envoyproxy/envoy/blob/3d59b0b79b29cc839e149527720f9dd340709239/api/envoy/service/status/v2/csds.proto#L68
So it's implementable in that sense. In general, if there is an issue in the proto, I'd suggest filing an issue against Envoy to fix the proto, since this repo is not the source of truth for protobuf definitions.

therealmitchconnors:
The interface is only implemented by the config dump protos when they are used inside the Status package (and therefore have access to all the unexported interfaces and functions).  If you attempt the snippet I included in the description, you will see that you cannot use the config dump objects from outside the status package as isPerXdsConfig.

It seems to me this is not so much a problem with the protos as with the go code generation.  Do you know where the generator for these files is configured?

kyessenov:
It's using standard golang proto code generator. The generator doesn't really take any parameters, and it's driven via bazel in envoy repo. Did you try `status.PerXdsConfig { PerXdsConfig: &status.PerXdsConfig_ClusterConfig{}}` (note the pointer)?

therealmitchconnors:
That did the trick.  not sure how I missed that.  thanks.

PR #319, opened 2020-06-25
Upgrade Go dependencies.

jpeach:
Upgrade all the Go dependencies. The biggest change is the GRPC one,
which pulls in a numnber of other modules.

Signed-off-by: James Peach <jpeach@vmware.com>


jpeach:
> Just to double check, did you try to run go mod tidy?

Good point. Updated.

> Is there a reason to force gRPC upgrade for all users of this library?

Ugh, running `go mod tidy` always upgrades protobuf even on master :(

```
diff --git go.mod go.mod
index dbb7a23..fca6cf8 100644
--- go.mod
+++ go.mod
@@ -6,10 +6,11 @@ require (
        github.com/census-instrumentation/opencensus-proto v0.2.1
        github.com/cncf/udpa/go v0.0.0-20200313221541-5f7e5dd04533
        github.com/envoyproxy/protoc-gen-validate v0.1.0
-       github.com/golang/protobuf v1.3.2
-       github.com/google/go-cmp v0.2.0
+       github.com/golang/protobuf v1.4.1
+       github.com/google/go-cmp v0.5.0
        github.com/prometheus/client_model v0.0.0-20190812154241-14fe0d1b01d4
        github.com/stretchr/testify v1.5.1
-       google.golang.org/genproto v0.0.0-20190819201941-24fa4b261c55
-       google.golang.org/grpc v1.25.1
+       google.golang.org/genproto v0.0.0-20200526211855-cb27e3aa2013
+       google.golang.org/grpc v1.27.0
+       google.golang.org/protobuf v1.25.0
 )
```

kyessenov:
It's probably genproto pulling updates. I'm fine with this smaller update. Let's check with @jyoti @jessicayuen.

fxposter:
> using older protobuf with newer generated files causes perf degradation

we had ~4-5x performance drop when upgrading control plane that use current library version to newer protobuf module. I would say that all .pb.go should either be regenerated with new grpc/proto compiler or old proto module should be used :(

jpeach:
> > using older protobuf with newer generated files causes perf degradation
> 
> we had ~4-5x performance drop when upgrading control plane that use current library version to newer protobuf module.

Ouch. IIUC this upgrade doesn't force module clients to use protov2 module, but I'm not all that certain about whether that would happen implicitly with the gRPC version bump. I'm happy to drop this PR, or to remove all the module updates that pull in protobuf changes.

> I would say that all .pb.go should either be regenerated with new grpc/proto compiler or old proto module should be used :(

I think that would force all consumers of this module to use the protov2 package, right? That's probably a step we should take with more thought and discussion than this PR :)

kyessenov:
The way I understand it, is that protov2 is backwards compatible with protov1 but causes severe performance degradation. Hence, we _should_ make protov2 as a lower bound requirement since the protos here are generated with v1.22.0 to help users avoid making the mistake of using protov1 with protov2 generated protos.

WDYT?

fxposter:
That's kinda hard to reason about all this stuff.

For us - bumping "github.com/golang/protobuf" dependency to 1.4.x caused perf degradation.
I guess, if all .pb.go files in this repo will be recompiled with new protobuf version, you bump "github.com/golang/protobuf" to 1.4.x and release new version of the library - migration should be smooth for users. IE: after bumping "github.com/envoyproxy/go-control-plane" dependency they will get new .pb.go files (compiled by newer version of protobuf) and new version of "github.com/golang/protobuf" library that works fine with them, while maintaining old API.

Not updating "github.com/golang/protobuf" to 1.4.x or later in this library is (at least for us) not a very good choice - cause, for example golang prometheus client (and probably other libraries) depends on "github.com/golang/protobuf" 1.4.0 in newer versions - hence right now we cannot just use it's latest version.

kyessenov:
Sent https://github.com/envoyproxy/go-control-plane/pull/322 to update protobuf to 1.4.2. The generated files already assume 1.4.x.

PR #294, opened 2020-04-30
pkg/wellknown: update wellknown const types

stevesloka:
Updates #293 

This updates the wellknown types inside `pgk/wellknown` for Envoy v1.14.0. It also skips the `envoy.cors` update since that's covered in #291 

Signed-off-by: Steve Sloka <slokas@vmware.com>

PR #320, opened 2020-06-26
Add forward upstream cluster wellknown name

EItanya:
Update wellknown names of more filters

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "go.inferGopath": false
+}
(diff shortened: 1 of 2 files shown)


stevesloka:
I think this duplicates https://github.com/envoyproxy/go-control-plane/pull/294

EItanya:
> I think this duplicates #294

Indeed, completely missed that, sorry!

EItanya:
@stevesloka any idea when that other PR will be merged, it seems like it's been open for quite a while. Also, I reopened this one with just one new wellknown name, but the check doesn't seem to be rerunning

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "go.inferGopath": false
```
kyessenov, line 2: Can we remove this file from the commit?

kyessenov:
Folks, you don't have to specify the name to apply the extension. Starting early this year, envoy relies on the type URL of the extension config to select an extension.

PR #322, opened 2020-07-01
update protobuf to v1.4.2 and grpc to v1.27.0

kyessenov:
gRPC update is required with the newer protobuf.


jyotimahapatra:
I'm on my paternity leave and been late to the discussion.

Based on [this](https://github.com/envoyproxy/go-control-plane/blob/master/envoy/api/v2/lds.pb.go#L1-L24), looks like autogen .pb.go files are already using protobuf v2. Is my assumption right?

kyessenov:
Thanks and congratulations @jyotima .

Issue #323, opened 2020-07-13
Add opaque resources for xDS server


kyessenov:
We should allow xDS resources with unknown types in the server. This is needed to support future xDS resource types (and solves the two phase commit problem with adding new xDSes). Fixes #310 as well. cc @rshriram

kyessenov:
Added and being tested.

PR #326, opened 2020-07-17
cache: implement linear versioning and mux strategies

kyessenov:
Adds a cache that is more appropriate for passing opaque resources via xDS. It is a single collection with incremental versions, and it suits EDS pretty well.
Adds muxing routing to combine multiple caches.
Adds an integration test.

This is needed to support new types of discovery services (e.g. extension config discovery service) without baking it into the existing snapshots.


kyessenov:
Fixes #323

pkg/cache/v2/cache.go
```diff
@@ -69,7 +69,6 @@ type Response interface {
 // RawResponse is a pre-serialized xDS response containing the raw resources to
 // be included in the final Discovery Response.
 type RawResponse struct {
-	Response
```
kyessenov, line 72: I don't quite understand the reason for embedding here... Any insight?

kyessenov:
This is mostly additive. Should be safe for updates.

pkg/cache/v2/cache.go
kyessenov: @jessicayuen Is it OK to remove this? There is no explanation why this field is defined and it builds fine without it.

jessicayuen: I believe the intention here was to make it mandatory for `RawResponse` and `PassthroughResponse` to implement the contract in the `Response` interface.  Otherwise, accidental removal of one of the [methods](https://github.com/envoyproxy/go-control-plane/pull/326/files#diff-559e68cd870b75c9dbc9c282d7592466L143) or changes to the `Response` interface would make these structs incompatible with the `CreateWatch` and `Cache` functionalities.

kyessenov: I see. I think the idiomatic way to do that is with `var _ Response = &RawResponse`. Adding an anonymous field would just delegate the calls to the inner Response field, see https://putridparrot.com/blog/anonymous-fields-in-go/ for example.

jessicayuen: I see, ok thanks for the fix!

pkg/cache/v2/mux.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+var _ Cache = &MuxCache{}
+
+func (mux *MuxCache) CreateWatch(request Request) (chan Response, func()) {
+	key := mux.Classify(request)
+	cache, exists := mux.Caches[key]
+	if !exists {
+		value := make(chan Response, 1)
+		value <- RawResponse{Request: request}
```
jessicayuen, line 36: How does the receiver identify that this is a scenario where the request was not able to be classified? It feels like this is an error scenario.

kyessenov: Yeah, I wasn't sure what to return. This will effectively just leave it hanging until the client decides to close it. We may need to change the interface so that the server closes the stream in some cases.

kyessenov: After giving this a bit of thought, returning nil as channel works as intended. It is equivalent to a closed empty channel, so the server will close the stream.

pkg/cache/v2/linear.go
```diff
@@ -0,0 +1,219 @@
...
+
+	// TODO: batch watch closures to prevent rapid updates
+	cache.notifyAll(map[string]struct{}{name: struct{}{}})
+	return nil
+}
+
+func (cache *LinearCache) CreateWatch(request Request) (chan Response, func()) {
+	value := make(chan Response, 1)
+	if request.TypeUrl != cache.typeURL {
+		value <- RawResponse{Request: Request{TypeUrl: cache.typeURL}, Version: strconv.FormatUint(cache.version, 10)}
```
jessicayuen, line 147: Should we be closing the channel here as well? It's unclear to me what implications there are to the receiver if we send back a response with a different type url than the original request.

kyessenov: Good point. Returning a closed channel makes sense here, too.

kyessenov:
@jessicayuen Do you mind re-approving if the latest commit looks good to you?

PR #329, opened 2020-07-24
update ci image

kyessenov:


kyessenov:
Seems like 1.15 takes longer to connect to ADS server. Will need to add readiness.

kyessenov:
This appears to be a bug in Envoy:
```open watch 1 for type.googleapis.com/envoy.service.discovery.v2.Runtime[runtime-1 runtime-0] from nodeID "", version ""```

Envoy is not sending node ID for some reason on the first runtime request.

PR #330, opened 2020-07-26
Help for newcomers

acnodal-tc:
I noticed some comments in issues and PR's about newcomers having trouble figuring out how to use go-control-plane to make their own control planes.  The integration test is a great example, but its primary mission is to be a test, not an example, so it's got a lot of code that gets in the way of understanding how to integrate go-control-plane.

Here are three commits that helped me figure out what's going on:

* better documentation of the many integration test options (especially the ports used since there are a lot of them)
* break the "upstream" HTTP server into its own process, which made it easier for me to build a mental model of what's happening during the tests
* a trivial, minimal example server that newcomers can run easily and copy into their projects.

### DIFF
--- a/internal/example/README.md
+++ b/internal/example/README.md
@@ -1,0 +1,16 @@
+# Example xDS Server
+
+This is an example of a trivial xDS V3 control plane server.  It serves an Envoy configuration that's roughly equivalent to the one used by the Envoy ["Quick Start"](https://www.envoyproxy.io/docs/envoy/latest/start/start#quick-start-to-run-simple-example) docs: a simple http proxy.  You can run the example using the project top-level Makefile, e.g.:
+
+```
+go-control-plane$ make example
+```
+
+The Makefile builds the example server and then runs `build/example.sh` which runs both Envoy and the example server.  The example server serves a configuration defined in `internal/example/resource.go`.  If everything works correctly, you should be able to open a browser to [http://localhost:10000](http://localhost:10000) and see Acnodal.io's website.
+
+## Files
+
+* [main/main.go](main/main.go) is the example program entrypoint.  It instantiates the cache and xDS server and runs the xDS server process.
+* [resource.go](resource.go) generates a `Snapshot` structure which describes the configuration that the xDS server serves to Envoy.
+* [server.go](server.go) runs the xDS control plane server.
+* [logger.go](logger.go) implements the `pkg/log/Logger` interface which provides logging services to the cache.
--- a/internal/example/README.md
+++ b/internal/example/README.md
@@ -6,7 +6,7 @@ This is an example of a trivial xDS V3 control plane server.  It serves an Envoy
 go-control-plane$ make example
 ```
 
-The Makefile builds the example server and then runs `build/example.sh` which runs both Envoy and the example server.  The example server serves a configuration defined in `internal/example/resource.go`.  If everything works correctly, you should be able to open a browser to [http://localhost:10000](http://localhost:10000) and see Acnodal.io's website.
+The Makefile builds the example server and then runs `build/example.sh` which runs both Envoy and the example server.  The example server serves a configuration defined in `internal/example/resource.go`.  If everything works correctly, you should be able to open a browser to [http://localhost:10000](http://localhost:10000) and see Envoy's website.
 
 ## Files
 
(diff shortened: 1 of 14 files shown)


acnodal-tc:
FYI since these are mostly independent commits with a common theme, it will probably be easier to review them commit-by-commit instead of looking at the changed files: https://github.com/envoyproxy/go-control-plane/pull/330/commits

internal/example/README.md
```diff
@@ -0,0 +1,16 @@
+# Example xDS Server
+
+This is an example of a trivial xDS V3 control plane server.  It serves an Envoy configuration that's roughly equivalent to the one used by the Envoy ["Quick Start"](https://www.envoyproxy.io/docs/envoy/latest/start/start#quick-start-to-run-simple-example) docs: a simple http proxy.  You can run the example using the project top-level Makefile, e.g.:
+
+```
+go-control-plane$ make example
+```
+
+The Makefile builds the example server and then runs `build/example.sh` which runs both Envoy and the example server.  The example server serves a configuration defined in `internal/example/resource.go`.  If everything works correctly, you should be able to open a browser to [http://localhost:10000](http://localhost:10000) and see Acnodal.io's website.
```
kyessenov, line 9: Do you think we can use something neutral, e.g. envoyproxy.io?

internal/upstream/main.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+func main() {
+	var (
+		debug        bool
+		upstreamPort uint
+		hello        string
+	)
+
+	flag.BoolVar(&debug, "debug", false, "Use debug logging")
```
kyessenov, line 30: Unused, please remove.

caboteria:
@alecholmez FYI

alecholmez:
Oh awesome I didn't even see this. Will definitely help and contribute

internal/example/README.md
acnodal-tc: Yup, I pushed a fixup commit to change it to www.envoyproxy.io.  LMK if you'd like me to squash the fixup commits and force push.

internal/upstream/main.go
acnodal-tc: Done.

shivam009:
@acnodal-tc @alecholmez 
I have few questions regarding using the control plane :
1) In the main.go i see you are taking NodeID as input, i don't understand this. I might have a management server(control plane) from which multiple envoy instances would be fetching data. So now How can i pass the NodeID of these Envoy instances before hand to generate the snapshot. There should be a mechanism that whenever a new node creates a request from the management server, then we should fetch the Node id and create a snapshot for that node and setSnapshot in our cache. 
2) I am just trying to implement Route discovery service and everything else will be static, hence i tries to run the control plane as RDS, now when i am calling streamRoutes i am able to get the routes from control plane (provided there was a snapshot for the nodeid), but now i want whenever there is a change in rule so lets say whenever there is a change in rule i do a setSnapshot for all the nodes, then how would server stream new routes to all the active streams? (Inside server.go, inside function process, i understand there is var values watches, now i am not sure how to put value to values.routes whenever a route is updated

velcrine:
> @acnodal-tc @alecholmez
> I have few questions regarding using the control plane :
> 
>     1. In the main.go i see you are taking NodeID as input, i don't understand this. I might have a management server(control plane) from which multiple envoy instances would be fetching data. So now How can i pass the NodeID of these Envoy instances before hand to generate the snapshot. There should be a mechanism that whenever a new node creates a request from the management server, then we should fetch the Node id and create a snapshot for that node and setSnapshot in our cache.
> 
>     2. I am just trying to implement Route discovery service and everything else will be static, hence i tries to run the control plane as RDS, now when i am calling streamRoutes i am able to get the routes from control plane (provided there was a snapshot for the nodeid), but now i want whenever there is a change in rule so lets say whenever there is a change in rule i do a setSnapshot for all the nodes, then how would server stream new routes to all the active streams? (Inside server.go, inside function process, i understand there is var values watches, now i am not sure how to put value to values.routes whenever a route is updated

Please somebody answer this. Especially the first point. I don't have the nodeId initially, but as each envoy comes and makes request, I access the ID and update the cache. Sadly, there is no procedure for doing this.

acnodal-tc:
@velcrine I think that callbacks might be what you're looking for. There are a couple of callbacks.go files; one for v2 and one for v3.

velcrine:
Hy ya,  I got this. It works fine. Thanks.

PR #333, opened 2020-07-30
add readiness probe

kyessenov:


pkg/probe/probe.go
```diff
@@ -0,0 +1,85 @@
...
+	StatLDSConfigReload = "listener_manager.lds.config_reload"
+	StatCDSConfigReload = "cluster_manager.cds.config_reload"
+)
+
+var (
+	// General stats URI in prometheus format
+	StatsURI = "/stats/prometheus"
+	// Stats endpoint for reading readiness status (usedonly because some stats default to 0)
+	ReadinessURI = fmt.Sprintf("%s?usedonly&filter=(%s|%s)",
+		StatsURI,
```
yangminzhu, line 46: nit: `StatsURI` is only used here, probably just encode it directly in the string.

pkg/probe/probe.go
```diff
@@ -0,0 +1,85 @@
...
+		return false, err
+	}
+	// server state must be 0
+	if state, exists := metrics[PromStatName(StatServerState)]; !exists {
+		return false, nil
+	} else if len(state.Metric) != 1 || state.Metric[0].Gauge == nil ||
+		state.Metric[0].Gauge.Value == nil || *state.Metric[0].Gauge.Value != 0 {
+		return false, nil
+	}
+	// workers started must be 1
```
yangminzhu, line 77: it must be 1 because the envoy is just started (no traffic)?

pkg/probe/probe.go
```diff
@@ -0,0 +1,85 @@
...
+	// Values described here:
+	// https://www.envoyproxy.io/docs/envoy/v1.15.0/api-v3/admin/v3/server_info.proto#envoy-v3-api-enum-admin-v3-serverinfo-state
+	// LIVE is value 0.
+	StatServerState = "server.state"
+
+	// Boolean indicating whether listeners have been initialized on the workers.
+	StatListenerManagerWorkersStarted = "listener_manager.workers_started"
+
+	// Total number of updates that resulted in a config reload.
+	StatLDSConfigReload = "listener_manager.lds.config_reload"
```
yangminzhu, line 37: These seems not used?

pkg/probe/probe.go
kyessenov: I was thinking of extending it with other stat queries later on.

pkg/probe/probe.go
kyessenov: 1 means "true", so the workers have started.

PR #335, opened 2020-08-05
Incremental xDS

alecholmez:
## Status
- [PASSING] Unit tests are failing :(
- [PASSING] Integration tests are passing!
```
alecholmes@DESKTOP-I0NS4FL:~/Developer/alecholmez/go-control-plane$ make test
--> testing go-control-plane
?       github.com/envoyproxy/go-control-plane/pkg/cache/types  [no test files]
ok      github.com/envoyproxy/go-control-plane/pkg/cache/v2     0.267s
ok      github.com/envoyproxy/go-control-plane/pkg/cache/v3     0.266s
ok      github.com/envoyproxy/go-control-plane/pkg/conversion   0.015s
ok      github.com/envoyproxy/go-control-plane/pkg/log  0.009s
?       github.com/envoyproxy/go-control-plane/pkg/resource/v2  [no test files]
?       github.com/envoyproxy/go-control-plane/pkg/resource/v3  [no test files]
--- FAIL: TestDeltaResponseHandlers (1.00s)
    --- FAIL: TestDeltaResponseHandlers/unknown-type (1.00s)
        delta_test.go:226: got no response
FAIL
FAIL    github.com/envoyproxy/go-control-plane/pkg/server/v2    1.021s
--- FAIL: TestDeltaResponseHandlers (1.00s)
    --- FAIL: TestDeltaResponseHandlers/unknown-type (1.00s)
        delta_test.go:227: got no response
FAIL
FAIL    github.com/envoyproxy/go-control-plane/pkg/server/v3    1.014s
?       github.com/envoyproxy/go-control-plane/pkg/test [no test files]
?       github.com/envoyproxy/go-control-plane/pkg/test/main    [no test files]
?       github.com/envoyproxy/go-control-plane/pkg/test/resource/v2     [no test files]
?       github.com/envoyproxy/go-control-plane/pkg/test/resource/v3     [no test files]
?       github.com/envoyproxy/go-control-plane/pkg/test/v2      [no test files]
?       github.com/envoyproxy/go-control-plane/pkg/test/v3      [no test files]
?       github.com/envoyproxy/go-control-plane/pkg/wellknown    [no test files]
FAIL
make: *** [Makefile:27: test] Error 1
```

This failure is coming from the opaque resource handling and I'm currently looking into this. But incremental is now functional!

### Versions
- Envoy 1.15.0+
- Go 1.11+

### Quickstart
From the root of the project:
```
./build/run_docker.sh make build test
```

To run through the integration test:
```
./build/run_docker.sh make integration
```


alecholmez:
@jessicayuen just reopened this, still a WIP as the integration test is failing right now, looking into the issue though.

pkg/cache/v2/resource.go
```diff
@@ -132,3 +154,25 @@ func GetResourceReferences(resources map[string]types.Resource) map[string]bool
 	}
 	return out
 }
+
+// HashResources will take a list of resources and create a SHA256 hash sum out of the marshaled bytes
+func HashResources(resources []types.Resource) (string, error) {
+	var source []byte
```
joshua-rutherford, line 160: I think it might be cleaner and more efficient to stream the serialized bytes to the hasher as they are serialized.

```
hasher := sha256.New()

buffer := proto.NewBuffer(nil)
buffer.SetDeterministic(true)

    err := buffer.Marshal(resource)
    if err != nil {
        return "",  err
    }

    hasher.Write(buffer.Bytes())

    buffer.Reset()   
}

return hex.EncodeToString(hasher.Sum(nil)), nil
```
This is analogous to reading a file a chunk at a time vs. reading a file into memory.  Hashers don't need all the bytes at once, they just need their current state which is the result of all bytes that have passed through.

pkg/cache/v2/resource_test.go
```diff
@@ -37,6 +37,7 @@ const (
 
 var (
 	testEndpoint = resource.MakeEndpoint(clusterName, 8080)
+
```
joshua-rutherford, line 40: Superfluous white space addition.

pkg/cache/v2/simple.go
```diff
@@ -1,3 +1,5 @@
+// Code generated by create_version. DO NOT EDIT.
```
joshua-rutherford, line 1: This shouldn't be here I think.

pkg/cache/v2/simple.go
```diff
@@ -119,7 +132,12 @@ func (cache *snapshotCache) SetSnapshot(node string, snapshot Snapshot) error {
 				if cache.log != nil {
 					cache.log.Debugf("respond open watch %d%v with new version %q", id, watch.Request.ResourceNames, version)
 				}
-				cache.respond(watch.Request, watch.Response, snapshot.GetResources(watch.Request.TypeUrl), version)
+				cache.respond(
```
joshua-rutherford, line 135: Probably not worth the change since its functionally equivalent and doesn't seem to violate linting).

pkg/cache/v2/simple_delta.go
```diff
@@ -0,0 +1,297 @@
+// Copyright 2018 Envoyproxy Authors
```
joshua-rutherford, line 1: Curious if entirely new files shouldn't have the correct copyright year on them.

pkg/cache/v2/simple_delta.go
```diff
@@ -0,0 +1,297 @@
...
+
+	for key, value := range deltaState {
+		mb[key] = value
+	}
+
+	// Check our diff map to see what has changed
+	// Even is an underlying resource has changed we need to update the diff
+	for key, value := range resources {
+		if _, found := mb[key]; !found {
+			cache.log.Debugf("found new key: %s", key)
```
joshua-rutherford, line 173: I think this and the following access of `cache.log` belong inside a `nil` check.

pkg/cache/v2/simple_delta.go
```diff
@@ -0,0 +1,297 @@
...
+	return nil
+}
+
+// difference returns the elements in `a` that aren't in `b`.
+// TODO: SLOW this will need to be revisited
+func (cache *snapshotCache) checkState(resources, deltaState map[string]types.Resource) map[string]types.Resource {
+	mb := make(map[string]types.Resource, len(deltaState))
+	diff := make(map[string]types.Resource, 0)
+
+	for key, value := range deltaState {
```
joshua-rutherford, line 165: Is there a reason you are cloning the `deltaState`?  You don't mutate `mb` below so the only reason is if you think `deltaState` is mutated in another thread.

pkg/cache/v2/simple_delta.go
```diff
@@ -0,0 +1,297 @@
...
+		mb[key] = value
+	}
+
+	// Check our diff map to see what has changed
+	// Even is an underlying resource has changed we need to update the diff
+	for key, value := range resources {
+		if _, found := mb[key]; !found {
```
joshua-rutherford, line 172: There is some inefficiency here that is easy to remove by only fetching the value once:

```
for key, value := range resources {
    resource, found := deltaState[key]
    if !found || resource != value {
        diff[key] = value
    }
}
```
Won't save much assuming the map implementation is decent.

pkg/cache/v2/simple_delta.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+
+	return nil
+}
+
+// difference returns the elements in `a` that aren't in `b`.
+// TODO: SLOW this will need to be revisited
+func (cache *snapshotCache) checkState(resources, deltaState map[string]types.Resource) map[string]types.Resource {
+	mb := make(map[string]types.Resource, len(deltaState))
+	diff := make(map[string]types.Resource, 0)
```
joshua-rutherford, line 163: If speed is the issue, attempt to minimize future memory allocations by setting this to the max size possible (i.e., `len(deltaState)`).  Not sure on the actual implementation on map and what this really buys you but if you are worried about speed, not memory that would be the decision.

pkg/cache/v2/simple_delta.go
```diff
@@ -0,0 +1,297 @@
...
+// Unscrubscribe will remove resources from the tracked list in the management server when received from a client
+func (cache *snapshotCache) unsubscribe(resources []string, deltaState map[string]Resources) map[string]Resources {
+	// here we need to search and remove from the current subscribed list in the snapshot
+	for _, items := range deltaState {
+		for i := 0; i < len(resources); i++ {
+			delete(items.Items, resources[i])
+		}
+	}
+
+	return deltaState
```
joshua-rutherford, line 193: Not sure why you need to return `deltaState` here as you have mutated the input `deltaState` so the caller can access the updated state anyway.

Makefile
```diff
@@ -35,15 +47,16 @@ check_version_dirty:
...
 #-----------------
-.PHONY: $(BINDIR)/test $(BINDIR)/upstream integration integration.ads integration.ads.v3 integration.xds integration.xds.v3 integration.rest integration.rest.v3 integration.ads.tls integration.xds.mux.v3
+.PHONY: $(BINDIR)/test $(BINDIR)/upstream integration integration.ads integration.ads.v3 integration.xds integration.xds.delta integration.xds.v3 integration.rest integration.rest.v3 integration.ads.tls integration.xds.mux.v3
 
 $(BINDIR)/upstream:
 	@go build -race -o $@ internal/upstream/main.go
 
 $(BINDIR)/test:
-	@go build -race -o $@ pkg/test/main/main.go
+	echo "Building test binary"
```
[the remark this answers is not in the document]
alecholmez: Thank you for the catch

Makefile
```diff
@@ -35,15 +47,16 @@ check_version_dirty:
...
+	$(TEST_BIN_FLAGS) go build -race -a -tags netgo -ldflags '-w -extldflags "-static"' -o $@ pkg/test/main/main.go
```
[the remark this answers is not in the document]
alecholmez: Yup good catch, removed

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/envoyproxy/go-control-plane
 
-go 1.11
+go 1.13
```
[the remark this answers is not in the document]
alecholmez: Good call. This actually should be downgraded anyways, upstream went back to `go 1.11` in their module

internal/example/main/main.go
```diff
@@ -70,6 +70,6 @@ func main() {
 	// Run the xDS server
 	ctx := context.Background()
 	cb := &testv3.Callbacks{Debug: l.Debug}
-	srv := serverv3.NewServer(ctx, cache, cb)
+	srv := serverv3.NewServer(ctx, cache, cb, l)
```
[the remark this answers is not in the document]
alecholmez: Understood, I did it because I needed to add some logging within the server delta handlers but if that needs to be removed I can take it out.

pkg/cache/v2/resource.go
alecholmez: Good call, I've been debating taking this out of the resource file now because it isn't entirely needed. I originally had the idea that if users want to hash their resources and use that as their version for the new deltaStream they could use this. @kyessenov or @jessicayuen might want to weigh in on that

pkg/cache/v2/resource_test.go
alecholmez: Thank you thank you

pkg/cache/v2/simple_delta.go
alecholmez: Ah yes good call there, I was originally thinking I didn't like that fact this ran `O(n)` and wanted that to be more efficient but I told myself I can worry about getting this 100% functional then optimize later

pkg/cache/v2/simple_delta.go
alecholmez: If I don't copy it here it could get mutated by another thread or stream. I originally tried to not copy it and was seeing some odd behavior like things missing or not being detected.

pkg/cache/v2/simple_delta.go
alecholmez: Ah yes thank you, I did the other way just to have some better log statements but this is more efficient

pkg/cache/v2/simple_delta.go
alecholmez: Yup, I ended up consolidating to one and adding the check back in

pkg/cache/v2/simple_delta.go
alecholmez: Yup thank you, removed that

fxposter:
I wonder - maybe it would be better to split incremental XDS implementation into 2 PRs: the one that handles "server" part, adding all necessary interfaces for "cache" side but with empty or panic-ing empty implementations, and followup PR with simple cache (if it's worth implementing this feature in simple cache at all - I don't actually think that simple cache in it's current form fits incremental xds at all). WDYT?

fxposter:
BTW, same "split" can (and in my opinion should) be done for v2/v3 versions. IE: first PR may implement incremental XDS only for v2 or v3 and followup one will be the backport to another version.

internal/example/main/main.go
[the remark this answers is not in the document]
joshua-rutherford: Yeah, my point isn't against logging, my point was that this is a signature change to a "public" method (i.e., one in the `pkg` not `internal` directory) in a library.  As such, this is a breaking change.  Hopefully the rest of this functionality can be considered an added capability, but this one is clearly a change and will break downstream builds (likely all of them).

I realize this repository is not yet at 1.0.0 and this may not be a concern yet but the logger injection is not, strictly speaking, a requirement for incremental xDS.

pkg/cache/v2/simple_delta.go
joshua-rutherford: Hmm, and now it looks like it's not being used at all below?

alecholmez: Ah yes thank you, I was confused for a second but you're right, it wasn't being used `mb` was basically useless. Deleted.

alecholmez:
@fxposter I had a thought as well, I would leave that determination up to the maintainers. I can certainly break this into 2 if needed, it would probably be easier to review that way. 

As far as the split for the versions, I'm not entirely sure that's possible with the way CI is setup. All you have to do to make v3 is run `make create_version` which is done for CI testing anyways so I just execute that script on changes and push regardless. 

@kyessenov @jessicayuen any input? I can break this into multiple PRs if need be, but then the feature won't be merged with full implementation right away. Not sure how that might interfere with the xDS relay work as well.

pkg/cache/v2/cache.go
```diff
@@ -131,30 +198,88 @@ func (r RawResponse) GetDiscoveryResponse() (*discovery.DiscoveryResponse, error
 	}, nil
 }
 
+func (r RawDeltaResponse) GetDeltaDiscoveryResponse() (*discovery.DeltaDiscoveryResponse, error) {
+	if r.isResourceMarshaled {
+		return r.marshaledResponse, nil
```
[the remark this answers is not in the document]
alecholmez: Yup you're right thats a bad case of copy pasta on my part. The upstream `GetDiscoveryResponse()` dues the same thing. I'll fix this thank you for the catch

pkg/cache/v2/cache.go
```diff
@@ -1,3 +1,5 @@
+// Code generated by create_version. DO NOT EDIT.
```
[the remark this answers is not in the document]
alecholmez: It is, i'm not sure why that gets their either, it seems to always happen when I try to get rid of it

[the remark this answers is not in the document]
alecholmez: Got rid of it

pkg/cache/v2/cache.go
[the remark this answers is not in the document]
alecholmez: Addressed.

alecholmez:
Hi all just wanted to respond on this PR so everyone knows I'm still active:

I ended up rewriting the caching/server logic to better handle versions and it seems to be performing better as well as faster. I now have the integration test mostly passing, meaning envoy is receiving configs and proxying as it should, but about 3/4 of the way through envoy seg faults, I'm looking into this error now but will hopefully like to clean up the code and have this PR ready to go in!

pkg/server/v2/server.go
```diff
@@ -67,6 +65,10 @@ type Callbacks interface {
 	OnStreamRequest(int64, *discovery.DiscoveryRequest) error
 	// OnStreamResponse is called immediately prior to sending a response on a stream.
 	OnStreamResponse(int64, *discovery.DiscoveryRequest, *discovery.DiscoveryResponse)
+
+	OnStreamDeltaRequest(int64, *discovery.DeltaDiscoveryRequest) error
+	OnStreamDeltaResponse(int64, *discovery.DeltaDiscoveryRequest, *discovery.DeltaDiscoveryResponse)
```
fxposter, line 70: I've tried implementing delta xds and I guess this one is not correct. In delta XDS the response cannot necessarily be attributed to particular request, hence IMO - this one should be without the "request" part

joshua-rutherford: So you can definitely correlate the initial response to the initial request.  Is your point that you might receive multiple requests on the same stream and therefore can't correctly attribute a response to a specific the request?  If so, I think that makes sense especially since the subsequent requests may simply add (or remove) a few new resource subscriptions therefore attributing a response that is not called out in those requests would be a bit odd. @alecholmez, do you concur?

fxposter: Yes, the sotw api has both, because to any request there is a corresponding response. For delta api requests are for modification of internal state and that state is the trigger for doing responses, hence you cannot attribute response to particular request.
That being said - if we drop the request here - we would not be able to easily find out what node this response belongs to.

alecholmez: I do concur that makes sense. I guess I could pull the request object out of that callback. Is the general consensus to remove that? I'd rather keep it in as there is some useful information in that request object, though you are both correct, a strict correlation between the initial request and following responses wouldn't make sense.

alecholmez:
closing this PR..... git history got super messed up. reopened here: https://github.com/envoyproxy/go-control-plane/pull/355

PR #338, opened 2020-08-10
marshaledResponse not use

kkHAIKE:
oops :smirk:

### DIFF
--- a/pkg/cache/v2/cache.go
+++ b/pkg/cache/v2/cache.go
@@ -122,12 +122,12 @@ func (r RawResponse) GetDiscoveryResponse() (*discovery.DiscoveryResponse, error
 		}
 	}
 
-	r.isResourceMarshaled = true
 	r.marshaledResponse = &discovery.DiscoveryResponse{
 		VersionInfo: r.Version,
 		Resources:   marshaledResources,
 		TypeUrl:     r.Request.TypeUrl,
 	}
+	r.isResourceMarshaled = true
 
 	return r.marshaledResponse, nil
 }
--- a/pkg/cache/v2/cache.go
+++ b/pkg/cache/v2/cache.go
@@ -123,12 +123,13 @@ func (r RawResponse) GetDiscoveryResponse() (*discovery.DiscoveryResponse, error
 	}
 
 	r.isResourceMarshaled = true
-
-	return &discovery.DiscoveryResponse{
+	r.marshaledResponse = &discovery.DiscoveryResponse{
 		VersionInfo: r.Version,
 		Resources:   marshaledResources,
 		TypeUrl:     r.Request.TypeUrl,
-	}, nil
+	}
+
+	return r.marshaledResponse, nil
 }
 
 // GetRequest returns the original Discovery Request.
--- a/pkg/cache/v3/cache.go
+++ b/pkg/cache/v3/cache.go
@@ -124,12 +124,13 @@ func (r RawResponse) GetDiscoveryResponse() (*discovery.DiscoveryResponse, error
 	}
 
 	r.isResourceMarshaled = true
-
-	return &discovery.DiscoveryResponse{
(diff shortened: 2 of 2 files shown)


pkg/cache/v2/cache.go
```diff
@@ -123,12 +123,13 @@ func (r RawResponse) GetDiscoveryResponse() (*discovery.DiscoveryResponse, error
 	}
 
 	r.isResourceMarshaled = true
```
joshua-rutherford, line 125: There is a race condition here were for some period of time `r.isResourceMarshaled` is `true` and the value of `r.marshaledResponse` is `nil`. This all needs to go in a mutex or potentially just set `r.isResourceMarshaled` after setting the value of `r.marshaledResponse` and accept that occasionally you will marshal a few times but no always.

joshua-rutherford: Also, note that as per #339 this will still not fix the lack of caching since on subsequent invocations the `r.isResourceMarshaled` is always `false`.

kyessenov: Good point about the race condition. The server is concurrent, so there could be multiple routines operating on the same cache response. I wonder if mutex is going to lead to a contention (maybe it's best to pre-compute it using Passthrough response). @jessicayuen @jyotimahapatra any thoughts on this?

joshua-rutherford: I also opened #340 with a test and more aggressive fix.  @alecholmez noted that there are mutexes being copied around there so we'd need to fix that in mine as he'd done in #335 now.

joshua-rutherford: I too would avoid a mutex if possible for this and just live with the potential for the occasional wasted cycles but assuming that you'll marshal everything once precomputing makes much more sense.

jyotimahapatra: This makes sense to me. However, I think go might reorder statements and can still cause subtle bugs. I guess using an atomic store/load is also necessary. WDYT?

jyotimahapatra: Another possible way could be to remove the dependency on the extra `isResourceMarshaled` variable and base the check on the `nil`ness of `r.marshaledResponse`. I think this could remove the concern about statement reordering.

joshua-rutherford: The nil check works for me.  Of course this still doesn’t change the fact that it will always be nil until this method operates on a pointer of a RawMessage not a RawMessage value.

kkHAIKE: i think change order is gool...

alecholmez: I'm in favor of the `nil` check with the pointer upgrades, I can make that PR

joshua-rutherford: hehe, made the change in mine this morning and never pushed.  #340

joshua-rutherford:
@kyessenov, I'm a bit perplexed by this as it doesn't actually do what is intended as per point number two of issue #339.  I think this is better than it was but it's still just wasted cycles as the cached value is garbage collected is unreferenced and free for garbage collection once the method exits.  Totally possible that I am I missing something as I've only been looking at this codebase for a few days.

kyessenov:
Oh, you're right @joshua-rutherford. Can you please send make the receiver a pointer? I think we really need a test that caching works per resource.

joshua-rutherford:
I opened this #340 with that.  I can understand if it's not ideal since it does potentially break things for downstream consumers. @alecholmez was there anything else needed in that to prevent mutex copying?  I didn't see what you did in #335 for that.

alecholmez:
@joshua-rutherford Nope I think you covered all the bases. I made the same exact changes in mine just on the delta end

jyotimahapatra:
I think #340 is capturing the changes well.

kyessenov:
This was addressed in #340, so I'm going to close this. Thanks for a PR!

PR #341, opened 2020-08-10
cache: add version prefix

kyessenov:
Envoy persists versioninfo across re-connections to the control plane servers. If the version is auto-generated as in the linear cache, we should use unique prefixes to distinguish clients joining from other instances of the control plane.
This changes adds a version prefix option.


jyotimahapatra:
I missed looking at it. Doing it now.

Issue #342, opened 2020-08-11
Example fails to proxy http://localhost:1000


danehans:
The [example](https://github.com/envoyproxy/go-control-plane/blob/master/internal/example/README.md) fails due to Envoy [preferring IPv6](https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/cluster.proto#enum-cluster-dnslookupfamily) as the address family for cluster DNS lookups. Since the upstream (www.envoyproxy.io) resolves to an IPv6 address [1], Envoy attempts to use IPv6 instead of IPv4 for the upstream but fails to create the connection with error `immediate connection error: 65`.

__Full Debug Log__
```
[2020-08-10 20:41:46.845][12543336][debug][http] [external/envoy/source/common/http/conn_manager_impl.cc:261] [C2] new stream
[2020-08-10 20:41:46.845][12543336][debug][http] [external/envoy/source/common/http/conn_manager_impl.cc:808] [C2][S13775225100775145700] request headers complete (end_stream=true):
':authority', 'localhost:10000'
':path', '/favicon.ico'
':method', 'GET'
'connection', 'keep-alive'
'user-agent', 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/84.0.4147.105 Safari/537.36'
'accept', 'image/webp,image/apng,image/*,*/*;q=0.8'
'sec-fetch-site', 'same-origin'
'sec-fetch-mode', 'no-cors'
'sec-fetch-dest', 'image'
'referer', 'http://localhost:10000/'
'accept-encoding', 'gzip, deflate, br'
'accept-language', 'en-US,en;q=0.9'

[2020-08-10 20:41:46.845][12543336][debug][http] [external/envoy/source/common/http/conn_manager_impl.cc:1377] [C2][S13775225100775145700] request end stream
[2020-08-10 20:41:46.845][12543336][debug][router] [external/envoy/source/common/router/router.cc:426] [C2][S13775225100775145700] cluster 'example_proxy_cluster' match for URL '/favicon.ico'
[2020-08-10 20:41:46.845][12543336][debug][router] [external/envoy/source/common/router/router.cc:583] [C2][S13775225100775145700] router decoding headers:
':authority', 'www.envoyproxy.io'
':path', '/favicon.ico'
':method', 'GET'
':scheme', 'http'
[...]

PR #343, opened 2020-08-11
Sets cluster DNS lookup to IPv4 only

danehans:
Sets the example cluster DNS lookup family to IPv4 only.

Fixes https://github.com/envoyproxy/go-control-plane/issues/342

/cc @acnodal-tc


kyessenov:
Please fix DCO. The test failure is a flake that I have a PR in progress.

danehans:
@kyessenov PTAL, I think the PR is ready to be tagged.

PR #344, opened 2020-08-11
server: fix handling of opaque resource types

kyessenov:
- fix nil error with nil cancel
- fix muxing of goroutines (double closure of response channel)


jyotimahapatra:
@kyessenov The changes make sense. Can you describe a bit about the issues which necessitate the changes. I'm coming back from paternity leave and have lost some context in the last few months.

kyessenov:
There were two problems with the special handling of extra type URLs (not the standard E/C/R/L/SDS):
- if cancel func is returned as nil, the defer in the server would call on nil and crash
- because watches need to be multiplexed for those types and select does not work over dynamic sets, it's fairly tricky to handle errors on those watches correctly.

jyotimahapatra:
LGTM!

kyessenov:
@jyotimahapatra do you mind approving again? there was a merge conflict.

kyessenov:
Issue: #323

PR #345, opened 2020-08-21
[refactoring] Prepare for incremental xds

fxposter:
Given the fact that incremental XDS PR (#335) is very large and that it requires changes in a lot of places - I propose this refactoring which basically extracts REST and SOTW parts into separate servers, which live in corresponding packages, alongside with splitting the responsibilities of all those servers

- REST and SOTW (and incremental in future) servers will implement corresponding protocols and handle nonces.
- generic server will mostly handle the GRPC interfaces and pass the protocol part to corresponding rest/sotw/incremental servers

WDYT?


pkg/server/v2/server.go
```diff
@@ -50,28 +47,13 @@ type Server interface {
...
-	OnStreamRequest(int64, *discovery.DiscoveryRequest) error
-	// OnStreamResponse is called immediately prior to sending a response on a stream.
-	OnStreamResponse(int64, *discovery.DiscoveryRequest, *discovery.DiscoveryResponse)
-	// OnFetchRequest is called for each Fetch request. Returning an error will end processing of the
-	// request and respond with an error.
-	OnFetchRequest(context.Context, *discovery.DiscoveryRequest) error
-	// OnFetchResponse is called immediately prior to sending a response.
-	OnFetchResponse(*discovery.DiscoveryRequest, *discovery.DiscoveryResponse)
+	rest.Callbacks
+	sotw.Callbacks
```
alecholmez, line 56: So i'm not opposed to this, but we need to keep in mind that we are breaking the existing external API so if we merge this, all pre-existing users of go-control-plane will need to update their logic. My incremental PR only adds to functions into the `Callbacks interface{}`: `OnStreamDeltaRequest()` and `OnStreamDeltaResponse()`. I'm good with the change but this is breaking so a release will need to be cut

@kyessenov what do you think?

pkg/server/v2/server.go
```diff
@@ -136,381 +118,20 @@ func (c CallbackFuncs) OnFetchResponse(req *discovery.DiscoveryRequest, resp *di
...
-	for _, cancel := range values.cancellations {
-		if cancel != nil {
-			cancel()
-		}
-	}
-	for _, terminate := range values.terminations {
-		close(terminate)
-	}
+func NewServerAdvanced(restServer rest.Server, sotwServer sotw.Server) Server {
+	return &server{rest: restServer, sotw: sotwServer}
```
alecholmez, line 125: Why a wrapper constructor here? I'd prefer something like a functional opts pattern here if we're going to keep the separate callback functionality

```go
type ServerOpts func(*Server)

func WithSotwCallbacks(callbacks sotw.Callbacks) ServerOpts {
	return func(s *Server) {
		s.SotwCallbacks = callbacks
	}
}

func WithRestCallbacks(callbacks rest.Callbacks) ServerOpts {
	return func(s *Server) {
		s.RestCallbacks = callbacks
	}
}

func NewServer(ctx, config, WithSotwCallbacks(), WithRestCallbacks()) {
... do our opts processing here
}
```

Just in case you haven't seen the pattern before here's a good example of how to do so:
https://www.sohamkamani.com/golang/options-pattern/

pkg/server/v2/server.go
fxposter: Why, the interface by itself remains the same.

alecholmez: Oh oops good point I thought you changed the underlying callbacks too, i'll remove the comment sorry about that!

pkg/server/v2/server.go
fxposter: It doesn't make any sense. Server by itself does not need any ctx, config, cause server itself consists of 2 extracted "subservers". There is an existing API of NewServer that we cannot break and it doesn't make any sense to extend it.

My options were - either this new method or expose `server` type and it's fields externally

joshua-rutherford: I concur with @fxposter that maintaining the signature of `NewServer` means fewer breakages so should be preferred.

That said, a functional opts pattern, if desired, could be used in the `NewServerAdvanced` implementation but I think it would make most sense if we expect that servers would be instantiated without one or more of the specific server implementations (i.e., REST, SotW or Delta).

pkg/server/v2/server.go
```diff
@@ -50,28 +47,13 @@ type Server interface {
 
 	// Fetch is the universal fetch method.
 	Fetch(context.Context, *discovery.DiscoveryRequest) (*discovery.DiscoveryResponse, error)
-	// StreamHandler is the universal stream method.
-	StreamHandler(stream Stream, typeURL string) error
```
joshua-rutherford, line 54: I might be missing it, but without `StreamHandler` defined here `Server` implementation are only required to implement the REST `Server` interface, not the `SotW` interface.  It is harmless at present but seems like an oversight.  It seems you should require both or neither.

pkg/server/v2/server.go
alecholmez: @fxposter I see what you're saying. Unless we choose to break the original NewServer function but I don't know if the maintainers want to do that. If we want I could make this change in my incremental PR since it's already breaking API changes.

alecholmez: Agreed @joshua-rutherford my reasoning behind the functional opts was to have less code needed to create the needed servers since some users might not want all the server types.

pkg/server/v2/server.go
fxposter: In reality I would prefer to have neither :) But that breaks ability to use `Server` in `gateway.go`. Probably I can just add sotw.Server and rest.Server here and leave that (otherwise not necessarily needed, IMO) functionality as is.

fxposter: Restored the original behaviour by adding sotw.Server and rest.Server here as well. We can drop that if that is not needed in later PRs. rest.Server (IMO) should be there due to `gateway.go`, while sotw.Server is a different thing - when we add incremental.Server - should we add it there? Cause it requires all other implementors of this server.Server interface to provide it.

alecholmez: I wouldn't be opposed. The only thing about incremental is it can't use the same StreamHandler as defined now because it uses a different request/response object: `DeltaDiscoveryRequest` and `DeltaDiscoveryResponse`

pkg/server/v2/server.go
```diff
@@ -136,381 +118,20 @@ func (c CallbackFuncs) OnFetchResponse(req *discovery.DiscoveryRequest, resp *di
...
-	marshalledResponse, err := resp.GetDiscoveryResponse()
-	if err != nil {
-		return nil, err
-	}
-
-	return marshalledResponse, nil
-}
-
-// process handles a bi-di stream request
-func (s *server) process(stream Stream, reqCh <-chan *discovery.DiscoveryRequest, defaultTypeURL string) error {
```
kyessenov, line 245: Can you do git mv v2/server.go to v2/sotw/server.go? It's very difficult to see what changed with the code.

fxposter: I can try, but I really doubt it would help. Basically v2/server.go was split into sotw/v2/server.go and rest/v2/server.go without *code* changes, only methods moved from struct in v2/server.go to implementations of corresponding protocol.

kyessenov: I think it would help. sotw has over 300 lines of code, while rest is just 50. It'd help resolving merge conflicts (if any happen) and preserve git history.

fxposter: Tried doing that. Realized that it is not about rest/server.go, it's about the stuff that in current PR is left in original server/server.go - CallbackFuncs, envoy/api methods/etc.

fxposter: So we either save history for server/server.go or sotw/server.go

fxposter: https://github.com/envoyproxy/go-control-plane/pull/350 another PR, we should chose which to leave :)

fxposter: @kyessenov ping

kyessenov:
Thanks for the review folks. The alternative version with git history would be preferable for maintainability.

kyessenov:
Merged the alternative version.

PR #347, opened 2020-08-25
docs: fix a few doc nit-picks

acnodal-tc:
The recent improvements to the docs are great but there are a few small nits that trigger my pedantry.


alecholmez:
@acnodal-tc thanks for all these!!! Should've proof read on my part, embarrassing! While you have the PR open, I actually noticed one of the links on the front page in the docs folder is actually broken, would you mind fixing that? Thanks again for all the fixups!

PR #350, opened 2020-08-29
refactor: split server into rest and sotw

fxposter:
Code is the same as in #345, but using git mv to save history for SOTW implementation


fxposter:
cc @kyessenov

kyessenov:
Sorry for the delay. I don't see that sotw/v2/server.go is actually moved. If you want, I can try to do it myself, and see what's wrong?

fxposter:
https://github.com/envoyproxy/go-control-plane/pull/350/commits/580ddbc07daee6e99aa1393f4bc4ee03c81df6e5 - isn't this what you needed?..

kyessenov:
Thanks. Github just makes it hard to see. I manually checked with `git blame`.

pkg/server/rest/v2/server.go
```diff
@@ -0,0 +1,49 @@
+package rest
```
kyessenov, line 1: Can you add an attribution comment to Envoyproxy Authors for consistency and a one line package comment for godoc?

pkg/server/sotw/v2/server.go
```diff
@@ -0,0 +1,404 @@
+package sotw
```
kyessenov, line 1: Same here. Please add a package comment and top-level comment.

kyessenov:
Thanks, I took a look, and we can merge it after adding package comments. Sorry about the delay!

kyessenov:
Btw, a trick to duplicate a file in github with history is to `git mv src dst; git checkout HEAD^ src`.

kyessenov (approved):

fxposter:
🎉

rene-m-hernandez:
I'm curious to know what the motivation was to add a REST interface? I tried tracking this ticket back to the source issue and it seems to be related to motivated by incremental xDS; but, I don't believe REST is a requirement for such

kyessenov:
REST is the original xDS before SotW and Delta. I think it's here for backwards compatibility with old clients.

PR #352, opened 2020-09-13
cache: log ignoring watch at warn level

jaricftw:
https://github.com/envoyproxy/go-control-plane/pull/250 caused quite some surprises which suppressed several
useful "info" logs on watch registration/response in the cache.

While it's debatable whether we should bring all of them back, at least the log on rejecting the ADS watch
should be logged at a higher level -- we've seen a few outages caused by a combination of that and the Envoy data plane
failing to establish new watches, leading to missing updates. So the log was super useful in identifying the root cause.

Log it at warn level, because it's not frequent and potentially can cause above issues.

### DIFF
--- a/pkg/cache/v2/simple.go
+++ b/pkg/cache/v2/simple.go
@@ -326,7 +326,7 @@ func (cache *snapshotCache) respond(request *Request, value chan Response, resou
 	if len(request.ResourceNames) != 0 && cache.ads {
 		if err := superset(nameSet(request.ResourceNames), resources); err != nil {
 			if cache.log != nil {
-				cache.log.Debugf("ADS mode: not responding to request: %v", err)
+				cache.log.Warnf("ADS mode: not responding to request: %v", err)
 			}
 			return
 		}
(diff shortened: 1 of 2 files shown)


jaricftw:
build fails with
> error: Make sure to not edit the auto generated files in pkg/cache/v3
But this PR is not editing that, not sure why it fails

alecholmez:
Thanks for the catch! I think that's a fair argument to the logging. The reason I had originally moved that so down low in the logging stack was there can be a lot of log pollution when running at scale making things hard to read so I just made an effort to quiet things down. 

Also for CI, run `make create_version` then push again, I think it's complaining that v3 wasn't also updated

freeloyerkick:
bump, can I get a review :)

jaricftw:
bump for a review, thanks!

jaricftw:
Bump since the PR is marked as stale. We have been running with a patch with the PR for over half an year right now with a few thousands of Envoy nodes in production, the log has not been noisy.

mattklein123:
Can you merge main and we can get this merged? Thanks.

jaricftw:
@mattklein123 thanks, done!

mattklein123 (approved):
Thanks!

Issue #59, opened 2018-03-12
Ordering on xDS responses


kyessenov:
In ADS mode, it is preferable to respond to xDS queries in the following order, to match the topological ordering of cross-references between xDS resources:
1. CDS
2. EDS
3. LDS
4. RDS

We should support sequencing the responses in this order. This is mostly applicable to ADS mode, since in xDS mode, snapshots may be partial (e.g. only RDS).

yskopets:
@kyessenov Hi Kuat!

May I wonder why this issue hasn't got any attention since it was opened ? Isn't it a well-known issue that is sitting there to happen ?

For the context, I've recently described the issue I'm facing on #control-plane-dev Slack: 

---

I’ve run into an issue where `envoy` keeps updated `cluster` in a `warming` state for indefinite amount of time.

As far as I understand, such behaviour is explained in the `Resource warming` section of the `xDS Protocol` (https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#resource-warming):

> Clusters and Listeners go through warming before they can serve requests. This process happens both during Envoy initialization and when the Cluster or Listener is updated. Warming of Cluster is completed only when a ClusterLoadAssignment response is supplied by management server. … Management server is expected to provide the EDS/RDS updates during warming. If management server does not provide EDS/RDS responses, Envoy will not initialize itself during the initialization phase and the updates sent via CDS/LDS will not take effect until EDS/RDS responses are supplied.

In my case, while `Cluster` definition is updated, endpoints do not change. That’s why my `go-control-plane`-based control plane doesn’t send a response to `EDS` request.

Am I interpreting `envoy` behaviour correctly ? Shouldn’t this case be handled by `go-control-plane` ?

kyessenov:
You should make sure the snapshot is consistent if you use ADS mode. That means every cluster has an endpoint assignment (there is a helper function `Consistent()` to help with that).

Keep in mind, that most production xDS servers are only eventually consistent. That means they generally expect EDS server to catch up to CDS server and tolerate partially broken config in envoys (which can sometimes lead to 5xx/4xx). You can simulate that behavior in go-control-plane by not using ADS mode.

The entire sequencing is complex, so I simply didn't have time to do it right. It appears folks are fine with eventual consistency for now.

yskopets:
Thank you!

PR #353, opened 2020-09-15
Order responses when setting snapshot to avoid traffic drops

fxposter:
According to https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#eventual-consistency-considerations

### DIFF
--- a/pkg/cache/v2/simple.go
+++ b/pkg/cache/v2/simple.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/envoyproxy/go-control-plane/pkg/cache/types"
 	"github.com/envoyproxy/go-control-plane/pkg/log"
+	"github.com/envoyproxy/go-control-plane/pkg/resource/v2"
 )
 
 // SnapshotCache is a snapshot-based cache that maintains a single versioned
@@ -102,8 +103,16 @@ func NewSnapshotCache(ads bool, hash NodeHash, logger log.Logger) SnapshotCache 
 	}
 }
 
+type responseInfo struct {
+	id      int64
+	watch   ResponseWatch
+	version string
+}
+
 // SetSnapshotCache updates a snapshot for a node.
 func (cache *snapshotCache) SetSnapshot(node string, snapshot Snapshot) error {
+	toRespond := map[string][]responseInfo{}
+
 	cache.mu.Lock()
 	defer cache.mu.Unlock()
 
@@ -116,16 +125,27 @@ func (cache *snapshotCache) SetSnapshot(node string, snapshot Snapshot) error {
 		for id, watch := range info.watches {
 			version := snapshot.GetVersion(watch.Request.TypeUrl)
 			if version != watch.Request.VersionInfo {
-				if cache.log != nil {
-					cache.log.Debugf("respond open watch %d%v with new version %q", id, watch.Request.ResourceNames, version)
-				}
-				cache.respond(watch.Request, watch.Response, snapshot.GetResources(watch.Request.TypeUrl), version)
+				toRespond[watch.Request.TypeUrl] = append(toRespond[watch.Request.TypeUrl], responseInfo{
+					id:      id,
+					watch:   watch,
+					version: version,
+				})
 
 				// discard the watch
 				delete(info.watches, id)
 			}
 		}
 		info.mu.Unlock()
+	}
+
+	for _, typeUrl := range resource.TypeUrls {
+		resources := snapshot.GetResources(typeUrl)
+		for _, info := range toRespond[typeUrl] {
+			if cache.log != nil {
+				cache.log.Debugf("respond open watch %d%v with new version %q", info.id, info.watch.Request.ResourceNames, info.version)
+			}
+			cache.respond(info.watch.Request, info.watch.Response, resources, info.version)
+		}
 	}
 
 	return nil
--- a/pkg/cache/v2/simple_test.go
+++ b/pkg/cache/v2/simple_test.go
@@ -275,3 +275,70 @@ func TestSnapshotClear(t *testing.T) {
 		t.Errorf("keys should be empty")
 	}
 }
+
+func TestResponseOrdering(t *testing.T) {
+	c := cache.NewSnapshotCache(true, group{}, logger{t: t})
+	endpointWatch, _ := c.CreateWatch(&discovery.DiscoveryRequest{TypeUrl: rsrc.EndpointType, ResourceNames: names[rsrc.EndpointType]})
+	clusterWatch, _ := c.CreateWatch(&discovery.DiscoveryRequest{TypeUrl: rsrc.ClusterType, ResourceNames: names[rsrc.ClusterType]})
+	routeWatch, _ := c.CreateWatch(&discovery.DiscoveryRequest{TypeUrl: rsrc.RouteType, ResourceNames: names[rsrc.RouteType]})
+	listenerWatch, _ := c.CreateWatch(&discovery.DiscoveryRequest{TypeUrl: rsrc.ListenerType, ResourceNames: names[rsrc.ListenerType]})
+	typeUrlCh := make(chan []string)
+	go func() {
+		var responseTypeUrls []string
+		for len(responseTypeUrls) < 4 {
+			chosen := map[string]bool{}
+			select {
+			case <-clusterWatch:
+				chosen[rsrc.ClusterType] = true
+			case <-endpointWatch:
+				chosen[rsrc.EndpointType] = true
+			case <-listenerWatch:
+				chosen[rsrc.ListenerType] = true
+			case <-routeWatch:
+				chosen[rsrc.RouteType] = true
+			}
+			if !chosen[rsrc.ClusterType] {
+				select {
+				case <-clusterWatch:
+					chosen[rsrc.ClusterType] = true
+				default:
+				}
+			}
+			if !chosen[rsrc.EndpointType] {
+				select {
+				case <-endpointWatch:
+					chosen[rsrc.EndpointType] = true
+				default:
+				}
+			}
+			if !chosen[rsrc.ListenerType] {
+				select {
+				case <-listenerWatch:
+					chosen[rsrc.ListenerType] = true
+				default:
+				}
+			}
+			if !chosen[rsrc.RouteType] {
+				select {
+				case <-routeWatch:
+					chosen[rsrc.RouteType] = true
+				default:
+				}
+			}
+			for _, key := range []string{rsrc.ClusterType, rsrc.EndpointType, rsrc.ListenerType, rsrc.RouteType} {
+				if chosen[key] {
+					responseTypeUrls = append(responseTypeUrls, key)
+				}
+			}
+		}
+		typeUrlCh <- responseTypeUrls
+	}()
+	if err := c.SetSnapshot(key, snapshot); err != nil {
+		t.Fatal(err)
+	}
+	typeUrls := <-typeUrlCh
+	if !reflect.DeepEqual(typeUrls, []string{rsrc.ClusterType, rsrc.EndpointType, rsrc.ListenerType, rsrc.RouteType}) {
+		t.Fatalf("incorrect types order: got %v, expected %v", typeUrls, []string{rsrc.ClusterType, rsrc.EndpointType, rsrc.ListenerType, rsrc.RouteType})
+	}
+
+}
--- a/pkg/cache/v3/simple.go
+++ b/pkg/cache/v3/simple.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/envoyproxy/go-control-plane/pkg/cache/types"
 	"github.com/envoyproxy/go-control-plane/pkg/log"
+	"github.com/envoyproxy/go-control-plane/pkg/resource/v3"
 )
 
 // SnapshotCache is a snapshot-based cache that maintains a single versioned
@@ -103,8 +104,16 @@ func NewSnapshotCache(ads bool, hash NodeHash, logger log.Logger) SnapshotCache 
 	}
 }
 
+type responseInfo struct {
+	id      int64
+	watch   ResponseWatch
+	version string
+}
+
 // SetSnapshotCache updates a snapshot for a node.
 func (cache *snapshotCache) SetSnapshot(node string, snapshot Snapshot) error {
+	toRespond := map[string][]responseInfo{}
+
 	cache.mu.Lock()
 	defer cache.mu.Unlock()
 
@@ -117,16 +126,27 @@ func (cache *snapshotCache) SetSnapshot(node string, snapshot Snapshot) error {
 		for id, watch := range info.watches {
 			version := snapshot.GetVersion(watch.Request.TypeUrl)
 			if version != watch.Request.VersionInfo {
-				if cache.log != nil {
-					cache.log.Debugf("respond open watch %d%v with new version %q", id, watch.Request.ResourceNames, version)
-				}
-				cache.respond(watch.Request, watch.Response, snapshot.GetResources(watch.Request.TypeUrl), version)
+				toRespond[watch.Request.TypeUrl] = append(toRespond[watch.Request.TypeUrl], responseInfo{
+					id:      id,
+					watch:   watch,
+					version: version,
+				})
 
 				// discard the watch
 				delete(info.watches, id)
 			}
 		}
 		info.mu.Unlock()
+	}
+
+	for _, typeUrl := range resource.TypeUrls {
+		resources := snapshot.GetResources(typeUrl)
+		for _, info := range toRespond[typeUrl] {
+			if cache.log != nil {
+				cache.log.Debugf("respond open watch %d%v with new version %q", info.id, info.watch.Request.ResourceNames, info.version)
+			}
+			cache.respond(info.watch.Request, info.watch.Response, resources, info.version)
+		}
 	}
 
 	return nil
--- a/pkg/cache/v3/simple_test.go
+++ b/pkg/cache/v3/simple_test.go
@@ -276,3 +276,70 @@ func TestSnapshotClear(t *testing.T) {
 		t.Errorf("keys should be empty")
 	}
 }
+
+func TestResponseOrdering(t *testing.T) {
+	c := cache.NewSnapshotCache(true, group{}, logger{t: t})
+	endpointWatch, _ := c.CreateWatch(&discovery.DiscoveryRequest{TypeUrl: rsrc.EndpointType, ResourceNames: names[rsrc.EndpointType]})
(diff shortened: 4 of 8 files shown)


fxposter:
cc @kyessenov, cause you opened https://github.com/envoyproxy/go-control-plane/issues/59.

I am really not sure this is the way to go, but for us this is needed to have the ability to reliably change clusters and ensure corresponding endpoints are sent afterwards. Didn't benchmark yet.

kyessenov:
Envoy usually tolerates ordering violation with warming. If you enable cluster and listener warming, then Envoy would just wait until the resources arrive, so you don't have to do the work on the server. The problem is that EDS is often served separately, so it cannot really follow the ordering requirements imposed by a different server.

The really interesting bit is deletion of resources. If you remove a cluster in the next snapshot, then the current set of listeners (applied, since next one is warming), might start dropping traffic.

fxposter:
Basically, due to heavy use of channels and the fact that server.go uses one big select - passing data between cache and server loses ordering and we need to do these hacks on the server side. What I am trying to solve there is the situation when multiple channels become ready at the same time and (in that case) - we get responses from all channels and sort them afterwards as part of a single loop iteration.
Alternative is to completely rewrite the interface between server and cache not to use channels.

fxposter:
@kyessenov we see different behavior - if endpoints are sent before cluster and are not re-sent again (and from versioning perspective they should not be) - such clusters will return 503s after 15 seconds of warmup.
Imagine you have a cluster X with version C1, and endpoints X with version E1 that are already propagated to envoy and after that we update both to C2 and E2, where C2's cluster state is different somehow (for example, has different timeout settings or circuit breakers).
Good case: C2 is sent first, cluster is transitioned into warming state, E2 is sent afterwards, finishing the warming state.
Bad case: E2 is sent first, it is applied to C1 on envoy side. C2 is sent next, transitioning cluster into warming state. After 15 seconds if there is no E3 - cluster will return 503s and "no healthy upstream".

fxposter:
This one is handled in java-control-plane: https://github.com/envoyproxy/java-control-plane/blob/master/cache/src/main/java/io/envoyproxy/controlplane/cache/SimpleCache.java#L231, but due to their implementation not using channels - they don't lose ordering. In theory they can, cause they submit sending tasks to executor, but (as was mentioned in slack - it rarely happens https://envoyproxy.slack.com/archives/C7LDJTM6Z/p1600096898081600?thread_ts=1599576829.051000&cid=C7LDJTM6Z)

kyessenov:
@fxposter Thanks for explaining. Yes, I agree we should respect the order, and ship updates in reverse dependency order. Let me think what we can do about implementation.

kyessenov:
I think this can only work reliably if CDS and EDS are combined into ADS. So we can either ship the entire snapshot from cache to server, or lookup versions in the resources somehow to associate CDS and EDS requests (some EDS requests might be stale for a pending CDS update).

kyessenov:
Here is an idea:
- make createwatch take a stream_id and return a watch_id for the watch, and server call out to cache when the response is sent with `Responded(watch_id, api_type)`
- group watches by stream_id within statusInfo
- when we respond to watches, only respond to CDS/LDS immediately, leave EDS/RDS watches open
- respond to EDS after CDS is responded, RDS after LDS is responded

This is fairly complicated. Not sure if there is a simpler way to coordinate this.

kyessenov:
To be honest, I think it's wrong for Envoy to be warming a cluster that just has a field changed and not the associated EDS. We can also try to make Envoy smarter and not warm clusters with available EDS info.

fxposter:
> To be honest, I think it's wrong for Envoy to be warming a cluster that just has a field changed and not the associated EDS. We can also try to make Envoy smarter and not warm clusters with available EDS info.

https://github.com/envoyproxy/envoy/issues/13009 :) probably in the future we will.

fxposter:
> I think this can only work reliably if CDS and EDS are combined into ADS. So we can either ship the entire snapshot from cache to server, or lookup versions in the resources somehow to associate CDS and EDS requests (some EDS requests might be stale for a pending CDS update).

well, you cannot provide any guarantees within a library for non-ads mode. In ADS mode, though, such changes are required when you plan to change clusters or listeners :)



> Here is an idea:
> 
>     * make createwatch take a stream_id and return a watch_id for the watch, and server call out to cache when the response is sent with `Responded(watch_id, api_type)`
> 
>     * group watches by stream_id within statusInfo
> 
>     * when we respond to watches, only respond to CDS/LDS immediately, leave EDS/RDS watches open
> 
>     * respond to EDS after CDS is responded, RDS after LDS is responded
> 
> 
> This is fairly complicated. Not sure if there is a simpler way to coordinate this.

I don't get your picture :) My understanding is that ideally we should send "warmup" EDS/RDS after the "ACK" from envoy, but that's quite complicated to do from current state of library, especially given the fact that it does not support ack/nack at all.
That's why I went this route - small rewrite and no API breaking changes.

In my implementation of control plane I do use "application version" as part of envoy resource versions, ie: when we release new version of the app - all state is resent to envoys (this was the easiest and most generic way to actually propagate code changes that are related to proto generators). I tried simulating doing such "releases" with changes to clusters (same as I've described above C1/E1 -> C2/E2).
I have a control plane and single envoy connected to it. I shutdown control plane and start new one that generates different versions and different clusters state. And I can easily get "no healthy upstreams" bug with this simulation locally. With this patch it's gone (at least in my tests, I will do more changes). And given similar functionality exist in java-control-plane - I do think it should be supported here.

kyessenov:
Yeah, I am still thinking on how to approach it. Your proposal makes the server hold EDS response which may be out of order with CDS (requested for older cluster response for some reason). I feel like some consistency is lost by trying to re-build the snapshot in the server.

fxposter:
> Your proposal makes the server hold EDS response which may be out of order with CDS (requested for older cluster response for some reason)

Totally possible with custom cache implementations. With combination of linear cache and simple cache - probably, yes. But: my change only applies to situation where multiple channels are filled with data at the same time. And previously same "out of order" events could happen too, cause there were no guarantees on which channels will be selected

fxposter:
Personally I feel like the better approach would be to have single channel passed into cache for all resources (at least I personally would start with this refactoring), but - that requires changes in public API and I don't know what is the status with API compatibility of this library

kyessenov:
If we want to push the entire snapshot from the cache, how would we dispatch watches? We still have to wait for Envoy to ask for resources before pushing them.

fxposter:
Currently go-control-plane doesn't handle nacks, hence more or less assuming that everything will succeed. Which means - "ordering" is done when Send(DiscoveryResponse) is called.

kyessenov:
@fxposter I am just concerned about some scenario like this:
1. xds pushes c1 
2. xds pushes c1, c2
3. envoy requests c1

If (3) comes late, after (2) is received, we might be holding (3) unnecessarily. Just want to understand that something like that cannot happen.

Lack of NACKs is an old problem. There really isn't much to do with NACKs on the server, and unfortunately Envoy can NACK intermittently, so you just have to keep trying.

fxposter:
Not sure I follow. What does "xds pushes c1, c2" mean? if you are talking about pushing same type several times - then IMO my changes don't affect this behaviour, cause my concern is basically - XDS fill multiple types at the same time at cache layer and server layer cannot keep up with those changes.

Also, I guess that after pushing c2 if envoy will re-request c1 - control plane will consider the nonce as "stale" and will just ignore this request (regardless of this PR).

fxposter:
ideally we should not have all those channels, especially buffered channels, here. Cause we lose the ability to control responses order in cache layer, hence we need to recreate ordering on the server side if we can. In most cases we really cannot (if there were multiple updates of clusters or endpoints and they don't reach their channels at the same time when select happens), but we minimize the probability of such "out of order" updates if we can.

kyessenov:
@fxposter IMHO Ordering is a bad smell that is really caused by xDS failure to mux various xDSes together. One of the requirements for this library was to make it work for all flavors of xDS. So channels were natural because they are easy to mux in golang. Do you think we could fix in the cache by triggering a watch with a "fake" version after CDS/LDS are pushed? I can try that approach? My example was that even if we respond to EDS after CDS, we may still respond to a wrong EDS request, that does not warm the CDS. E.g.

1. Envoy has C1, C2, issues EDS(1) for C1, C2.
2. GCP pushes updated C1, C2 puts them into warming
3. GCP pushes EDS(1) response.
4. Envoy accepts EDS(1) but does not warm C1, C2 because it has not issued EDS(2) yet.

fxposter:
I am really not sure how it can work cleanly at all in presence of cluster updates and warming :( Truth is - currently it is easy to reproduce a behavior where go-control-plane fails to do what it ideally should do. Istio went writing their ADS server from scratch and they don't support non-ADS mode to maintain consistency. java-control-plane decided to basically use callbacks instead of channels, and by default executes "send" directly on the caller's thread, ie: send is called from "cache" gorouting, if we talk about go-control plane's architecture, which allows controlling order on cache side AND it does have code to restructure sends to avoid traffic drops.

fxposter:
Well, I guess I'm lost in terminology :)

Let's try to start with this:
C(1), C(2), ..., C(n) - different versions of the same cluster, that include changes to the value of the cluster (if you send multiple versions of the same cluster that do not change the cluster *value* - envoy does not start warming)
E(1), E(2), ..., E(n) - versions of endpoints that "should" match versions of changing cluster, ie: C(1) should ideally have E(1) endpoints, etc...

Problem lies in the fact that neither GCP, nor envoy really care about C(n) <-> E(n) match. And GCP cannot ensure that match on envoy side. Envoy only cares that if, for some reason, cluster is changed by value - it should receive an endpoint update afterwards.

1. If E(any n) is compatible to C(any m) - we just want to maintain order, everything else does not matter. This PR always chose right order, current implementation gives no guarantees about that.
2. If E(n) is not compatible with C(m, where m != n). And we need to transition from E(n) + C(n) to E(n+1) + C(n+1), we can have 3 cases, depending on the state of "server" channels:
2.1. endpoints channel has E(n) in it, clusters channel has C(n+1). in new implementation the only possible outcome is C(n+1), then E(n), which leads to incorrectly working cluster on envoy side. *but* - there is no guarantee that cannot happen in old implementation as we cannot guarantee the order.
2.2. endpoints channel has E(n+1) in it, clusters channel has C(n) - this is the case when both C(n), then E(n+1) and E(n+1), C(n) will lead to problems in both implementations.
2.2. endpoints channel has E(n+1) in it, clusters channel has C(n+1) - this is the case where this PR choses the obviously right one, while current implementation give no guarantees.

fxposter:
To follow java-control-plane implementation - we basically need to change the interface between cache and server to callbacks :( which is breaking change for writers of their own caches.

fxposter:
https://github.com/fxposter/go-control-plane/commit/e4067cda31236b19ba15f31a217385474e949919 - this one will allow control over ordering from cache perspective. And rewriting caches to such interface is usually straightforward - just follow the types.

kyessenov:
Wouldn't you need some locking discipline without channels? We can't respond from cache go routine if server is busy processing some gRPC stream, right? Channels were used to avoid any mutexes at steady state.

The more I think about this, the more I dislike the brokenness of Envoy client. If you separate EDS and CDS into two servers, then EDS server can basically never be sure that the version is accepted and has to keep re-pushing EDS over and over... Maybe that's what we need to do, too? That's what Istio was doing, which I hated at the time warming was introduced.

kyessenov:
What do you think about a workaround of timer-pased updates to EDS in the snapshot? Basically, keep incrementing the version of EDS every second and then call SetSnapshot() again. I think that won't trigger watches for non-EDS resources, and would resolve dead-locked clusters.

fxposter:
> Wouldn't you need some locking discipline without channels? We can't respond from cache go routine if server is busy processing some gRPC stream, right? Channels were used to avoid any mutexes at steady state.

The overall behavior is changed, indeed, SetSnapshot will block now, that can be handled at the library user level, though.

> What do you think about a workaround of timer-based updates to EDS in the snapshot? Basically, keep incrementing the version of EDS every second and then call SetSnapshot() again. I think that won't trigger watches for non-EDS resources, and would resolve dead-locked clusters.

I prefer not going into details when possible. My concern is that we should not lose ordering dictated by cache layer.

fxposter:
https://github.com/fxposter/go-control-plane/commit/1bb625e563f01ab392dcb0c182deccfbc1bb981e WDYT about this. This also breaks API, but this is what I wanted from the beginning :) My main concern was the fact that in order to keep ordering you need to either use blocking channels, have unbounded channel (cause you have no idea how many responses can you get on the library side) or configure channel buffer somehow (questionable, cause as a library user you may guess the number of responses, but not be 100% sure).
The workaround in that commit is to have an additional channel that will be blocking and separate goroutine will read data from that channel and put into a slice, and then do a single non-blocking notification to main loop, which will basically take the slice and send data to all items in that slice in order.

fxposter:
as a consequence - there is no real need to have separate flows for endpoints/clusters/etc, compared to additional "opaque" resources, cause they are doing the same thing now and should be unified. and, also, additional goroutines are gone and problems related to their termination are gone too.

fxposter:
In such setup your concerns are gone - server will not be able to reshuffle what cache is sent to it in a wrong way and ordering can be fully controlled on a cache layer. But the main downside is that it is not backwards compatible for 3rd party cache implementations.

fxposter:
After thinking more - https://github.com/fxposter/go-control-plane/commit/1bb625e563f01ab392dcb0c182deccfbc1bb981e - there is one more thing that was possible to do, but not possible with this change - "delayed" error. IE: previously it was possible to close response channel at some later time and, therefore, error out the stream. In my case it is not possible anymore, but no built-in caches need that.

kyessenov:
Awesome idea @fxposter ! I think we can simplify server further (see https://github.com/kyessenov/go-control-plane/pull/new/single_channel) and avoid mutexes altogether. This approach sounds like both simpler and more flexible, so we should just do it. I'm not too concerned about breaking the interface since it's not particularly different from a cache perspective, and we haven't stabilized interfaces.

I can continue the refactor to share the response channel if you'd like.

fxposter:
> Awesome idea @fxposter ! I think we can simplify server further (see https://github.com/kyessenov/go-control-plane/pull/new/single_channel) and avoid mutexes altogether. This approach sounds like both simpler and more flexible, so we should just do it. I'm not too concerned about breaking the interface since it's not particularly different from a cache perspective, and we haven't stabilized interfaces.
> 
> I can continue the refactor to share the response channel if you'd like.

I still believe my solution is better, due to the fact, that you don't need to guess size of responses buffer. For example, for people still using non-ADS endpoints - the actual envoy behaviour is to send a request per resource, not per resource type, hence hundreds of such requests are quite likely to happen - making cache.setSnapshot blocking.

fxposter:
Just though that another alternative might be to use chan []cache.Response (probably, will think about that more).

kyessenov:
I don't think we can fully guarantee the ordering since this is just racey in principle (three actors, envoy, server, and cache acting independently). So whether or not the cache sends a slice, there is a chance envoy could have sent something between "cache send" and "server received".

Re: drain vs buffer: it depends on what's worse: blocked cache or uncontrolled memory growth in the server. As long as there are compute resources, it seems that it's better to be busy pushing results rather than busy processing requests, although, we may need to find some balance there.

fxposter:
> or uncontrolled memory growth in the server
not "uncontrolled", but "dependent on number of requests to control-plane", which depends on cluster-size in specific cases. Hence, we can guarantee being non-blocking, with "bounded memory usage by cluster size" implementation. Sending slices of responses guarantees ordering from perspective of "cache" - ie: "SetSnapshot" can, if needed, write to channel once for whole node.

fxposter:
After trying to implement `chan []cache.Response` - the API without large rewrite looks weird and inconsistent, so in the current situation - I'd stick with additional slice approach, because
1. it does actually guarantee non-blocking nature of caches
2. it does not necessarily mean "unbounded memory usage" compared to current implementation, I think. we would basically have exactly the same number of slice elements as we currently have channels, that are created in each CreateWatch, no?
3. in the end in both my and your approach we are bounded by the fact that we have backpressure on the actual XDS layer - the amount of data in slice or channel will depend on how many requests envoy is doing to control plane, and it does all those requests after it receives updates.

fxposter:
@kyessenov re-checked envoy's behavior when not using ADS. So this buffer is only applicable when using ADS, cause envoy actually opens separate *stream* per resource. And the size of this buffer should be limited to the number of possible resource types that client of control plane supports.

https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#resource-types mentions 8 types for v2 and 8 types for v3.
From cache's perspective - simple caches supports 6 resource types.

We probably need more than that number of items in buffer when envoy re-requests some resources and GCP processes that request when the response for the same type is already in responses channel - hence we can have multiple responses for the same type.

So ideally - I would still want to have my solution in place (cause we guarantee to never block, and it is VERY unlikely that we will insert many responses there), but having no intermediate slice and be able to configure the size of buffer probably is a reasonable solution (without the need for inner loop, cause we assume we actually never block on writes to the channel anyway). The downside of the latter solution is that it is not actually possible to determine, whether the cache was blocked on that channel (and hence it needs to be resized for that particular situation) or not.

kyessenov:
ADS muxes streams, so there should be at most the number of different xDSes outstanding watches AFAICT. Subsequent requests should be cancelling prior watches if the cache implements them correctly. Maybe there's an outstanding response sent just before the watch got cancelled, but bounding the buffer by <100 should provide ample space for the responses (and these are pointers, so not stored by value).

Using an intermediate slice implements an unlimited queue. Apparently, golang considered that for channels (https://github.com/golang/go/issues/20352). I think it's fine, too. If you insist, we can have an option to use a limited or unlimited work queue. It adds a cost of an extra goroutine and a mutex, but I can't really tell if it matters without benchmarks.

fxposter:
> Subsequent requests should be cancelling prior watches if the cache implements them correctly
Not necessarily, I think. Imagine the situation where the CDS response was sent and EDS response was generated and is already in response channel, but not yet `select`-ed. New EDS request comes in at the same time - cancelling previous one - but it already was written to channel, so it may be in responses channel alongside with another request for same type.

fxposter:
If you don't think we need that - I am fine with opening this: https://github.com/envoyproxy/go-control-plane/compare/master...fxposter:new-api-between-cache-and-server?expand=1 (no intermediate slice, no inner responses loop, configurable buffer size) :)

kyessenov:
Please open a PR and we can discuss it. We can add an unbuffered mode as a follow-up. The exceptional case you mentioned is possible, and I agree, we should add some extra space to the buffer beyond # xDS types. I suppose draining the entire channel would it make less likely (although select fairness is pretty crappy in my experience).

Issue #293, opened 2020-04-26
Wellknown filter names are deprecated


jpeach:
```
[2020-04-26 22:20:49.645][1][warning][misc] [bazel-out/k8-opt/bin/source/extensions/common/_virtual_includes/utility_lib/extensions/common/utility.h:65] Using deprecated extension name 'envoy.router' for 'envoy.filters.http.router'. This name will be removed from Envoy soon. Please see https://www.envoyproxy.io/docs/envoy/latest/intro/deprecated for details.
[2020-04-26 22:20:49.645][1][warning][misc] [bazel-out/k8-opt/bin/source/extensions/common/_virtual_includes/utility_lib/extensions/common/utility.h:65] Using deprecated extension name 'envoy.gzip' for 'envoy.filters.http.gzip'. This name will be removed from Envoy soon. Please see https://www.envoyproxy.io/docs/envoy/latest/intro/deprecated for details.
```

Any program using the `wellknown` package constants for filter names will run into the problem that these filter names are deprecated in Envoy. However, just changing to the new names seems hard because the API might be used by programs that server older versions of Envoy that don't support the new names.

acnodal-tc:
Just curious - does the merge of #294 resolve this issue?

kyessenov:
Protobuf type takes precedence, so the name is not important as of last couple of releases of Envoy.

acnodal-tc:
Cool, thanks for the info.  Can we close this issue?

kyessenov:
Yeah, let's close it. The action item is just to document.

jpeach:
> On 30 Jul 2020, at 8:59 am, Kuat <notifications@github.com> wrote:
> 
> 
> Yeah, let's close it. The action item is just to document.

Would you take a PR documenting the names as deprecated and recommending use of the type URLs?

kyessenov:
> Would you take a PR documenting the names as deprecated and recommending use of the type URLs?
Absolutely.

fxposter:
> Protobuf type takes precedence, so the name is not important as of last couple of releases of Envoy.

at least not in 1.12 :)

PR #354, opened 2020-09-18
pkg/wellknown: Update gzip wellknown name

stevesloka:
This bumps the the Gzip HTTP filter wellknown name. 

Signed-off-by: Steve Sloka <slokas@vmware.com>

### DIFF
--- a/pkg/wellknown/wellknown.go
+++ b/pkg/wellknown/wellknown.go
@@ -32,7 +32,7 @@ const (
 	// GRPCWeb HTTP filter
 	GRPCWeb = "envoy.filters.http.grpc_web"
 	// Gzip HTTP filter
-	Gzip = "envoy.filters.http.gzip"
+	Gzip = "envoy.filters.http.compressor"
 	// IPTagging HTTP filter
 	IPTagging = "envoy.filters.http.ip_tagging"
 	// HTTPRateLimit filter


stevesloka:
As an aside, I was trying to create the filter via a typedConfig to avoid updating these all the time, but I get this error: `Proto constraint validation failed (HttpConnectionManagerValidationError.HttpFilters[i]: ["embedded message failed validation"] | caused by HttpFilterValidationError.Name: ["value length must be at least " '\x01' " bytes"])` 

I think this is resolved in the v3 api though? 

```
&http.HttpFilter{
  ConfigType: &http.HttpFilter_TypedConfig{
	TypedConfig: &any.Any{
		TypeUrl: "type.googleapis.com/envoy.config.filter.http.gzip.v2.Gzip",
	},
   },
},
```

jpeach:
IIUC you have to put a name but it can be anything.

fxposter:
> Doesn't changing the well known names have compatibility issues? If an app upgrades this package but not envoy, it would break right?

well, last modification of this file already broke envoy 1.12 support

jpeach:
> > Doesn't changing the well known names have compatibility issues? If an app upgrades this package but not envoy, it would break right?
> 
> well, last modification of this file already broke envoy 1.12 support


https://github.com/envoyproxy/go-control-plane/issues/293

stevesloka:
> IIUC you have to put a name but it can be anything.

If I put anything in for the name I get this error:

```
[source/common/config/grpc_subscription_impl.cc:100] gRPC config for type.googleapis.com/envoy.api.v2.Listener rejected: Error adding/updating listener(s) ingress_http: Didn't find a registered implementation for name: 'gzip'
```

This is the config I'm using:
```
&http.HttpFilter{
  Name: "gzip",
  ConfigType: &http.HttpFilter_TypedConfig{
	TypedConfig: &any.Any{
		TypeUrl: "type.googleapis.com/envoy.extensions.filters.http.compressor",
	},
    },
},
```

From another Slack thread (can't find the link now), this should work in v3 api's but not v2. 

Is there something I'm missing? Or should this only work with v3?

jpeach:
@htuch is the well-known filter name still required in the v2 API?

htuch:
@jpeach as long as you are using some extension config wrapper that has explicit type URLs, you shouldn't need to use an explicit well known name.

stevesloka:
@htuch can you look at my last comment in this thread? I tried to do that, but still get errors unless the name is something that matches.

htuch:
It looks like we are registering both v2 and v3 type URLs in the registry, so not sure why it wouldn't work. @kyessenov any thoughts?

kyessenov:
I think it has something to do with HCM being v2 here. Let me try it out locally.

kyessenov:
BTW the type should be envoy.extensions.filters.http.compressor.v3.Compressor.

kyessenov:
Re: well known names: yes, it does break updates of this library in non-trivial ways. Istio has this arbitrary patching mechanism (EnvoyFilter) that can refer to parts of xDS by name, so when we bump this library, the output changes slightly and breaks some EnvoyFilters. Unless you really have to use the name, I'd just use the type since it takes precedence anyways.

kyessenov:
This works:
```yaml
          - name: gzip
            typed_config:
              "@type": type.googleapis.com/envoy.extensions.filters.http.compressor.v3.Compressor
              compressor_library:
                name: "text_optimized"
                typed_config:
                  "@type": "type.googleapis.com/envoy.extensions.compression.gzip.compressor.v3.Gzip"
                  memory_level: 3
                  window_bits: 10
                  compression_level: "best_compression"
                  compression_strategy: "default_strategy"

```

stevesloka:
Thanks for the snippet @kyessenov, but I'm trying to get this to work with v2 at the moment. I'll just leave this as-is since we're in the middle of upgrading to v3. Once there I think this will work then. Thanks!

stevesloka:
Just an update, we finally moved to xDS v3 and with the following config it works. Thanks for the tips @kyessenov!

```
&http.HttpFilter{
			Name: wellknown.Gzip,
			ConfigType: &http.HttpFilter_TypedConfig{
				TypedConfig: protobuf.MustMarshalAny(&compressor.Compressor{
					CompressorLibrary: &envoy_core_v3.TypedExtensionConfig{
						Name: "gzip",
						TypedConfig: &any.Any{
							TypeUrl: "type.googleapis.com/envoy.extensions.compression.gzip.compressor.v3.Gzip",
						},
					},
				}),
			},
		},
```

PR #355, opened 2020-09-19
Incremental xDS

alecholmez:
## Status
- [PASSING] Unit tests are passing!
- [PASSING] Integration tests are passing!

Ready for review! @kyessenov @jessicayuen 

### Versions
- Envoy 1.15.0+
- Go 1.11+

### Quickstart
From the root of the project:
```
./build/run_docker.sh make build test
```

To run through the integration test:
```
./build/run_docker.sh make integration
```


pkg/cache/v2/snapshot.go
```diff
@@ -122,3 +184,42 @@ func (s *Snapshot) GetVersion(typeURL string) string {
...
+	if s == nil {
+		return nil
+	}
+
+	versionMap := initializeVMap()
+	for i := 0; i < int(types.UnknownType); i++ {
+		typeURL := GetResponseTypeURL(types.ResponseType(i))
+		resources := s.GetResources(typeURL)
+		for _, resource := range resources {
+			// hash our verison in here and build the version map
```
oblazek, line 199: What if `resource` is a nil? Does it make sense to skip those cases like this?:

```
if reflect.ValueOf(resource).IsNil() {
    continue
} else {
    v, err := HashResource(resource)
    .
    .
    .
```

pkg/server/callbacks/v2/callbacks.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"context"
+
+	discovery "github.com/envoyproxy/go-control-plane/envoy/api/v2"
+)
+
+// Callbacks is a collection of callbacks inserted into the server operation.
+// The callbacks are invoked synchronously
+type Callbacks interface {
```
fxposter, line 11: Why did you decide to unify callbacks back to single interface?
IE: why users that want to do sotw.NewServer now need to provide all methods, while sotw would not use them?

alecholmez: So I unified these here again because as I was writing Incremental, I realized that I reused `OnStreamOpen` and `OnStreamClosed`, and I didn't like the duplication everywhere. If we can come up with a solution with minimal duplication I'm all for it, But they don't need to implement the method, writing something like: 
```go
func OnStreamDeltaRequest(int64, *discovery.DeltaDiscoveryRequest) error {
    return nil
}
```
would be perfectly acceptable, I've seen quiet a few projects do this already with the REST methods when using SOTW. I personally thought it made the code easier to understand and cleaned up the duplication but if we need to change this I'm open to a change.

pkg/server/delta/v2/server.go
```diff
@@ -0,0 +1,494 @@
...
+			if !more {
+				return nil
+			}
+			if req == nil {
+				return status.Errorf(codes.Unavailable, "empty request")
+			}
+
+			// Log out our error detail from envoy if we get one but don't do anything crazy here yet
+			if req.ErrorDetail != nil {
+				s.log.Errorf("received error from envoy: %s", req.ErrorDetail.String())
```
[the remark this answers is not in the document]
alecholmez: Yup thanks for the catch.

pkg/server/callbacks/v2/callbacks.go
fxposter: It's not about duplication, actually. It is about separation of concerns. The split was done to make those parts separate. Unifying part of it and making separate parts depend on single part, which is the union of all of those does not make actual sense.
I would use different methods like OnDeltaStreamOpen and OnDeltaStreamClosed, while the default implementation in server.go may reuse same callback for both delta and non-delta methods, unless delta callbacks are explicitly specified.

fxposter: Also, from go 1.14 it is allowed to have overlapping methods, but I would still go another route and allow users of this library to actually understand whether they received a delta or non-delta stream.

alecholmez: I understand why the split was done. I'm a fan of having explicit DeltaStream callbacks. I'll make that change, I think providing those explicit callback methods does make more sense to the other. I do see how reusing could be confusing

justincely:
@fxposter @oblazek could we get a new review on this?

justincely:
@jessicayuen @kyessenov  too

justincely:
bump

s-matyukevich:
@alecholmez I was trying to use this PR for our control-plane and I think I found a couple of issues
1. The code ignores `ResourceNamesSubscribe` field of the incoming xDS request. You read this field [here](https://github.com/alecholmez/go-control-plane/blob/incremental-alt/pkg/cache/v3/simple_delta.go#L30) into `aliases` variable but then you don't pass `aliases` to the  [respondDelta](https://github.com/alecholmez/go-control-plane/blob/incremental-alt/pkg/cache/v3/simple_delta.go#L105) and [createDeltaResponse](https://github.com/alecholmez/go-control-plane/blob/incremental-alt/pkg/cache/v3/simple_delta.go#L114) My understanding is that inside `createDeltaResponse` method we should filter resources and include only the ones specified in the `versionMap` + `aliases`. The result is that control plane always returns all available resources.
1. [unsubscribed](https://github.com/alecholmez/go-control-plane/blob/incremental-alt/pkg/cache/v3/simple_delta.go#L114) parameter is never used. You actually handle unsubscription [here](https://github.com/alecholmez/go-control-plane/blob/incremental-alt/pkg/server/delta/v3/server.go#L306-L310)
1. [here](https://github.com/alecholmez/go-control-plane/blob/incremental-alt/pkg/server/delta/v3/server.go#L306-L310) `versionChange` is set to true whenever there is no version change, which is confusing. Also here we should take into account only resources the client is subscribed to, otherwise, we will be sending empty responses whenever there is a change in an unsubscribed resource. 
1. [here](https://github.com/alecholmez/go-control-plane/blob/incremental-alt/pkg/cache/v3/snapshot.go#L201) you are using resource hash to calculate its version. This works fine, with the caveat that all items in the inner collections of the resource should be sorted in the exact same order every time we are calculating the hash. My original code generates the endpoint list for the `endpoint.ClusterLoadAssignment` from a map, and this results in a random ordering whenever I generate a snapshot. We may consider implementing a better hash function or at least documenting this behavior.

alecholmez:
> @alecholmez I was trying to use this PR for our control-plane and I think I found a couple of issues
> 
> 1. The code ignores `ResourceNamesSubscribe` field of the incoming xDS request. You read this field [here](https://github.com/alecholmez/go-control-plane/blob/incremental-alt/pkg/cache/v3/simple_delta.go#L30) into `aliases` variable but then you don't pass `aliases` to the  [respondDelta](https://github.com/alecholmez/go-control-plane/blob/incremental-alt/pkg/cache/v3/simple_delta.go#L105) and [createDeltaResponse](https://github.com/alecholmez/go-control-plane/blob/incremental-alt/pkg/cache/v3/simple_delta.go#L114) My understanding is that inside `createDeltaResponse` method we should filter resources and include only the ones specified in the `versionMap` + `aliases`. The result is that control plane always returns all available resources.
> 2. [unsubscribed](https://github.com/alecholmez/go-control-plane/blob/incremental-alt/pkg/cache/v3/simple_delta.go#L114) parameter is never used. You actually handle unsubscription [here](https://github.com/alecholmez/go-control-plane/blob/incremental-alt/pkg/server/delta/v3/server.go#L306-L310)
> 3. [here](https://github.com/alecholmez/go-control-plane/blob/incremental-alt/pkg/server/delta/v3/server.go#L306-L310) `versionChange` is set to true whenever there is no version change, which is confusing. Also here we should take into account only resources the client is subscribed to, otherwise, we will be sending empty responses whenever there is a change in an unsubscribed resource.
> 4. [here](https://github.com/alecholmez/go-control-plane/blob/incremental-alt/pkg/cache/v3/snapshot.go#L201) you are using resource hash to calculate its version. This works fine, with the caveat that all items in the inner collections of the resource should be sorted in the exact same order every time we are calculating the hash. My original code generates the endpoint list for the `endpoint.ClusterLoadAssignment` from a map, and this results in a random ordering whenever I generate a snapshot. We may consider implementing a better hash function or at least documenting this behavior.

@ s-matyukevich i'll look into these issues asap. Thanks for voicing your concerns! Will respond shortly

s-matyukevich:
Thanks, @alecholmez! There is one more thing that we need in order to start using this functionality: mux and linear cache support (We want to use linear cache for EDS and simple cache for CDS/RDS/LDS). If you want I can submit another PR against your branch to implement this or address some of the issues I mentioned above.

alecholmez:
@s-matyukevich some responses to your comments:

> 1. The code ignores `ResourceNamesSubscribe` field of the incoming xDS request. You read this field [here](https://github.com/alecholmez/go-control-plane/blob/incremental-alt/pkg/cache/v3/simple_delta.go#L30) into `aliases` variable but then you don't pass `aliases` to the  [respondDelta](https://github.com/alecholmez/go-control-plane/blob/incremental-alt/pkg/cache/v3/simple_delta.go#L105) and [createDeltaResponse](https://github.com/alecholmez/go-control-plane/blob/incremental-alt/pkg/cache/v3/simple_delta.go#L114) My understanding is that inside `createDeltaResponse` method we should filter resources and include only the ones specified in the `versionMap` + `aliases`. The result is that control plane always returns all available resources.

You are correct in that assumption. That was a slight overlook on my part (I had intentions of doing that but I guess I had forgotten to do so). Thanks for the catch! Right now it's just function as a SOTW stream and returning all the objects like you said. 

> 2. [unsubscribed](https://github.com/alecholmez/go-control-plane/blob/incremental-alt/pkg/cache/v3/simple_delta.go#L114) parameter is never used. You actually handle unsubscription [here](https://github.com/alecholmez/go-control-plane/blob/incremental-alt/pkg/server/delta/v3/server.go#L306-L310)

Yup I can pull this out. That is remnants of old code so thank you for pointing that out.

> 3. [here](https://github.com/alecholmez/go-control-plane/blob/incremental-alt/pkg/server/delta/v3/server.go#L306-L310) `versionChange` is set to true whenever there is no version change, which is confusing. Also here we should take into account only resources the client is subscribed to, otherwise, we will be sending empty responses whenever there is a change in an unsubscribed resource.

So I agree but I tried to mimic the logic here from SOTW as closely as possible. You'll notice here:
```go
	if !exists || request.VersionInfo == version {
		watchID := cache.nextWatchID()
		if cache.log != nil {
			cache.log.Debugf("open watch %d for %s%v from nodeID %q, version %q", watchID,
				request.TypeUrl, request.ResourceNames, nodeID, request.VersionInfo)
		}
		info.mu.Lock()
		info.watches[watchID] = ResponseWatch{Request: request, Response: value}
		info.mu.Unlock()
		return value, cache.cancelWatch(nodeID, watchID)
	}
```
That in their `CreateWatch` function, they trigger a new watch when the version evaluates to `true`. Though I do completely agree that it's confusing.

To the second point, 100% agree let's make that change so that we only respond to the subscribed resources. The tricky thing here to be careful about is wildcard mode with LDS and CDS. Meaning if `RequestResourceNamesSubscribed` is emtpy, Envoy expects all resources to be sent over the wire.

> 4. [here](https://github.com/alecholmez/go-control-plane/blob/incremental-alt/pkg/cache/v3/snapshot.go#L201) you are using resource hash to calculate its version. This works fine, with the caveat that all items in the inner collections of the resource should be sorted in the exact same order every time we are calculating the hash. My original code generates the endpoint list for the `endpoint.ClusterLoadAssignment` from a map, and this results in a random ordering whenever I generate a snapshot. We may consider implementing a better hash function or at least documenting this behavior.

Ah interesting I could potentially move this into an `interface` or `func()` param and allow users to implement their own hashing function. That way we can offload the sorting to the user since I would assume most people will treat their resources differently.

s-matyukevich:
@alecholmez I created a PR to your branch with a few fixes https://github.com/alecholmez/go-control-plane/pull/2

PR #357, opened 2020-09-22
api: new interface between cache and server

fxposter:
As a followup of https://github.com/envoyproxy/go-control-plane/pull/353 - this changes the API between cache and server packages to allow to maintain the order of responses of different types from the cache side. The ordering itself will be implemented in separate PR when/if we stabilize this.

In general - instead of separate channels for each requests this PR introduces a single channel for a stream that cache is writing responses into, hence - ordering those within a single channel.

### DIFF
--- a/pkg/cache/v2/cache.go
+++ b/pkg/cache/v2/cache.go
@@ -38,13 +38,13 @@ type ConfigWatcher interface {
 	// An individual consumer normally issues a single open watch by each type URL.
 	//
 	// Once the requested resources are available, they should be pushed into responses channel.
-	// If there are any failures - the error should be returned and the consumer should close the
+	// If there are any failures - nil should be written into the responses channel and the consumer should close the
 	// corresponding stream.
 	// Cache implementation should not close responses channel.
 	//
 	// Cancel is an optional function to release resources in the producer. If
 	// provided, the consumer may call this function multiple times.
-	CreateWatch(request *Request, responses chan<- Response) (cancel func(), err error)
+	CreateWatch(request *Request, responses chan<- Response) (cancel func())
 }
 
 // ConfigFetcher fetches configuration resources from cache
--- a/pkg/cache/v2/cache.go
+++ b/pkg/cache/v2/cache.go
@@ -44,7 +44,7 @@ type ConfigWatcher interface {
 	//
 	// Cancel is an optional function to release resources in the producer. If
 	// provided, the consumer may call this function multiple times.
-	CreateWatch(*Request, chan Response) (cancel func(), err error)
+	CreateWatch(*Request, chan<- Response) (cancel func(), err error)
 }
 
 // ConfigFetcher fetches configuration resources from cache
--- a/pkg/cache/v2/cache.go
+++ b/pkg/cache/v2/cache.go
@@ -37,14 +37,14 @@ type ConfigWatcher interface {
 	// CreateWatch returns a new open watch from a non-empty request.
 	// An individual consumer normally issues a single open watch by each type URL.
 	//
-	// Value channel produces requested resources, once they are available.  If
-	// the channel is closed prior to cancellation of the watch, an unrecoverable
-	// error has occurred in the producer, and the consumer should close the
+	// Once the requested resources are available, they should be pushed into responses channel.
+	// If there are any failures - the error should be returned and the consumer should close the
 	// corresponding stream.
+	// Cache implementation should not close responses channel.
 	//
 	// Cancel is an optional function to release resources in the producer. If
 	// provided, the consumer may call this function multiple times.
-	CreateWatch(*Request, chan<- Response) (cancel func(), err error)
+	CreateWatch(request *Request, responses chan<- Response) (cancel func(), err error)
 }
 
 // ConfigFetcher fetches configuration resources from cache
--- a/pkg/cache/v2/cache.go
+++ b/pkg/cache/v2/cache.go
@@ -44,7 +44,7 @@ type ConfigWatcher interface {
 	//
 	// Cancel is an optional function to release resources in the producer. If
 	// provided, the consumer may call this function multiple times.
-	CreateWatch(*Request) (value chan Response, cancel func())
+	CreateWatch(*Request, chan Response) (cancel func(), err error)
 }
 
 // ConfigFetcher fetches configuration resources from cache
--- a/pkg/server/sotw/v2/server.go
+++ b/pkg/server/sotw/v2/server.go
@@ -48,15 +48,34 @@ type Callbacks interface {
 	OnStreamResponse(int64, *discovery.DiscoveryRequest, *discovery.DiscoveryResponse)
 }
 
+type Opt func(*server)
+
+func WithResponseBuffer(bufferSize int) Opt {
+	return func(s *server) {
+		s.bufferSize = bufferSize
+	}
+}
+
+func WithCallbacks(callbacks Callbacks) Opt {
+	return func(s *server) {
+		s.callbacks = callbacks
+	}
+}
+
 // NewServer creates handlers from a config watcher and callbacks.
-func NewServer(ctx context.Context, config cache.ConfigWatcher, callbacks Callbacks) Server {
-	return &server{cache: config, callbacks: callbacks, ctx: ctx}
+func NewServer(ctx context.Context, config cache.ConfigWatcher, opts ...Opt) Server {
+	s := &server{cache: config, ctx: ctx, bufferSize: 16}
+	for _, opt := range opts {
+		opt(s)
+	}
+	return s
 }
 
 type server struct {
-	cache     cache.ConfigWatcher
-	callbacks Callbacks
-	ctx       context.Context
+	cache      cache.ConfigWatcher
+	callbacks  Callbacks
+	ctx        context.Context
+	bufferSize int
 
 	// streamCount for counting bi-di streams
 	streamCount int64
@@ -78,8 +97,8 @@ type watches struct {
 }
 
 // Initialize all watches
-func (values *watches) Init() {
-	values.responses = make(chan cache.Response, 16)
+func (values *watches) Init(bufferSize int) {
+	values.responses = make(chan cache.Response, bufferSize)
 	values.cancellations = make(map[string]func())
 	values.nonces = make(map[string]string)
 }
@@ -108,7 +127,7 @@ func (s *server) process(stream Stream, reqCh <-chan *discovery.DiscoveryRequest
 
 	// a collection of stack allocated watches per request type
 	var values watches
-	values.Init()
+	values.Init(s.bufferSize)
 	defer func() {
 		values.Cancel()
 		if s.callbacks != nil {
--- a/pkg/server/sotw/v2/server.go
+++ b/pkg/server/sotw/v2/server.go
@@ -108,7 +108,6 @@ var errorResponse = &cache.RawResponse{}
 
 // Cancel all watches
 func (values *watches) Cancel() {
-	close(values.responses)
 	for _, cancel := range values.cancellations {
 		if cancel != nil {
 			cancel()
--- a/pkg/server/sotw/v2/server.go
+++ b/pkg/server/sotw/v2/server.go
@@ -169,6 +169,9 @@ func (s *server) process(stream Stream, reqCh <-chan *discovery.DiscoveryRequest
 			return nil
 		// config watcher can send the requested resources types in any order
 		case resp := <-values.responses:
+			if resp == nil {
+				return status.Errorf(codes.Unavailable, "watch failed")
+			}
 			typeUrl := resp.GetRequest().TypeUrl
 			nonce, err := send(resp, typeUrl)
 			if err != nil {
@@ -216,11 +219,7 @@ func (s *server) process(stream Stream, reqCh <-chan *discovery.DiscoveryRequest
 				if cancel, seen := values.cancellations[typeUrl]; seen && cancel != nil {
 					cancel()
 				}
-				var err error
-				values.cancellations[typeUrl], err = s.cache.CreateWatch(req, values.responses)
-				if err != nil {
-					return status.Errorf(codes.Unavailable, err.Error())
-				}
+				values.cancellations[typeUrl] = s.cache.CreateWatch(req, values.responses)
 			}
 		}
 	}
--- a/pkg/server/sotw/v2/server.go
+++ b/pkg/server/sotw/v2/server.go
@@ -72,41 +72,16 @@ type Stream interface {
 
 // watches for all xDS resource types
 type watches struct {
-	endpoints chan cache.Response
-	clusters  chan cache.Response
-	routes    chan cache.Response
-	listeners chan cache.Response
-	secrets   chan cache.Response
-	runtimes  chan cache.Response
-
-	endpointCancel func()
-	clusterCancel  func()
-	routeCancel    func()
-	listenerCancel func()
-	secretCancel   func()
-	runtimeCancel  func()
-
-	endpointNonce string
-	clusterNonce  string
-	routeNonce    string
-	listenerNonce string
-	secretNonce   string
-	runtimeNonce  string
-
-	// Opaque resources share a muxed channel. Nonces and watch cancellations are indexed by type URL.
 	responses     chan cache.Response
 	cancellations map[string]func()
 	nonces        map[string]string
-	terminations  map[string]chan struct{}
 }
 
 // Initialize all watches
 func (values *watches) Init() {
-	// muxed channel needs a buffer to release go-routines populating it
-	values.responses = make(chan cache.Response, 5)
+	values.responses = make(chan cache.Response, 16)
 	values.cancellations = make(map[string]func())
 	values.nonces = make(map[string]string)
-	values.terminations = make(map[string]chan struct{})
 }
 
 // Token response value used to signal a watch failure in muxed watches.
@@ -114,31 +89,11 @@ var errorResponse = &cache.RawResponse{}
 
 // Cancel all watches
 func (values *watches) Cancel() {
-	if values.endpointCancel != nil {
-		values.endpointCancel()
-	}
-	if values.clusterCancel != nil {
-		values.clusterCancel()
-	}
-	if values.routeCancel != nil {
-		values.routeCancel()
-	}
-	if values.listenerCancel != nil {
-		values.listenerCancel()
-	}
-	if values.secretCancel != nil {
-		values.secretCancel()
-	}
-	if values.runtimeCancel != nil {
-		values.runtimeCancel()
-	}
+	close(values.responses)
 	for _, cancel := range values.cancellations {
 		if cancel != nil {
 			cancel()
 		}
-	}
-	for _, terminate := range values.terminations {
-		close(terminate)
 	}
 }
 
@@ -195,78 +150,13 @@ func (s *server) process(stream Stream, reqCh <-chan *discovery.DiscoveryRequest
 		case <-s.ctx.Done():
 			return nil
 		// config watcher can send the requested resources types in any order
-		case resp, more := <-values.endpoints:
-			if !more {
-				return status.Errorf(codes.Unavailable, "endpoints watch failed")
-			}
-			nonce, err := send(resp, resource.EndpointType)
+		case resp := <-values.responses:
+			typeUrl := resp.GetRequest().TypeUrl
+			nonce, err := send(resp, typeUrl)
 			if err != nil {
 				return err
 			}
-			values.endpointNonce = nonce
-
-		case resp, more := <-values.clusters:
-			if !more {
-				return status.Errorf(codes.Unavailable, "clusters watch failed")
-			}
-			nonce, err := send(resp, resource.ClusterType)
-			if err != nil {
-				return err
-			}
-			values.clusterNonce = nonce
-
-		case resp, more := <-values.routes:
-			if !more {
-				return status.Errorf(codes.Unavailable, "routes watch failed")
-			}
-			nonce, err := send(resp, resource.RouteType)
-			if err != nil {
-				return err
-			}
-			values.routeNonce = nonce
-
-		case resp, more := <-values.listeners:
-			if !more {
-				return status.Errorf(codes.Unavailable, "listeners watch failed")
-			}
-			nonce, err := send(resp, resource.ListenerType)
-			if err != nil {
-				return err
-			}
-			values.listenerNonce = nonce
-
-		case resp, more := <-values.secrets:
-			if !more {
-				return status.Errorf(codes.Unavailable, "secrets watch failed")
-			}
-			nonce, err := send(resp, resource.SecretType)
-			if err != nil {
-				return err
-			}
-			values.secretNonce = nonce
-
-		case resp, more := <-values.runtimes:
-			if !more {
-				return status.Errorf(codes.Unavailable, "runtimes watch failed")
-			}
-			nonce, err := send(resp, resource.RuntimeType)
-			if err != nil {
-				return err
-			}
-			values.runtimeNonce = nonce
-
-		case resp, more := <-values.responses:
-			if more {
-				if resp == errorResponse {
-					return status.Errorf(codes.Unavailable, "resource watch failed")
-				}
-				typeUrl := resp.GetRequest().TypeUrl
-				nonce, err := send(resp, typeUrl)
-				if err != nil {
-					return err
-				}
-				values.nonces[typeUrl] = nonce
-			}
+			values.nonces[typeUrl] = nonce
 
 		case req, more := <-reqCh:
 			// input stream ended or errored out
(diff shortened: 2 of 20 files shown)


pkg/server/sotw/v2/server.go
```diff
@@ -195,78 +169,13 @@ func (s *server) process(stream Stream, reqCh <-chan *discovery.DiscoveryRequest
...
-			}
-			values.endpointNonce = nonce
-
-		case resp, more := <-values.clusters:
-			if !more {
-				return status.Errorf(codes.Unavailable, "clusters watch failed")
-			}
-			nonce, err := send(resp, resource.ClusterType)
+		case resp := <-values.responses:
+			typeUrl := resp.GetRequest().TypeUrl
```
kyessenov, line 173: Should we drain `values.responses` here? It might help with buffer overrun:
```
			for {
				typeUrl := resp.GetRequest().TypeUrl
				nonce, err := send(resp, typeUrl)
				if err != nil {
					return err
				}
				values.nonces[typeUrl] = nonce

				select {
				case resp = <-values.responses:
				default:
					break
				}
			}
```

kyessenov (commented):
LG with a few comments.

@jessicayuen FYI since it's an API change.

pkg/server/sotw/v2/server.go
fxposter: To be honest - I don't know. Do you really think it is worth doing and it will not add delays to processing of incoming requests from envoy?

fxposter:
Updated according to the comments, except for draining part.

pkg/server/sotw/v2/server.go
kyessenov: Don't know either. But thinking aloud, if server gets a lot of requests, it'll create a lot of watches that may be immediately pushed onto buffer. So even if go select is fair, it'll never clean up the buffer (each request generates produces one buffer item, each response removes one buffer item).

kyessenov:
LGTM as-is. 
@jyotimahapatra @jessicayuen do you have any comments?

pkg/server/sotw/v2/server.go
fxposter: "a lot of requests" in this particular case can be up to 8 :) our responses channel most of the time will be filled with  up to 8 resource types in ADS mode and 1 resource type in non-ADS mode, hence I think it all should work similarly to the way it worked before.
In any case - it is easy to add additional for loop as a separate PR.

pkg/cache/v2/cache.go
```diff
@@ -37,14 +37,14 @@ type ConfigWatcher interface {
...
-	// error has occurred in the producer, and the consumer should close the
+	// Once the requested resources are available, they should be pushed into responses channel.
+	// If there are any failures - the error should be returned and the consumer should close the
 	// corresponding stream.
+	// Cache implementation should not close responses channel.
 	//
 	// Cancel is an optional function to release resources in the producer. If
 	// provided, the consumer may call this function multiple times.
-	CreateWatch(*Request) (value chan Response, cancel func())
+	CreateWatch(request *Request, responses chan<- Response) (cancel func(), err error)
```
jessicayuen, line 47: This isn't backwards compatible and will break a bunch of server implementations - possible to implement as a separate API? @kyessenov May be worth discussing major/minor versions for go-control-plane as adoption increases.

jessicayuen: For some more context, we have a management server implementation that relies on closing the channel in order to terminate the ack/nack flow from the client, forcing client-issued retries. Under this model, is that capability lost?

kyessenov: @jessicayuen There is a bunch of discussion why the change seems like a good idea [here](https://github.com/envoyproxy/go-control-plane/pull/353). Stable versions are important, but we kept it at 0.x.y because we felt it's better to give time to iron out issues in APIs. We should focus on getting 1.0 at some point.

There is an error returned in the interface stub. I think the difference is deferred signalling of errors. Does you cache implementation close the channel asyncrhonously?

fxposter:
@jessicayuen @jyotimahapatra you basically are talking about what I've described here as a "lost feature": https://github.com/envoyproxy/go-control-plane/pull/353#issuecomment-694842280. It is not used by public implementations and wondered if anyone uses it in private.

I have 2 ways of solving that problem, but let's step back a bit: the addition of `error` return parameter was done cause we needed to support "wrong type url" errors in both linear and mux implementations. Previously we only could signal a boolean flag - whether there is something wrong or not, not actually passing the error itself. So, the actual return type could be (cancelfunc, bool) instead of (cancelfunc, error). I just added error there, because it feels a bit more idiomatic and covers more cases (if server side wants to distinguish between errors, for example).

Given said the above - here are my 2 ~~cents~~ solutions:
1. send the `nil` response into responses channel and document that this signals "some" failure. this is basically returning the ability to say "there was an error, please close the connection" in a async manner, but we lose the `error` here, getting back to "`bool`". Personally, I am in favor of this solution, cause it makes the API simpler and currently `error` that was introduced is not actually used. The downside of this approach is that you cannot signal the server to close the connection "right now", it would need to process everything that is already buffered in response channel and only that will notice the error.
2. Change the CreateWatch signature to:
```CreateWatch(request *Request, responses chan<- Response, errors chan<- error) (cancel func())```
This would allow to separate error and response flow and from server implementation server would not need to wait to process all buffered items, but can catch error faster. The `errors` channel will be buffered with 1 item and the doc should say that cache implementations should do a non-blocking write to that channel to propagate the error.
2.1. There may be a simplification of (2) if we don't necessarily need to provide and error:
```CreateWatch(request *Request, responses chan<- Response, errors chan<- bool) (cancel func())```
or
```CreateWatch(request *Request, responses chan<- Response, errors chan<- struct{}) (cancel func())```
In this case `errors` might be renamed to "terminateConnection" or something like that.

Personally I am in favour of (1), but do understand that in some cases the fact that `nil` will be processed in order with other responses might be a problem.

File: pkg/server/sotw/v2/server.go [hunk 3]
```diff
@@ -195,78 +168,13 @@ func (s *server) process(stream Stream, reqCh <-chan *discovery.DiscoveryRequest
...
+			nonce, err := send(resp, typeUrl)
```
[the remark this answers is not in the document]
kyessenov: The way would be to return an error immediately, or as @fxposter mentions, we could send a token response as an erroneous value.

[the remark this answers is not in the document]
jyotimahapatra: sounds good...a killpill to represent closing channels sounds good, but it will involve much more code and maintenance for the control plane to do it. Right now, its easy to close channels. May be we can show how to perform that as part of the PR?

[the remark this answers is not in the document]
fxposter: How closing channel is different than sending nil to it?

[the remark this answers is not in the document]
jyotimahapatra: yeah..sending nil is ok. I was not sure what the special token would be.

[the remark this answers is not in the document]
jyotimahapatra: although, im not sure if sending nil is ok from a xds protocol stand point. The special killpill should be special enough and not a nil, in case nil is a valid xds response. In our private control plane, we dont return nil today, but there is no guardrails to say nil is a special value.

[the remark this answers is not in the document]
jyotimahapatra: may be another way to deal with this could be passing a cancel func , and the control plane implementations can call cancelFunc to denote closed channel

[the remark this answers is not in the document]
fxposter: https://github.com/envoyproxy/go-control-plane/blob/master/pkg/server/sotw/v2/server.go#L165-L168 - current implementation already handles `nil` as an error and propagates it further.

pkg/server/sotw/v2/server.go
jyotimahapatra: I have some questions about the buffered approach. At lyft we dont use ADS and use separate cds/eds/lds/rds channels. At scale EDS channel causes issues when the transport layer is overflowing and we sometimes drop messages. In this case we have devised mechanisms to ensure correctness by retry and dropping redundant updates for eds. In the PR, we are changing the model to use the same channel to use all xds. I feel at scale eds could starve other channels. Also, in case of a missed update, we have to device different mechanisms to ensure correctness for cds/rds/lds. This is an explicit ask for the control plane operators to implement. WDYT about this issue?

fxposter: If I understand it correctly - you won't use single channel in non-ADS case. this channel is created per stream, and each resource/resource type will create separate stream (hence separate channel).

pkg/server/sotw/v2/server.go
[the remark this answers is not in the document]
jyotimahapatra: i see.. using nil as killpill sounds good.

pkg/server/sotw/v2/server.go
jyotimahapatra: ah right..and we still have a way to specify a buffered channel with length 1.

pkg/cache/v2/cache.go
jyotimahapatra, line 47: can we name the param to singular `responseChan`

pkg/server/sotw/v2/server.go
fxposter: In this case - better chose 2. Cause it is possible to have a value written to channel before cancel() is called due to subsequent request from envoy coming.

pkg/server/sotw/v2/server.go
```diff
@@ -48,15 +48,34 @@ type Callbacks interface {
...
+	return func(s *server) {
+		s.callbacks = callbacks
+	}
+}
+
 // NewServer creates handlers from a config watcher and callbacks.
-func NewServer(ctx context.Context, config cache.ConfigWatcher, callbacks Callbacks) Server {
-	return &server{cache: config, callbacks: callbacks, ctx: ctx}
+func NewServer(ctx context.Context, config cache.ConfigWatcher, opts ...Opt) Server {
+	s := &server{cache: config, ctx: ctx, bufferSize: 16}
```
jyotimahapatra, line 67: can we have a way to override the buffersize?

fxposter: You would basically need to do
`server.NewServerAdvanced(rest.NewServer(...), sotw.NewServer(..., sotw.WithBufferSize(2)))`

fxposter: there is a helper above, working via Opt-s

jyotimahapatra: Sorry , i see you have `WithResponseBuffer`

pkg/cache/v2/cache.go
fxposter: Will do later. Given the things you've said above - the best thing to do now is to agree if the approach I propose can be accepted at all. Cause mimic-ing exactly the same behavior with different implementation can be very tricky and if you depend on existing behavior - I really wonder if we can proceed.
Cause my changes are the foundation for more consistent ADS experience in subsequent PRs and don't necessarily improve non-ADS.

fxposter: Cause with previous implementation - there was a guarantee that if envoy does the subsequent request - the previous one will be cancelled and there would be no attempts to read from previous channel - only the new one will be used. With my PR that is changing - you might get 2 responses in a row if cancel() is called after something is already written to the channel.

jyotimahapatra: I dont have major concerns with the approach, since we dont use ADS and im not sure of the different nuances. Is it possible to demonstrate the scenario you're changing with a test case?

fxposter: Not within this PR, but in the end I want to implement this: 
https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#eventual-consistency-considerations, ie: reorder responses to make clusters go first, then endpoints, etc...
This can be done in a hacky way from server side, but it is reasonable to do on a cache side (when you do SetSnapshot and pass whole set of resources - at that point the ordering should be applied)

So in the end I want this to pass:

```go
func TestResponseOrdering(t *testing.T) {
       c := cache.NewSnapshotCache(true, group{}, logger{t: t})
       ch := make(chan cache.Response, 4)
       c.CreateWatch(&discovery.DiscoveryRequest{TypeUrl: rsrc.EndpointType, ResourceNames: names[rsrc.EndpointType]}, ch)
       c.CreateWatch(&discovery.DiscoveryRequest{TypeUrl: rsrc.ClusterType, ResourceNames: names[rsrc.ClusterType]}, ch)
       c.CreateWatch(&discovery.DiscoveryRequest{TypeUrl: rsrc.RouteType, ResourceNames: names[rsrc.RouteType]}, ch)
       c.CreateWatch(&discovery.DiscoveryRequest{TypeUrl: rsrc.ListenerType, ResourceNames: names[rsrc.ListenerType]}, ch)
       if err := c.SetSnapshot(key, snapshot); err != nil {
               t.Fatal(err)
       }
       expectedTypeUrls := []string{rsrc.ClusterType, rsrc.EndpointType, rsrc.ListenerType, rsrc.RouteType}
       var typeUrls []string
       for range expectedTypeUrls {
               resp := <-ch
               typeUrls = append(typeUrls, resp.GetRequest().TypeUrl)
       }
       if !reflect.DeepEqual(typeUrls, expectedTypeUrls) {
               t.Fatalf("incorrect types order: got %v, expected %v", typeUrls, []string{rsrc.ClusterType, rsrc.EndpointType, rsrc.ListenerType, rsrc.RouteType})
       }
}
```

fxposter: Single channel would allow to guarantee that it pass in ADS mode and will make it much more likely to pass in non-ADS mode (but without any guarantees)

jyotimahapatra:
I think this PR is fine from my POV and with the improvements its making for ADS. We at Lyft can use the new api and make these changes work. @jessicayuen for comments.

fxposter:
@kyessenov Getting back to what we've discussed in previous PR, regarding intermediate slice of responses, this particular PR is subject to deadlock. If we (somehow) have filled the responses buffer and select will chose to receive another request, which for some reason envoy decides to send, simpleCache.respond() will block forever. :(
I don't really see any bullet-proof solutions for this except for intermediate slice of responses. Draining the channel cannot guarantee that, but makes it less likely. Another approach may be to try to write to the responses channel and if it fails - report error to the caller. If we are talking about this library's code - we have 2 callers: the cache.CreateWatch - if this errors - it will propagate error to grpc and close stream and cache.SetSnapshot - the error returned from it will need to be handled by control plane.

fxposter:
In current implementation this is basically solved by potentially unbounded growth of watches and corresponding channels. The most direct alternative to that is intermediate slice, which can grow indefinitely too.

pkg/cache/v2/cache.go
jessicayuen: In the xds-relay implementation, channels are closed asynchronously. My main concern with this approach is that the ownership is moving. The implementor no longer has control over the channel being open and runs the risk of writing to a closed channel. It's also not apparent what the implementor should do in the event the channel is closed - to return an error? I'll follow up more with the discussions over the weekend.

pkg/server/sotw/v2/server.go
```diff
@@ -302,88 +210,16 @@ func (s *server) process(stream Stream, reqCh <-chan *discovery.DiscoveryRequest
...
-				if values.listenerNonce == "" || values.listenerNonce == nonce {
-					if values.listenerCancel != nil {
-						values.listenerCancel()
-					}
-					values.listeners, values.listenerCancel = s.cache.CreateWatch(req)
+			typeUrl := req.TypeUrl
+			responseNonce, seen := values.nonces[typeUrl]
+			if !seen || responseNonce == nonce {
+				if cancel, seen := values.cancellations[typeUrl]; seen && cancel != nil {
+					cancel()
```
fxposter, line 217: @kyessenov To actually make guarantees about available space in channel - we can drain it here, after calling cancel. But it doesn't seem like a clean solution :(

fxposter: After thinking more about this and trying to implement draining - it does not look like a dirty hack anymore.
It allows us to actually get back to similar guarantees that we had in current implementation. And it allows to have buffer size exactly equal to number of types that will go through this channel.

fxposter:
@jyotimahapatra @jessicayuen @kyessenov Do we require cancel() being called if it's response was already sent?

fxposter:
Anyway, what I propose now:
1. remove error in return values, instead pass nil to responses channel
2. remove configurable buffer size and hardcode it to 8 items in ADS mode and 1 item in non-ADS mode.
3. set cancellations for particular type url to nil when response is sent - this is a breaking change, but for linear/simple caches not calling cancel when requests is already sent is 100% acceptable and I doubt if there are implementations that do require cancel being called if request was already processed and response being sent
4. add "draining" the responses channel after calling cancel

kyessenov:
One way to avoid dead-lock would be to allow returning a response synchronously as an output value. Not pretty, but would work I think.

fxposter:
It will work, but we would need to also keep error or bool for specifying
that there is an immediate error and support both delayed responses and
errors

вт, 29 сент. 2020 г., 20:00 Kuat <notifications@github.com>:

kyessenov:
We could implement a special error response. Then it could be sent immediately or deferred.

fxposter:
What `CreateWatch` signature do you see in that case?

kyessenov:
@fxposter Sorry, I wasn't clear enough. I was thinking of the following signature:
```go
CreateWatch(request *Request, responses chan<- Response) (resp Response, cancel func())
```

The assumption would be that is resp is not `nil`, then it is returned directly (`cancel` and the channel are not relevant in this case). We would add implement an ErrorResponse struct that implement Response, which would signal an error instead.

fxposter:
@kyessenov I still think that "draining" on request receive does provide more guarantees - we can have exact buffer size and provide non-blocking guarantees. If we return response immediately - we still may "overflow" the buffer, so we would need at least 2x capacity. The deadlock problem will be solved, though.

kyessenov:
Sure, draining never hurts. But there's always a chance that between draining and requesting a watch, the buffer gets filled. I think error response sounds useful to preserve the original behavior for some deferred errors and is a bit more efficient than using the channel indirectly.

fxposter:
If you drain after calling cancel() - you are guaranteed to have a single item space, aren't you. cause for current type url you no longer can have a value, cause you already called cancel() and did not call CreateWatch yet.

fxposter:
Can you check last 5 commits, please? :)

pkg/server/sotw/v2/server.go
```diff
@@ -218,6 +235,9 @@ func (s *server) process(stream Stream, reqCh <-chan *discovery.DiscoveryRequest
 			if !seen || responseNonce == nonce {
 				if cancel, seen := values.cancellations[typeUrl]; seen && cancel != nil {
 					cancel()
+					if err := processAll(); err != nil {
+						return err
+					}
```
kyessenov, line 225: If cancel is asynchronous, then the buffer could be full right here and CreateWatch would block. Imagine cache decided to push right before calling `cancel()`, gets `cancel()` but is already on the way to push it, which happens right after `processAll`.

fxposter: hm, I didn't think that cancel can be async

fxposter: Are there any real-world use-cases for async cancel?

fxposter: @kyessenov If we are dealing with async cancel-s/etc - maybe we should actually consider intermediate slice of responses? No size configuration, no need to deal with blocking stuff, can support both ads and non-ads with single code and no deadlocks.
Going the
```CreateWatch(request *Request, responses chan<- Response) (resp Response, cancel func())```
route basically fixes the deadlock problem, still making block on buffer overflow possible.

File: pkg/server/sotw/v2/server.go [hunk 7]
```diff
@@ -153,7 +107,11 @@ func (s *server) process(stream Stream, reqCh <-chan *discovery.DiscoveryRequest
 
 	// a collection of stack allocated watches per request type
 	var values watches
-	values.Init()
+	bufferSize := 1
+	if defaultTypeURL == resource.AnyType {
+		bufferSize = 8
```
kyessenov, line 112: Can you bring configurability for this? xDS is constantly evolving so it's not going to stay at 8 for long.

pkg/cache/v2/cache.go
```diff
@@ -37,14 +37,14 @@ type ConfigWatcher interface {
 	// CreateWatch returns a new open watch from a non-empty request.
 	// An individual consumer normally issues a single open watch by each type URL.
 	//
-	// Value channel produces requested resources, once they are available.  If
-	// the channel is closed prior to cancellation of the watch, an unrecoverable
-	// error has occurred in the producer, and the consumer should close the
+	// Once the requested resources are available, they should be pushed into responses channel.
+	// If there are any failures - nil should be written into the responses channel and the consumer should close the
```
kyessenov, line 41: Please replace `consumer` with `server` and format paragraph.

pkg/server/sotw/v2/server.go
```diff
@@ -162,7 +120,7 @@ func (s *server) process(stream Stream, reqCh <-chan *discovery.DiscoveryRequest
 	}()
 
 	// sends a response by serializing to protobuf Any
-	send := func(resp cache.Response, typeURL string) (string, error) {
+	send := func(resp cache.Response) (string, error) {
 		if resp == nil {
 			return "", errors.New("missing response")
```
kyessenov, line 125: This is an incorrect error message. `nil` is an intentional error not a missing response.

pkg/server/sotw/v2/server.go
```diff
@@ -302,89 +215,16 @@ func (s *server) process(stream Stream, reqCh <-chan *discovery.DiscoveryRequest
...
-						close(terminate)
-					}
-					if cancel, seen := values.cancellations[typeUrl]; seen && cancel != nil {
-						cancel()
+			typeUrl := req.TypeUrl
+			responseNonce, seen := values.nonces[typeUrl]
+			if !seen || responseNonce == nonce {
+				if cancel, seen := values.cancellations[typeUrl]; seen && cancel != nil {
+					cancel()
+					if err := processAll(); err != nil {
```
kyessenov, line 223: Why is processAll guarded by an individual type URL? What about other muxed type URLs?

kyessenov (commented):
In general looks fine. I'd appreciate if we could churn less since it looks like I have to review from scratch every time.

kyessenov:
@jyotima do you have any concerns about this version? My comments could be addressed in a follow up.

fxposter:
I will address tomorrow.

pkg/server/sotw/v2/server.go
fxposter: `processAll` will process all types that appear in responses channel, it just happens that if we are talking about synchronous `cancel` - to guarantee that there is a place in buffer - we can only drain if we cancel

pkg/server/sotw/v2/server.go
fxposter: I think we need 2 sizes here to be configurable - for ADS and for non-ADS separately, so that people who use both ADS and non-ADS within same process don't use big buffers while they are not needed. WDYT?

kyessenov: Sounds fine to me. "1" might be OK for non-ADS. ADS might have more than 8 types multiplexed.

jyotimahapatra: At Lyft, we use non ADS with buffer size 1 , but we sometimes experiment with other sizes. We would need configurability for the buffer size.

jyotimahapatra:
> @jyotima do you have any concerns about this version? My comments could be addressed in a follow up.

The changes lgtm. However I have some questions about the draining aspect. IIUC, every response from the go-control-plane will create a subsequent ack response from envoy.
In the non ads version, say we have a buffer size >1 for an eds channel. eds responses can get quite backlogged. I'm curious what happens if we enforce draining and each send queues up a new stream.

jyotimahapatra:
@kyessenov @fxposter Let me know if the question i mentioned above makes sense. I could be mistaken and I don't want to be blocking the PR.

fxposter:
I am not sure I understand the question. "send during drain queues up a new DiscoveryRequest from envoy". "queues up" where? In general there should be a single request per typeurl per stream. additional requests can queue up, but they will either be processed on next for loop iteration or skipped if they happen to be outdated by stale nonce.

jyotimahapatra:
The scenario i'm thinking about is this. Say we have a stream for eds. Per the xds protocol, each send from the server will cause an ack and a new request. 
If there were multiple eds responses to be drained, they will be send one after the other without waiting for the envoy ack. Is that a supported scenario in xds?

fxposter:
That's more of a question to semantics of cancel(). if cancel is synchronous, ie: after calling cancel you can guarantee that nothing will be written to responses channel for that particular watch - it guarantees that you always have at most 1 response in flight. Otherwise - in a bad case - you can have a situation where you've gotten multiple requests for single stream, each request finished their "watch" and written data to responses channel before actually doing cancel.

jyotimahapatra:
@fxposter Thanks for taking the time to explain the changes and I think it will be helpful to put in a picture about the flow before merging this in.
I will explain my question in some more detail below:
```		case req, more := <-reqCh:
			.....
			responseNonce, seen := values.nonces[typeUrl]
			if !seen || responseNonce == nonce {
				if cancel, seen := values.cancellations[typeUrl]; seen && cancel != nil {
					cancel()
					if err := processAll(); err != nil {
						return err
					}
				}
				values.cancellations[typeUrl] = s.cache.CreateWatch(req, values.responses)
			}
```

In the above block, whenever a new request comes in through `reqCh`, we always perform a `processAll` and then create watch using `CreateWatch`. Imagine during `processAll` there were was a queue of responses in eds.
They all will be sent first before `CreateWatch` happens.
```
	processAll := func() error {
		for {
			select {
			case resp := <-values.responses:
				if err := process(resp); err != nil {
					return err
				}
			default:
				return nil
			}
		}
	}
```
Due to the xds protocol, each of the sends inside `processAll` will cause a new request from envoy. 
Can you describe a bit about whether each of the new requests become a no-op, or whether they each will end up calling `CreateWatch` ?

This is how im thinking.
eds Channel has 6 responses v1 through v3.
envoy sidecar sends DiscoveryRequest(v0) -> ProcessAll sends DiscoveryResponse(v1), DiscoveryResponse(v2), DiscoveryResponse(v3) -> createsWatch(DiscoveryRequest(v0)) _This might be incorrect_
Each of the DiscoveryResponses:
DiscoveryResponse(v1) -> sidecar DiscoveryRequest(v1)
DiscoveryResponse(v2) -> sidecar DiscoveryRequest(v2)
DiscoveryResponse(v3) -> sidecar DiscoveryRequest(v3)

Im curious to know what happens to the three new DiscoveryRequests v1,v2,v3. Are they ignored?



Can you explain such a flow. We can put such an explanation in the README for future reference.

fxposter:
in my opinion - it will lead to CreateWatch. but I don't necessarily see what bothers you. If you remove processAll - the only difference would be that CreateWatch may block and deadlock (if we are talking about async cancel-s - it may still block or deadlock)

fxposter:
ie: you will send a response and envoy will send a new request. but you can send that response in exactly the same order without processAll, just on next iteration of for loop, aren't you?

jyotimahapatra:
The changes lgtm. We can merge them and I can try pulling these into our infra and try out next week.

kyessenov (approved):
Thanks @jyotima . I'm happy with this iteration as well, and some leftovers (customizable buffer size) are easy to fix. Thanks a lot @fxposter for this change!

fxposter:
Awesome! Thanks for merging! Sorry, was pretty busy this week.

jyotimahapatra:
@fxposter @kyssenov i tried to roll it out in our staging infra and faced downtime because pushing updates failed with this error: https://github.com/envoyproxy/xds-relay/blob/master/internal/app/orchestrator/orchestrator.go#L362

jyotimahapatra:
@fxposter are you using this change in your setup?

fxposter:
@jyotimahapatra sort of. I use a version with mostly equivalent implementation (https://github.com/fxposter/go-control-plane/tree/on-request-callback), which was written a bit before than this one was merged. After trying to understand what happens within xds-relay for 10 minutes - I'm a bit lost, there are quite a few things going on top of regular go-control-plane :)

The first 2 things I would start are:
1. https://github.com/envoyproxy/go-control-plane/pull/357/files#diff-d8ad87612eeb084487962f9097b132371b1cb8402432bc8ea190ab973fdba1feR149 - this basically means that there is no cancel being done for the typeUrl if there was a successful response (maybe xds-relay depends on those cancellations that were happening before, in this PR this code exists mostly for performance reasons).
2. https://github.com/envoyproxy/go-control-plane/pull/357/files#diff-d8ad87612eeb084487962f9097b132371b1cb8402432bc8ea190ab973fdba1feR222 - cancel is assumed to take place synchronously, maybe xds-relay does cancellations in an async-way?

jyotimahapatra:
@fxposter yes the xds-relay doesn't use simple snapshot and implements its own `CreateWatch`. The flow is straightforward, it maps an incoming request to a single key and requests the response from upstream control plane. Once the cache is filled up any update on uptream control planne updates the cache and propagates the response out to watches.

I believe the only breaking changes in this PR were the api signature of createWatch and cancelFunction being syncronous. 
`this basically means that there is no cancel being done for the typeUrl if there was a successful response` Is this a breaking change too?

jyotimahapatra:
@fxposter can you explain again about this ? `this basically means that there is no cancel being done for the typeUrl if there was a successful response`

fxposter:
When a response is sent (in process lambda) - we set cancellations for that typeUrl to nil. Initially there may be a cancel function related to this typeUrl. For built-in caches this is not a problem, cause they already do cleanups when sending successful response.

fxposter:
https://github.com/envoyproxy/go-control-plane/blob/master/pkg/cache/v2/simple.go#L122-L125 - this does the response and discards the watch.

jyotimahapatra:
Thanks for explaining. we should change it such that the cancel is always called for cleanup in all scenarios. Otherwise the cancel behavior is inconsistent.
In xdsrelay we keep the reference to the channel and clean it up in cancel func. I think that might explain the blocked channel stats we saw.

fxposter:
Reverting this part from GCP perspective would mean that we do "draining" on each request (basically because cancellations will always be present). From cache perspective - it does know when it sends the request and can do cleanup on it's side on that point of time to improve overall performance.

Otherwise - removing this line should just work: https://github.com/envoyproxy/go-control-plane/pull/357/files#diff-d8ad87612eeb084487962f9097b132371b1cb8402432bc8ea190ab973fdba1feR149

jyotimahapatra:
That would require each cache to implement an implicit behavior in addition to the cancel func. I think if each cache does an implicit behavior then cancelFunc loses relevance and can be removed altogether. 
Is the draining behavior applicable only for ads response ordering? since xdsrelay does not use ads, would it make sense to isolate the draining behavior only for ads?

jyotimahapatra:
Do we have a unit test around the draining behavior? A test could demonstrate the behavior in different use cases(eg, ads , eds. etc) and prevent future regressions too.

fxposter:
The behaviour for ads/non-ads is actually the same. Just for non-ads there multiple types exist at the same time, otherwise - it is treated exactly the same.

> That would require each cache to implement an implicit behavior in addition to the cancel func

Well, technically speaking - you are not supposed to pass additional responses to the envoy until you receive another request for it, are you?
Simple cache actually makes this guarantee. IE: if you get a request (for non-ads) - you provide a single response and wait for another request. With this guarantee - I don't really think that "assuming that watch is cancelled after response is sent" is additional behaviour.

jyotimahapatra:
> you are not supposed to pass additional responses to the envoy until you receive another request for it,

Sure we can make it the cache's responsibility to guarantee cleanup after each response. IMO the behavior should be consistent across all scenarios. Having cancel func in some cases and self cleanup in some others is hard to reason for the cache implementer. Also, its a bigger breaking change to move the cleanup functionality into cache. @kyessenov  what do you think about this scenario?

We are currently blocked on updating go-control-plane due to the issue. @fxposter Would you be able to help us with a fix ?

PR #359, opened 2020-09-24
cache: add support for xDS TTLs

### LINKED ISSUE #358: Support xDS TTLs
With https://github.com/envoyproxy/envoy/pull/13201 Envoy will have support for xDS TTLs. This issue tracks adding support for this in go-control-plane.

My thinking is to have the cache impls be aware of the TTLs and have it continuously push updates for TTL'd resources at some heartbeat interval (1/3 of the TTL duration? could be made configurable). This seems like it would be a fairly easy thing to do, but I haven't dug into the details yet.

snowp:
Adds support for xDS TTLs for SotW by introducing a goroutine per TTL'd
resource group which pushes the current config to open watches every 1/3
TTL time.

Fixes #358 

Signed-off-by: Snow Pettersen <snowp@lyft.com>


snowp:
Note: this is not complete due to the proto changes not having landed upstream Envoy, but this implements all the necessary bits for the cache to trigger watches based on the configured TTL. 

Mostly looking for high level feedback on this approach at this point, especially around whether we need to do something smarter around reducing the number of goroutines required for this feature. This approach is likely not feasible for large scale delta deployments, but since we're currently only supporting sotw this seems ok to me?

kyessenov:
I'm still a little confused about how this is supposed to work end-to-end. I thought TTL means that Envoy just "forgets" about some version of xDS it received, and needs to re-request. That implies the server needs nothing special to support TTL except pass it through. Server TTL can be orchestrated by just re-pushing a new, identical version.

snowp:
It does forget about it and would re-request, but if we rely on this then Envoy will flap between knowing about the resource and not knowing about it. The server will need to push the resource again before the TTL expires to ensure that the resource isn't forgotten. 

I guess one option would be to include a defer the deletion for some duration after the TTL expires? So that Envoy is able to request a new version of the resource before dropping the old one. I would increase the complexity in Envoy, but I'll raise it in the issue as an alternative solution and we can see what people think

kyessenov:
I think it makes more sense for Envoy not to delete expired resources if they are in use. Relying on the control plane to pro-actively re-push is fragile, and a lot to ask from all implementations.

snowp:
@jyotimahapatra @jessicayuen either of you wanna give this a look for the general approach? I've updated it to reflect the current implementation in Envoy, though some test coverage is missing (the v2/v3 handling makes v3 only features tricky to test, lmk what you think of the workaround I started using). Feel free to ping me if you need more context on TTLs

pkg/ttl/v3/ttl.go
```diff
@@ -0,0 +1,39 @@
+package ttl
```
jyotimahapatra, line 1: the workaround makes sense to me and we have done something similar at https://github.com/envoyproxy/go-control-plane/blob/master/pkg/resource/v3/resource.go . Can you check out if both workarounds can reside together in a common folder?

pkg/cache/v2/simple.go
```diff
@@ -110,6 +161,43 @@ func (cache *snapshotCache) SetSnapshot(node string, snapshot Snapshot) error {
...
+					return
+				case <-ticker.C:
+					wg := &sync.WaitGroup{}
+					cache.mu.Lock()
+
+					// It's possible for this goroutine to be canceled while we're waiting to grab the lock. In this case
+					// we want to avoid sending a heartbeat here and just end. If heartbeats should continue the another
+					// goroutine will have been created.
+					select {
+					case <-heartBeatCtx.Done():
```
jyotimahapatra, line 188: should we return when ctx is cancelled?

pkg/cache/types/types.go
```diff
@@ -9,6 +11,13 @@ type Resource interface {
 	proto.Message
 }
 
+// ResourceWithTtl is a Resource with an optional TTL.
+type ResourceWithTtl struct {
+	Resource Resource
+
+	Ttl *time.Duration
```
jyotimahapatra, line 18: is the Ttl config set in DiscoveryResponse? How will the control plane know what is a good  value to set for heartbeat interval?

pkg/cache/v2/simple.go
```diff
@@ -94,11 +100,56 @@ type snapshotCache struct {
...
+			// Spawn a goroutine here so that we can give up the lock before issuing responses. This prevents lock contention if the
+			// response channels aren't ready to receive the response.
+			go func(request *Request, response chan<- Response) {
+				defer wg.Done()
+				select {
+				case <-ctx.Done():
+					return
+				default:
+				}
+				cache.respond(request, response, resourcesWithTtl, version, true)
```
jyotimahapatra, line 146: sending hearbeatg is essentially sending an actual response and not a simple healthcheck style check. In a highly concurrent system, can this out of band send cause race condition bugs where the heartbeat code path overrides a more recent response? Also, what happens if the response is lost for any reason(flow control, etc) ?

jyotimahapatra (commented):
I echo Kuat's observation that having this complexity as part of control plane will cause multiple duplicate implementations in all control planes out there. The adoption of the feature will be delayed depending on control planes supporting this feature.

pkg/ttl/v3/ttl.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	discovery "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
+	"github.com/envoyproxy/go-control-plane/pkg/cache/types"
+	"github.com/golang/protobuf/ptypes"
+	"github.com/golang/protobuf/ptypes/any"
+)
+
+func MaybeCreateTtlResource(resource types.ResourceWithTtl, name string, heartbeat bool) (types.Resource, error) {
+	if resource.Ttl != nil {
```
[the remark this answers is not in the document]
snowp: No special meaning, so it would mean immediate removal

pkg/cache/types/types.go
snowp: Both TTL and heartbeat interval is specified by the user, so it's up to them to decide

pkg/cache/v2/simple.go
snowp: Not sure what you mean by a health check; the heartbeats special responses that only serve to refresh the TTL timeout for the resources and will never 

I could be wrong but my understanding of the cache/server interaction here is that the responses will be serially sequenced over the streams regardless of load due to the use of channels, so I don't see how an old heartbeat response could come before a newer update. We could add a safeguard to Envoy where we discard heartbeat-like responses if the version doesn't match if that makes things more palatable. 

Does GCP ever drop responses? That seems problematic in itself for a general purpose control plane as I'm pretty sure that would violate some of the eventual consistency expectations?

pkg/cache/v2/simple.go
snowp: yep good catch

snowp:
For more details on TTLs and heartbeats the Envoy configuration has some information: https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#ttl

This PR essentially implements two potentially distinct features: being able to set TTLs on resources (done by wrapping them in a Delta-style Resource wrapper) and a mechanism for sending lightweight sotw updates that only serve to refresh the TTL (aka heartbeats) that would be used when TTLs are used for rarely changing resources. 

Talking to @jyotimahapatra he suggested moving the heartbeat logic into server (to avoid having to implement it in each cache implementation), so I took a look. 

It can be pretty easily put in the server logic by having the server poll the cache for resources that requires heart beating, though this likely introduces more chances of heartbeats and real updates being processed out of order (since we lose out on the strict ordering of the response channel). A more specific cache lookup could probably be made (for the purposes of reducing lock contention on the cache) by tracking more state in the server (e.g. which types have TTL'd resources), but at some point we'd be duplicating state in the cache and server, so not sure if we want to go down that road. 

Moving it to the server puts the onus of configuring a heartbeat interval on the server, which might make it trickier to support variable intervals for different snapshots/types/etc., but that might be okay for now.

At the very least I'll push up a separate branch with the server driven heartbeating and anyone interested can take a look.

snowp:
Having spent some time trying to fit heartbeating into the SotW server.go code, I feel less confident that it will work without having to maintain a substantial amount of per resource state in the server. The key problem arises when the server polls the cache to get the resources that should be heartbeated but is unable to verify that the resource has been sent to the client or not; this opens up for a race where the first response containing a resource might be heartbeat response. The cache is unable to determine whether a resource has been sent to the client, as the use of a buffered channel means that handing over the request to the channel != the request being sent. 

I think that this + the fact that cache implementors might want to implement their own heartbeating strategies (e.g. elide heartbeat requests during frequent real updates) makes me lean towards keeping the responsibility of heart beating in the cache.

pkg/cache/v2/simple.go
```diff
@@ -94,11 +100,56 @@ type snapshotCache struct {
...
+		hash:              hash,
+		heartbeatRoutines: make(map[string]heartbeatHandle),
+	}
+}
+
+func (cache *snapshotCache) sendHeartbeats(ctx context.Context, wg *sync.WaitGroup, node string) {
+	snapshot := cache.snapshots[node]
+	if info, ok := cache.status[node]; ok {
+		info.mu.Lock()
+		for id, watch := range info.watches {
```
jessicayuen, line 116: Can this be done concurrently? I wonder if we'll run into performance problems doing this regularly for all watches.

jessicayuen (commented):
Personally I think it's OK for this logic to live within the cache. If I'm reading the code correctly, It doesn't appear that the implementation is forced to implement this TTL logic unless the mesh operator wants to utilize TTL - is that correct?

jessicayuen:
Also was there more conclusion discussed in Envoy surrounding the following?
> I think it makes more sense for Envoy not to delete expired resources if they are in use.

pkg/cache/v2/resource.go
```diff
@@ -82,13 +82,13 @@ func MarshalResource(resource types.Resource) (types.MarshaledResource, error) {
 
 // GetResourceReferences returns the names for dependent resources (EDS cluster
 // names for CDS, RDS routes names for LDS).
-func GetResourceReferences(resources map[string]types.Resource) map[string]bool {
+func GetResourceReferences(resources map[string]types.ResourceWithTtl) map[string]bool {
```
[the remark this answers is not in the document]
snowp: Oh yeah I like that, should simplify things quite a bit

pkg/cache/v2/simple.go
snowp: Hmm did you have any specific strategy in mind? We would still need to grab the lock to read/delete the watches so I'm not sure where we'd benefit from the concurrency here

Thinking about this some more, one improvement (somewhat in line with the TODO) would be to track the watches per type, which would allow us to look at which types have TTL'd resources and only iterate over the associated watches. A further optimization would be to track whether resources have seen an update recently and skip the heartbeat if so, which would reduce heartbeating during times of high cache activity. Not sure how much of this we'd want in an intial PR, so lmk what you think

pkg/cache/v2/resource.go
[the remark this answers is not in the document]
snowp: I spent some time trying to get this to work, the problem is that by adding a function here regular proto messages no longer satisfy the interface, which means that we end up having to wrap the resources in a struct regardless. I tried composing a `proto.Message` into `ResourceWithTtl` to make make them all just satisfy `types.Resource`, but it seems like the protobuf library does all kind of type conversions so just satisfying `proto.Message` isn't actually sufficient.

lmk if you have any ideas for how to get this working, otherwise the current approach seems like the simplest one (if verbose)

snowp:
> Also was there more conclusion discussed in Envoy surrounding the following?
> 
> > I think it makes more sense for Envoy not to delete expired resources if they are in use.

We've had support for expiring resources in use for EDS for a long time now (https://github.com/envoyproxy/envoy/pull/6477), so TTL just formalizes this for all resource types. Envoy is already able to operate without specific resources (i.e. what happens if you hit the initial_fetch_timeout), so TTL expirations should simply revert to that state. 

The whole purpose of xDS TTL is to ensure that certain resources that would do more harm than good are removed if the control plane is unavailable, which means that this logic has to live in Envoy and be controlled by control plane activity (hence the heartbeating). This is not meant as a general purpose TTL that applies to all resources, but just to resources that you don't want to go stale. ECDS, RTDS and EDS are all resource types where I've seen use cases where TTLs make sense.

pkg/cache/v2/snapshot.go
```diff
@@ -72,6 +83,30 @@ func NewSnapshot(version string,
 	return out
 }
 
+type ResourceWithTtl struct {
+	Resources []types.Resource
+	Ttl       *time.Duration
+}
+
+func NewSnapshotWithTtls(version string,
```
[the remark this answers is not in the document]
snowp: I don't think so, it all operates generically against xDS types (in line with the future of xDS where all types are opaque)

pkg/ttl/v3/ttl.go
snowp: The only thing I can think of would be to do something like `pkg/per_version/vX/<file>.go`, but that seems a little bit gross too. WDYT?

snowp:
@jessicayuen @jyotimahapatra I think I've responded/addressed all the comments now, ptal

snowp:
friendly ping for reviews :)

jyotimahapatra:
The changes lgtm.
2 final qs:
1. Is it possible to write an integration test. Can do in a subsequent PR
2. Is it possible to write a unit test showing that heartbeats send hollow responses ...may be in cache_test.go?

snowp:
@jyotimahapatra I think I was able to address both your concerns with the integration test that verifies that the heartbeat response is a hollow response, lmk if this is good enough.

@kyessenov do you wanna take a look as well?

pkg/cache/v2/simple.go
```diff
@@ -81,6 +85,9 @@ type snapshotCache struct {
 	// hash is the hashing function for Envoy nodes
 	hash NodeHash
 
+	// heartbeatRoutines keeps track of the active heartbeat goroutines
+	heartbeatRoutines map[string]heartbeatHandle
```
kyessenov, line 89: Why do we need so many go routines? Could we just have a single ticker that traverses the entire cache and re-pushes resources that aged?

pkg/cache/v2/simple.go
```diff
@@ -97,11 +104,56 @@ type snapshotCache struct {
...
+			if len(resourcesWithTtl) == 0 {
+				continue
+			}
+			if cache.log != nil {
+				cache.log.Debugf("respond open watch %d%v with heartbeat for version %q", id, watch.Request.ResourceNames, version)
+			}
+
+			wg.Add(1)
+			// Spawn a goroutine here so that we can give up the lock before issuing responses. This prevents lock contention if the
+			// response channels aren't ready to receive the response.
```
kyessenov, line 142: There's usually a spot in the response channel for every request. Can you explain why it could be not ready?

kyessenov (commented):
Seems reasonable but it's hard to reason about many go routines. Can we simplify it maybe by 1) turn Resource into struct, which hopefully reduce code changes and make it easier, 2) have a global heartbeat, which could keep track of how long ago resources were pushed?

pkg/ttl/v3/ttl.go
```diff
@@ -0,0 +1,47 @@
...
+			wrappedResource.Resource = any
+		}
+
+		return wrappedResource, deltaResourceTypeURL, nil
+	}
+
+	return resource.Resource, resourceTypeUrl, nil
+}
+
+func IsTTLResource(resource *any.Any) bool {
```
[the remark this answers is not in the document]
snowp: It's currently only used during tests to assert that a response emitted by the heartbeat routine is a TTL resource. I put it here since the V3 tests being generated from the V2 files makes per version tests a bit tricky, open to suggestions if you have a better idea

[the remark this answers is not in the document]
snowp: One option would be to only test this in the V3-only integration test and remove the tests from the versioned cache tests, wdyt?

pkg/ttl/v2/ttl.go
```diff
@@ -0,0 +1,17 @@
...
+
+// Helper functions for interacting with TTL resources for xDS V2. Since TTL resources are not supported for V2, these are
+// essentially noops.
+
+func MaybeCreateTtlResourceIfSupported(resource types.ResourceWithTtl, name string, resourceTypeUrl string, heartbeat bool) (types.Resource, string, error) {
+	return resource.Resource, resourceTypeUrl, nil
+}
+
+func IsTTLResource(resource *any.Any) bool {
+	return true
```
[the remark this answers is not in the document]
snowp: Thinking about it some more this should probably be false, it just so happens to only be used to test things we expect to be TTL responses, so this made the tests pass. Depending on the resolution of my other comment I'll update this

pkg/cache/v2/simple.go
snowp: It just felt the easiest at the time, I'll give the single goroutine a go

pkg/cache/v2/simple.go
snowp: I think this is a holdover from before when the channel was blocking (I think they used to be blocking?), and tests weren't set up to consume from the channel. I removed the go routine and tests pass, so I don't think this is actually necessary

pkg/ttl/v3/ttl.go
```diff
@@ -0,0 +1,47 @@
...
+			Name: name,
+			Ttl:  ptypes.DurationProto(*resource.Ttl),
+		}
+
+		if !heartbeat {
+			any, err := ptypes.MarshalAny(resource.Resource)
+			if err != nil {
+				return nil, "", err
+			}
+			any.TypeUrl = resourceTypeUrl
```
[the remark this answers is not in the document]
snowp: This preserves the behavior from the existing code where we always set the type URL to what the request asked for, I think this is used to allow marshalling V2 resources in response to a V3 request. I can add a comment to make this clearer

pkg/cache/v2/resource.go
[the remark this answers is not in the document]
snowp: I gave this a go, the problem here is that it means that a regular proto no longer passes as a `Resource` (at least from what I could tell), which means that all callers that currently pass `[]Resource{proto1, proto2}` would have to update it to `[]Resource{Resource{Message: proto1}, Resource{Message: proto2}`. Not sure if it would be worth the churn to change? wdyt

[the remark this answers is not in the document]
kyessenov: I think you could do `[]Resource{{proto1}, {proto2}}` which I can live with. What do others think? I think it makes sense to align Resource with xDS Resource proto which has TTL field. But I'm fine either way.

pkg/ttl/v2/ttl.go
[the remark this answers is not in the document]
kyessenov: If it's used only for testing, then let's just put that as comment and not worry about its efficiency.

pkg/cache/v2/resource.go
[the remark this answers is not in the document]
snowp: Omitting the field name works when there's only one field e.g. `type Resource struct { proto.Message }`. Once we introduce a second one for the TTL, they now need to be specified. 

I would expose the a function to get the TTL but that means having to embed the TTL within the proto, which then forces the user to use the `Resource` proto wrapper (not to be confused with the Go type) themselves to specify the TTL. This would be an option: we could provide a helper function to make it easier to create the proto wrapper and rely on it for holding the TTL configuration. LMK what you all think (cc @jyotimahapatra @jessicayuen)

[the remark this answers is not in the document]
snowp: I'll note that this adds a bunch of complexity around anything that cares about the actual resource, like consistency checks, as we need to handle potentially unwrapping the wrapper type to see the real resource.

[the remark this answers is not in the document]
snowp: Ok after much more playing around it seems like embedding `proto.Message` breaks marshalling, since the proto library makes use of some cute interface casts to do marshalling, which ends up failing when embedding the interface. As such I'd propose leaving things as is, unless someone has any ideas for how to work around this

snowp:
@kyessenov ptal, let me know if I've addressed everything

snowp:
ping @kyessenov

pkg/cache/v2/simple.go
kyessenov: Is this field needed now? I don't see it being updated.

pkg/integration/ttl_integration_test.go
```diff
@@ -0,0 +1,139 @@
+package integration
```
kyessenov, line 1: I think it would be nice to have an end-to-end example integration test with actual envoy. TTLs are very confusing, so a worked out example with envoy plus a control plane config would help understand it. This doesn't need to happen in this PR since envoy is pegged to some old version in CI, but a follow up would be helpful.

kyessenov (commented):
Can you also update CHANGELOG? This is an API change so it'll require work from consumers.

snowp:
@kyessenov Should be updated now, will follow up with an integration test as I get around to it

kyessenov (approved):
Thanks, and sorry it took so long. I'm fine with the change, so feel free to merge and iterate.

PR #363, opened 2020-10-14
api: add custom buffer size for ADS, docs

kyessenov:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Release v0.9.8
+
+### Changed
+
+- Interface between the server and the cache has been changed to support explicit ordering by the cache.
+
 ## Release v0.9.7
 
 ### Added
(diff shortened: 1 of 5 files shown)


fxposter:
I was thinking about that. Let's do `WithBufferSize(xdsSize, adsSize int)` to try to adopt cases where xds would require additional buffer if using custom caches that delay cancel() calls?

fxposter:
And set defaults to 1 for xds and 8 for ads

kyessenov:
Sure. I'm waiting if anyone reports an issue with the previously merged change in case we have to revert. If all is good, let's move on with this and other changes.

jyotimahapatra (approved):
Thanks for the change! lgtm.

jyotimahapatra:
I am still in the process of pulling the changes since the e2e tests are failing. I will report back when i have it pulled into our infra.

fxposter:
LGTM too

kyessenov:
@jyotimahapatra LMK if there is any issue with the latest changes. Otherwise, I think it's time to merge.

jyotimahapatra:
@kyessenov im currently bringing in gcp update into xdsrelay https://github.com/envoyproxy/xds-relay/pull/181
We should merge this PR irrespective of my effort.

PR #369, opened 2020-11-20
Add test for drain

jyotimahapatra:
Based on the conversation https://github.com/envoyproxy/go-control-plane/pull/357#issuecomment-728318405 , this PR fixes the cancellation and adds a test case to enforce the behavior.

Signed-off-by: Jyoti Mahapatra <jmahapatra@lyft.com>

### DIFF
--- a/pkg/server/sotw/v2/server.go
+++ b/pkg/server/sotw/v2/server.go
@@ -176,7 +176,6 @@ func (s *server) process(stream Stream, reqCh <-chan *discovery.DiscoveryRequest
 		}
 		typeUrl := resp.GetRequest().TypeUrl
 		values.nonces[typeUrl] = nonce
-		values.cancellations[typeUrl] = nil
 		return nil
 	}
 
--- a/pkg/server/sotw/v3/server.go
+++ b/pkg/server/sotw/v3/server.go
@@ -177,7 +177,6 @@ func (s *server) process(stream Stream, reqCh <-chan *discovery.DiscoveryRequest
 		}
 		typeUrl := resp.GetRequest().TypeUrl
 		values.nonces[typeUrl] = nonce
-		values.cancellations[typeUrl] = nil
 		return nil
 	}
 
--- a/pkg/server/v2/server_test.go
+++ b/pkg/server/v2/server_test.go
@@ -31,6 +31,7 @@ import (
 	rsrc "github.com/envoyproxy/go-control-plane/pkg/resource/v2"
 	"github.com/envoyproxy/go-control-plane/pkg/server/v2"
 	"github.com/envoyproxy/go-control-plane/pkg/test/resource/v2"
+	"github.com/stretchr/testify/assert"
 )
 
 type mockConfigWatcher struct {
@@ -43,17 +44,16 @@ type mockConfigWatcher struct {
 func (config *mockConfigWatcher) CreateWatch(req *discovery.DiscoveryRequest, out chan<- cache.Response) func() {
 	config.counts[req.TypeUrl] = config.counts[req.TypeUrl] + 1
 	if len(config.responses[req.TypeUrl]) > 0 {
-		out <- config.responses[req.TypeUrl][0]
-		config.responses[req.TypeUrl] = config.responses[req.TypeUrl][1:]
+		for _, r := range config.responses[req.TypeUrl] {
+			out <- r
+		}
 	} else if config.failWatch {
 		out <- nil
-	} else {
-		config.watches += 1
-		return func() {
-			config.watches -= 1
-		}
 	}
-	return nil
+	config.watches += 1
+	return func() {
+		config.watches -= 1
+	}
 }
 
 func (config *mockConfigWatcher) Fetch(ctx context.Context, req *discovery.DiscoveryRequest) (cache.Response, error) {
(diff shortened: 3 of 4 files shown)


jyotimahapatra:
@fxposter @kyessenov Please review

fxposter:
Sorry, was busy. Will check today or tomorrow. Not sure I follow into the test without digging more.

kyessenov:
Sorry, I missed the discussion earlier. I'm a bit confused why cancel should be called after response is supplied.

kyessenov:
Let me try to understand again what are the cases:
1. Response never produced by the cache. Cancel is called on re-request, and this seems to be OK.
2. Response produced and read by server. It seems unnecessary to call "cancel" since there is no watch anymore. It could be called synchronously when put into channel.
3. Response produced but not read by server. It feels like this would could leak, but I can't come up with a scenario when this happens.

jyotimahapatra:
> why cancel should be called after response is supplied.

Based on current impl cancelFunc from `CreateWatch` is always called on the next request. The cache implementor(for ex, xdsrelay) does not really know when its called. We just assume that it will be called at some point.
In the current state, we are making cancelFunc nil and that causes cancel not to be called on next DiscoveryRequest. This causes previous watches leaking.
If there is an implicit assumption that the cache implementor should always cleanup the watch after sending a response(I think that is the suggestion), we should get rid of the cancelFunc altogether.

jyotimahapatra:
Or may be  i was mistaken to take a dependency on cancelFunc to be always called. If this was a wrong assumption I could change the code to match the expectation. My point is that the new impl is breaking an existing assumption and I wanted to brainstorm how to convey the assumption that the cache implementor cancels watches in different ways in different situations. 
We also missed adding a test when we made the changes last time. Either way, I would like to make the assumption documented by a test case.

kyessenov:
Can you try xdsrelay with cancel done by response like I suggested? We should absolutely make this clear in documentation and tests. But first, let's see if we can unblock this way, or go for revert.

jyotimahapatra:
Sure. I will try that soon.

fxposter:
I think there is/was misunderstanding about "draining" part and my whole in general.
My assumptions were the following:
1. we do not actually assume that there will be multiple responses for a single request from the go-control-plane. simple cache was entirely built around this assumption. I don't remember exactly, but I think https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol does not necessarily deny multiple responses for single envoy request, but overall the idea is to respond with a single response for single request.
2. everything related to "draining" was implemented due to the fact that we no longer can "just ignore previous channel and create a new one" in CreateWatch, cause the channel is reused for the whole stream, which basically adds a case where CreateWatch already succeeded and put response into that channel, but envoy sent a new request not waiting for that response (and this is actually a valid behavior, described in doc above) - in that case the situation may behave like this: if the response for new request is ready immediately - it will not be able to put data into the channel, cause it may be full, causing the deadlock. IE: we solve a problem of a response being ready, but not yet sent, while new request arrives not waiting for that response. The "race" here between 2 events: new request comes in, while old request is not yet cancelled and finishes it's processing. The solutions discussed were: add "non-blocking channels" via an intermediate slice or doing what we did -  basically before accepting the new request we process already fulfilled responses to leave the room for an immediate response to fit into the channel.
What is not handled - "delayed cancels", ie: if calling `cancel()` still allows the fulfilled response for previous request to get into the channel. This one is not solve-able (at least I don't see how) with the solution that we came up with, I guess. But it is debatable if those "delayed cancels" do actually exist in real-world cache implementations and whether we should think about them.

fxposter:
And - if we don't allow multiple responses for the same request - the question about the necessity of calling cancel is valid. Most of the time implementations will do the cleanup after the successful response to prevent those multiple responses anyway, hence, we can skip doing those cancel-s and greatly benefit of not needing to drain all the time.

fxposter:
If we require multiple responses to be a valid thing for a single request - currently you can increase buffer size and *hope* that the number of such responses will not exceed that limit, which I personally don't like very much, but from practical standpoint it might make sense (thought, from theoretical POV - this is obviously a bad thing).
So if we require such behavior - I would go a route that I initially proposed with an "unbounded channel" via intermediate slice.

fxposter:
Regarding this PR - it seems like you both are far better than me as developers, cause I can reason about the code in server.go/sotw.go, but looking at server_test.go is quite a challenge for me, cause it is very hard to understand what TestDraining actually doing :)

kyessenov:
Envoy should only have only one outstanding request per resource type on ADS. Multiple responses could be supplied but the way nonces work, only one response wins.

I admit I don't quite follow the TestDrain as well. There's a general problem of lack of good conformance tests for xDS. There're a lot of quirks that are hard to test. The guilt is on both sides, client and server :)

jyotimahapatra:
I understand that the test is complicated. Iy was the closest i got to demonstrate multiple DiscoveryRequests and response and the drain sequence. I'm very open to suggestions for how to make it simple and easier to read.
There are 2 goroutines. One behaves as envoy and keeps sending a request for every response received and the other used for asserting on the responses. I imagine we just need one and the assert can happen in the main test flow.

I am going to try @kyessenov 's suggestion to clear the state after sending each response on the channel and report my findings.

PR #374, opened 2020-12-02
server: guard response channel sharing with an option

kyessenov:
Restore unique channel per watch behavior to support legacy clients that may block if the channel is shared.
It's likely other clients do not expect shared channel as well, so the default is the legacy behavior.

cc @fxposter

### DIFF
--- a/pkg/server/sotw/v2/server.go
+++ b/pkg/server/sotw/v2/server.go
@@ -59,6 +59,7 @@ func NewServer(ctx context.Context, config cache.ConfigWatcher, callbacks Callba
 		ctx:           ctx,
 		xdsBufferSize: 1,
 		muxBufferSize: 8,
+		unordered:     true,
 	}
 	for _, opt := range opts {
 		opt(out)
@@ -84,12 +85,25 @@ func WithXDSBufferSize(size int) ServerOption {
 	}
 }
 
+// WithOrdered enables reuse of the channel between watches which preserves the
+// ordering of responses in the server stream. Because of sharing, there is a
+// risk of a blocked channel if watch is responded twice (e.g. if cancellation
+// is asynchronous).
+// The default is to create a new channel of size 1 for all well-known type
+// watches.
+func WithOrdered() ServerOption {
+	return func(s *server) {
+		s.unordered = false
+	}
+}
+
 type server struct {
 	cache         cache.ConfigWatcher
 	callbacks     Callbacks
 	ctx           context.Context
 	xdsBufferSize int
 	muxBufferSize int
+	unordered     bool
 
 	// streamCount for counting bi-di streams
 	streamCount int64
@@ -108,6 +122,14 @@ type watches struct {
 	responses     chan cache.Response
 	cancellations map[string]func()
 	nonces        map[string]string
+
+	// Per-type channels for unordered watches
+	endpoints chan cache.Response
+	clusters  chan cache.Response
+	routes    chan cache.Response
+	listeners chan cache.Response
+	secrets   chan cache.Response
+	runtimes  chan cache.Response
 }
 
 // Initialize all watches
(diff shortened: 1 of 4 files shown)


fxposter:
Wouldn't it be easier and clearer to have a single unbounded channel?

kyessenov:
@fxposter How do you make an unbounded channel in go? I think bad clients can still mess up a channel by spamming it after server is done with the watch. Eventually, we want more discipline with the channel but we can't assume every client behaves well.

fxposter:
like in initial PR - via additional slice and goroutine. basically in previous version you could have "not limited" amount of channels. now you can have unbounded channel.

fxposter:
Another benefit (or curse :)) might be that you can actually do multiple non-blocking writes per single envoy request, which you previously could do in theory, but it was not done in SnapshotCache.

kyessenov:
@fxposter I want to migrate folks to using the new API. This is just a fail safe in case there is a bug in a client. Eventually, we'll just switch the default and then remove the unordered mode.

kyessenov:
@fxposter the unbounded channel would work, but think about what happens with a spammy client:
1. Before, the client would block in some goroutine trying to a full channel (server is no longer listening). Hopefully, the client has some context done version to terminate it.
2. After, the server would accumulate a number of stale (potentially more now that channel is not blocking) responses in an array. These would go down the wire eventually, which also makes Envoy process it and a CPU spike in the proxy.

Both options are not good, but 1) is a client responsibility which needs to be fixed to support new API anyways.

fxposter:
Both are correct statements :) Given the fact that we don't know how caches implement stuff - it indeed can be either a good or a bad thing.
PR LGTM.
@jyotimahapatra can you share your state of xds-relay. I don't follow it at all, but looking at this PR https://github.com/envoyproxy/xds-relay/pull/205 I guess you can work with new implementation anyway? :)

jyotimahapatra:
@fxposter @kyessenov My xds-relay PR to update gcp is now working functionally since the last 12 hours.
However, I noticed a significant increase in latency . Some responses take >1s.
Screen Shot 2020-12-02 at 9 50 04 AM

jyotimahapatra:
I also performed a stress test to see how quickly membership converges.
without go-control-plane update
Screen Shot 2020-12-02 at 11 18 11 AM


with go-control-plane update
Screen Shot 2020-12-02 at 11 16 29 AM

There could be a bug with xdsrelay due to how watch cleanup happens along withe concurrent updates on watches. However the new api model is definitely causing some difficulty in migrating to.

pkg/server/sotw/v2/server.go
```diff
@@ -256,7 +304,30 @@ func (s *server) process(stream Stream, reqCh <-chan *discovery.DiscoveryRequest
 						return err
 					}
 				}
-				values.cancellations[typeUrl] = s.cache.CreateWatch(req, values.responses)
+				value := values.responses
+				if s.unordered {
```
jyotimahapatra, line 308: i think in the older model, cancelWatch was guarenteed to be called before createWatch everytime

fxposter:
I think the easiest way to check changes related to cancels is to remove that single line that assigns nil to cancellations. At least for now.

pkg/server/sotw/v2/server.go
kyessenov: I think it still happens before CreateWatch at line 300?

kyessenov:
@fxposter Yeah, we can try to do that here -- call cancel on response as well. Let me push a change.

kyessenov:
@jyotima Can you try to pick up this PR and see if it helps with stress test? We can also try that option to cancel on response I just added. Sorry it's so hard to update.

jyotimahapatra:
Will try it out soon

jyotimahapatra:
@kyessenov @fxposter 
I tried the suggested changes and looks like some watches dont get the updates and membership diverges. I guess we need more observability/logging to investigate this.
Screen Shot 2020-12-04 at 1 55 26 PM

PR #375, opened 2020-12-04
Revert the new draining api

jyotimahapatra:
While trying out the new draining api https://github.com/envoyproxy/go-control-plane/pull/357 , we realized that we are not able to make xdsrelay work with it(results described in https://github.com/envoyproxy/go-control-plane/pull/374) 
I propose to revert them to keep master clean and bring the changes back with more testing in production environments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,5 @@
 # Changelog
 
-## Release v0.9.8
-
-### Changed
-
-- Interface between the server and the cache has been changed to support explicit ordering by the cache.
-
 ## Release v0.9.7
 
 ### Added
--- a/pkg/cache/v2/cache.go
+++ b/pkg/cache/v2/cache.go
@@ -37,10 +37,10 @@ type ConfigWatcher interface {
 	// CreateWatch returns a new open watch from a non-empty request.
 	// An individual consumer normally issues a single open watch by each type URL.
 	//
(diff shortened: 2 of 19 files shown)


kyessenov (approved):
FYI @fxposter we'll try to bring it back but need to make a release soon.

fxposter:
Huh, I'm a bit confused about all this stuff. What do you think should be the next steps?
IE: I am interested in "linearizing" cache responses for ADS, when it is possible and minimizing their effect for XDS. My implementation (probably) queues the responses, which would be discarded in original implementation. That's how I interpret graphs, but I don't really understand how that happens, cause in reality backpressure is done at envoy layer anyway.

fxposter:
BTW, given my previous commits in place it is possible to have my own SOTW implementation to live in my own repo and just be private, while still using this library from upstream, without needing to have a full fork. So I am not sure if it's worth pursuing such changes at all at this point.

fxposter:
What is even more strange is the fact that "unordered" implementation by @kyessenov should actually have the same characteristics as the original one, due to the fact that the only real difference between those is the place where channels are created - in server or in cache code...

kyessenov:
@fxposter We could not find figure out why xds-relay was behaving differently. We'll have to add instrumentation and to get better sense probably. It's a live system, so we can't really break it.

PR #379, opened 2020-12-08
cache: add snapshot support for ECDS

snowp:
This adds support for setting ECDS resources on the snapshot, which allows using ECDS with the simple cache.

Note that ECDS is not supported in V2, so we somewhat awkwardly use the V3 protos within the V2 code.


kyessenov (approved):
Thanks. I have used "opaque" type version but it doesn't cost much to add it explicitly. Would be nice to have an integration test, but not going to hold it for that.

snowp:
Happy to follow up with an integration test. I couldn't find and prior art for integration tests, or am I missing something?

kyessenov:
The integration test started from here https://github.com/envoyproxy/go-control-plane/blob/master/build/integration.sh is how we test envoy binary. We have to run it separately since envoy sometimes breaks xDS protocol.

PR #387, opened 2021-01-20
Incremental xDS

alecholmez:
This is an alternate PR to https://github.com/envoyproxy/go-control-plane/pull/355

The git history was too difficult to run through since some commits missed signing. Opening this instead.


nitgoy:
Hey @alecholmez, any updates on this PR?

alecholmez:
@nitgoy sorry no updates! No known issues either, I know a few people using this fork in projects. Still waiting on a review from the maintainers

snowp:
@alecholmez Do you think it would be possible to split this up into smaller PRs? The PR being this big is likely contributing to review delays, it's quite intimidating

alecholmez:
@snowp certainly!! I can break it down and slowly get it approved. Anything to get those code in

Phlak106:
What's the best way to get updates on this? Would love to see this go in!

mattklein123:
@snowp and I will make sure this gets merged, but we need to split this up into more easily reviewable chunks to do so. Thank you!

alecholmez:
@snowp @mattklein123 Thank you guys for pushing forward on this PR! For your guys convenience I plan to break down the pull request into 3 main sections:
- Cache
This will will be all code relating to the snapshot cache and should be merged first
- Server
This is all code relating to the gRPC server itself and any server functions required to make that work. This has a dependency on the cache and should go in second
- Integration Testing
This will be the last PR. It contains the full end to end integration under `pkg/test` that enables incremental. This obviously requires both the cache and server implementation to go in first so it will be the last. All previous PRs will have unit tests for validation sake but the integration test will come last.

If you guys have suggestions on whats easiest for you let me know! If not I'll proceed with these plans and move forward from there. Thanks again guys

snowp:
Looking at this PR there seems to be other smaller components that could be split out, like the proto changes (why are we couching cdsd.pb.go?) and smaller somewhat unrelated changes like adding additional logging to the Makefile.

Other than that the logical split you've outlined makes sense to me!

fxposter:
I think we don't really need to merge snapshot cache without server/etc - it doesn't make much sense, cause you cannot use that functionality.
What instead can be merged is interfaces for both cache and server without implementations at all. Given the fact that after REST/SOTW separation you can easily create and test even delta (by using NewServerAdvanced) if all the interfaces are in place. And then you can push default implementation.

alecholmez:
@fxposter i understand what you're saying, I guess the end result will be the same regardless I just want to make it easy for code review. But very true i like the idea of the interfaces going in first then the implementations after

alecholmez:
Here is the first part of the PR breakdown: https://github.com/envoyproxy/go-control-plane/pull/408

This includes both interface definitions for the cache + server. Under 1k lines of code so hopefully easy to review if you don't count the code dupe from v3.

alecholmez:
Just FYI for those following this PR, I will be closing this once the other implementation PRs go in. This is just being held open for reference.

Issue #398, opened 2021-02-25
Are we planning to fully support ECDS in go-control-plane?


delong-coder:
Hi folks,

I want to use ECDS in our project through go-control-plane, but I think ECDS is not fully supported yet.

I found a PR of "cache: add snapshot support for ECDS ": https://github.com/envoyproxy/go-control-plane/pull/379/files, but [pkg/server/v3/server.go](https://github.com/envoyproxy/go-control-plane/blob/main/pkg/server/v3/server.go) haven't implemented the [ExtensionConfigDiscoveryServiceServer](https://github.com/envoyproxy/go-control-plane/blob/main/envoy/service/extension/v3/config_discovery.pb.go) interface, so I can't register ECDS in this server.
Also, I think the [pkg/server/sotw/v3/server.go](https://github.com/envoyproxy/go-control-plane/blob/main/pkg/server/sotw/v3/server.go) needs to add some logic to deal with ECDS as well.
I'm not sure that's all the places to add ECDS support.

So I want to know are we planning to fully support ECDS?
Please correct me if I said anything wrong.

Thanks!

PR #400, opened 2021-03-02
Add ADS support for ECDS

delong-coder:
This adds support for setting ECDS resources in the ADS.

Note that ECDS is not supported in V2, only changed V3 server code.

Related issue: https://github.com/envoyproxy/go-control-plane/issues/398

To add ECDS, I have to change the pkg/server, I'm not sure why it can't be modified.

Please help review. Thanks!


fxposter:
1. v3 is generated from v2, so if you change v3 directly - it will be overwritten by running ./create_version.sh
2. given the fact that v2 and v3 have a generic handling of all resources and special-casing only a handful of them, do you really need to special-case this one? ie: if you just remove your code from server.go it should work out of the box.

delong-coder:
> 1. v3 is generated from v2, so if you change v3 directly - it will be overwritten by running ./create_version.sh
> 2. given the fact that v2 and v3 have a generic handling of all resources and special-casing only a handful of them, do you really need to special-case this one? ie: if you just remove your code from server.go it should work out of the box.

Sure, I got your point, I will change the v2 code, then let ./create_version.sh generate the v3 code. Thanks.

delong-coder:
Sorry, @fxposter , I tried adding my code into the v2 server, but it didn't work, because the ECDS is only enabled in V3, it needs the v3 discovery server.
I can't compile the ECDS with server v2 code unless I add v3 rest and sotw server to our /pkg/server/v2/server.go, that would be super ugly.
So maybe it's time to separate those 2 versions, stop creating v3 from v2 code, otherwise, we may never support the ECDS. 
@snowp  @kyessenov @jyotimahapatra

delong-coder:
Thanks @peergynt ! 
So I think it's a common pain for lacking supporting ECDS.
Shall we go forward?

pkg/server/sotw/v3/server.go
```diff
@@ -73,26 +73,29 @@ type Stream interface {
...
+	runtimeCancel         func()
+	extensionConfigCancel func()
+
+	endpointNonce        string
+	clusterNonce         string
+	routeNonce           string
+	listenerNonce        string
+	secretNonce          string
+	runtimeNonce         string
+	extensionConfigNonce string
```
fxposter, line 98: If you remove changes in this file everything will still work out of the box, isn't it?

delong-coder: Yes, the ECDS will still work, but we can't use it with the ADS server.

delong-coder: I mean pkg/server/v3/server.go, not quite sure about sotw server.

fxposter: yes, I am talking about this specific file :) maybe these can go via generic `responses`: https://github.com/envoyproxy/go-control-plane/pull/400/files#diff-0173d7909b484a388922cc5438a5b94b78b72fe780a531a0ed246bfa940a69f8R101

All other files are actually needed and the problem is - with current build setup it cannot be merged

delong-coder: Not sure lacking extension config watcher has other potential effects or not.

Yes, other files are needed, can I change the current build setup and stop copy those files from v2 code?
I think that should be approved by someone.

fxposter: I think nowadays it's up to @kyessenov maybe?

snowp:
FYI this should be much easier to land once we land #412

delong-coder:
> FYI this should be much easier to land once we land #412

Sounds good!

alecholmez:
@delong-coder Has this been fixed for the v2 removal?

alecholmez:
@snowp I think this is good? Just another type for SOTW

pkg/test/resource/v3/resource.go
```diff
@@ -288,6 +288,35 @@ func MakeRuntime(runtimeName string) *runtime.Runtime {
...
+			Rds: &hcm.Rds{
+				ConfigSource:    rdsSource,
+				RouteConfigName: route,
+			},
+		},
+		HttpFilters: []*hcm.HttpFilter{{
+			Name: wellknown.Router,
+		}},
+	}
+	pbst, err := ptypes.MarshalAny(manager)
```
alecholmez, line 309: Can we move away from ptypes? It's been deprecated. I believe the replace here is:

```
anypb.New()
```

pkg/server/v3/server_test.go
```diff
@@ -364,10 +378,13 @@ func TestFetch(t *testing.T) {
 		t.Errorf("unexpected empty or error for listeners: %v", err)
 	}
 	if out, err := s.FetchSecrets(context.Background(), &discovery.DiscoveryRequest{Node: node}); out == nil || err != nil {
-		t.Errorf("unexpected empty or error for listeners: %v", err)
+		t.Errorf("unexpected empty or error for secrets: %v", err)
```
alecholmez, line 381: Thanks for the catch :)

alecholmez (changes requested):
@snowp This addressed ADS support for ECDS, but I believe if we move to my refactor of SOTW to use the generic responses paradigm then we don't have to keep adding new hardcoded types for ADS support. Though I'm not entirely sure if the extra go routine overhead is worth it. Would it be worth comparing that refactor vs what we have now with a benchmark?

snowp:
>Would it be worth comparing that refactor vs what we have now with a benchmark?

Yeah, I think that would be helpful

pkg/server/v3/server_test.go
delong-coder: No problem.

pkg/test/resource/v3/resource.go
delong-coder: I see `ptypes` is used everywhere in this file. Shall we change all the `ptypes` to `anypb` at some point in the future?

alecholmez: Sure we can do that

delong-coder:
@alecholmez Is there any changes need to be done here?

alecholmez:
I think this is good to go in, let's make sure the build is all good first, @snowp any objections?

alecholmez:
Oh hmm @delong-coder seems like the build is failing can you take a look?

delong-coder:
> Oh hmm @delong-coder seems like the build is failing can you take a look?

Sure, updated the delta_test.go, please help run the build again, thanks.

delong-coder:
> > Oh hmm @delong-coder seems like the build is failing can you take a look?
> 
> Sure, updated the delta_test.go, please help run the build again, thanks.

@alecholmez

alecholmez:
Cool I think this is good to merge. I won't hold up this logic on a benchmark since that is going to take some time. Thanks @delong-coder

alecholmez (approved):

delong-coder:
@kyessenov Can I ask what time do we plan to push the new release? Thanks.

kyessenov:
We don't have a time-based policy. If there's demand then it's a matter of compiling the release notes and making sure things are not cut in-the-middle of some work in progress.

PR #404, opened 2021-03-16
xds: Update udpa -> xds repo

phlax:
test/fix https://github.com/envoyproxy/envoy/issues/15511


phlax:
on further investigation its the `make bin/example` target that is failing in envoy ci

i tried adding just a test for this target but it passes ci here - so not sure how to fix ci

i did also test this PR (with the udpated paths) and can confirm it fixes the envoy issue

.circleci/config.yml
```diff
@@ -9,6 +9,7 @@ jobs:
       - run: make create_version
       - run: make check_version_dirty
       - run: make build
+      - run: make bin/example
```
phlax, line 12: this didnt fail even without the fix not sure why - i can happily remove

phlax: ...although this is what is failing in envoy ci

htuch (approved):

PR #406, opened 2021-03-30
Incremental xDS Cache Implementation

alecholmez:

This is everything needed for a working delta xDS cache implementation. This comes from the original [Incremental PR](https://github.com/envoyproxy/go-control-plane/pull/387) but is one of 3 parts of the new logical breakdown to help review.

@snowp I can also breakdown the specific cache implementations if that helps? You're the reviewer here so let me know what makes your life easier. This is still a bit large but cut it in half since the v2/v3 code dupe bloats the numbers. 

This PR also contains unit tests for all the cache implementations to demonstrate adequate functionality. The later PRs in this breakdown will include more tests and will all add up to the integration test at the end to show the full feature implementation of the delta xDS protocol.


alecholmez:
Not entirely sure why the CI build isn't kicking off either, I didn't modify any of that yet.

snowp:
Yeah I think splitting this up by cache type would be helpful. I tend to aim for PRs that are under 1k LOC if possible

I also think that if we land the cache interface changes it would be easier to concurrently review a consuming PR even if we haven't implemented delta for each of the existing caches, so that might speed up the overall time to land all of this.

alecholmez:
@snowp that sounds good. I was looking at this and kind of realizing a breakdown of the logical cache types would be better for review. 

Ah I see what you're saying. Land the interface then we can have various reviews for the different types of caches. Cool that sounds good I'll see what I can do for splitting that up

Issue #407, opened 2021-03-31
Open-telemetry proto import problem


biefy:
Data plane API now has a reference to open-telemetry protos. However, go-control-plane API does not reference these new protos when generating go code.  Import of `logs` package is not correctly recognized.  See https://github.com/envoyproxy/go-control-plane/blob/main/envoy/extensions/access_loggers/open_telemetry/v3alpha/logs_service.pb.go#L16

```
package envoy_extensions_access_loggers_open_telemetry_v3alpha

import (
	_ "github.com/cncf/xds/go/udpa/annotations"
	v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/access_loggers/grpc/v3"
	_ "github.com/envoyproxy/protoc-gen-validate/validate"
	proto "github.com/golang/protobuf/proto"
	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
	protoimpl "google.golang.org/protobuf/runtime/protoimpl"
	logs "logs"
	reflect "reflect"
	sync "sync"
)
```
This needs to be fixed.  Otherwise, control planes using go-control-plane have no choice but to ignore this package.

snowp:
Should be resolved now, can you verify?

biefy:
The problem is gone.  Thanks @snowp for the fix.

PR #408, opened 2021-03-31
Cache and Server Incremental Interfaces

alecholmez:
@snowp this is everything for a stable non-intrusive build. This includes all interfaces needed for incremental in both the server and cache implementations but nothing is implemented. Let me know if this is okay


pkg/cache/v2/cache.go
```diff
@@ -145,30 +210,113 @@ func (r *RawResponse) GetDiscoveryResponse() (*discovery.DiscoveryResponse, erro
...
+
+// GetSystemVersion returns the response version.
+func (r *DeltaPassthroughResponse) GetSystemVersion() (string, error) {
+	if r.DeltaDiscoveryResponse != nil {
+		return r.DeltaDiscoveryResponse.SystemVersionInfo, nil
+	}
+	return "", fmt.Errorf("DeltaDiscoveryResponse is nil")
+}
+
+// GetDeltaVersionMap ...
```
dontlaugh, line 319: Fixup this doc string?

alecholmez:
Looks like there is an issue with how the protos were generated? They have a malformed input

pkg/cache/v2/status.go
```diff
@@ -77,11 +90,24 @@ type ResponseWatch struct {
 	Response chan Response
 }
 
+// DeltaResponseWatch is a watch record keeping both the delta request and an open channel for the delta response.
+type DeltaResponseWatch struct {
+	// Request is the original delta request for the watch
```
snowp, line 95: What does original mean in the context of delta? First of the stream? Or just the latest one since we recreated watches for each request?

pkg/server/delta/v2/server.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"context"
+
+	discovery "github.com/envoyproxy/go-control-plane/envoy/api/v2"
+	"github.com/envoyproxy/go-control-plane/pkg/cache/v2"
+	"github.com/envoyproxy/go-control-plane/pkg/server/stream/v2"
+)
+
+// Server is defined to implement the specific stream handler type
```
snowp, line 11: Hard to parse what this is trying to say, can you clarify?

pkg/server/v2/server.go
```diff
@@ -84,6 +92,22 @@ func (c CallbackFuncs) OnStreamClosed(streamID int64) {
 	}
 }
 
+// OnDeltaStreamOpen invokes StreamOpenFunc.
```
snowp, line 95: These comments aren't updated to say that it invokes the delta variant

In general these comments feel pretty redundant, I wouldn't mind getting rid of them (unless we have some godoc integration that really wants these)

pkg/server/delta/v2/server.go
```diff
@@ -0,0 +1,49 @@
...
+type server struct {
+	cache     cache.ConfigWatcher
+	callbacks Callbacks
+
+	// streamCount for counting bi-di streams
+	streamCount int64
+	ctx         context.Context
+}
+
+// NewServer creates handlers from a config watcher and callbacks.
```
snowp, line 38: it creates a server right? not handlers

pkg/cache/v2/cache.go
```diff
@@ -145,30 +210,113 @@ func (r *RawResponse) GetDiscoveryResponse() (*discovery.DiscoveryResponse, erro
...
 
+// GetSystemVersion returns the raw SystemVersion
+func (r *RawDeltaResponse) GetSystemVersion() (string, error) {
+	return r.SystemVersionInfo, nil
+}
+
+// GetDeltaVersionMap returns the delta version map built internally by the cache for the state of a snapshot
+func (r *RawDeltaResponse) GetDeltaVersionMap() map[string]string {
+	return r.VersionMap
+
```
snowp, line 280: extra newline

pkg/cache/v2/cache.go
```diff
@@ -94,7 +124,29 @@ type RawResponse struct {
...
+	// Request is the original request.
+	DeltaRequest *discovery.DeltaDiscoveryRequest
+
+	// System Version Info
+	SystemVersionInfo string
+
+	// Resources to be included in the response.
+	Resources []types.Resource
+
+	// RemovedResources is a list of unsubscribed aliases to be included in the response
```
snowp, line 138: Is this not the list of resources that should be removed on the client? I don't think this affects subscriptions

snowp (changes requested):
Thanks this is much easier to digest. 

Mostly commented on comments, I'm guessing a lot of it is copied from the sotw implementation, so if things I've commented on just matches the sotw code we can sort that out separately in another PR.

pkg/cache/v2/resource.go
```diff
@@ -45,12 +48,29 @@ func GetResponseType(typeURL string) types.ResponseType {
 		return types.Secret
 	case resource.RuntimeType:
 		return types.Runtime
-	case resource.ExtensionConfigType:
-		return types.ExtensionConfig
```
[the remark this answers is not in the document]
alecholmez: Oh oops sorry about that, I think that was accidental.

pkg/cache/v2/resource.go
```diff
@@ -45,12 +48,29 @@ func GetResponseType(typeURL string) types.ResponseType {
...
+	case types.Route:
+		return resource.RouteType
+	case types.Listener:
+		return resource.ListenerType
+	case types.Secret:
+		return resource.SecretType
+	case types.Runtime:
+		return resource.RuntimeType
+	}
+	return resource.AnyType
```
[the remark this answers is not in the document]
alecholmez: Ah thank you, this is old code I can actually delete this.

pkg/cache/v2/resource.go
```diff
@@ -138,3 +158,19 @@ func GetResourceReferences(resources map[string]types.ResourceWithTtl) map[strin
 	}
 	return out
 }
+
+// HashResource will take a resource and create a SHA256 hash sum out of the marshaled bytes
+func HashResource(resource types.Resource) (string, error) {
```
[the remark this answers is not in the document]
alecholmez: Ah yes thank you for that catch

pkg/cache/v2/resource.go
```diff
@@ -138,3 +158,19 @@ func GetResourceReferences(resources map[string]types.ResourceWithTtl) map[strin
...
+	hasher := sha256.New()
+	buffer := proto.NewBuffer(nil)
+	buffer.SetDeterministic(true)
+
+	err := buffer.Marshal(resource)
+	if err != nil {
+		return "", err
+	}
+	hasher.Write(buffer.Bytes())
+	buffer.Reset()
```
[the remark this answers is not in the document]
alecholmez: Removed this code since we can pass the marshaledBytes through instead of recreating the work

pkg/server/v2/server.go
alecholmez: Yup updated... I kept these for consistencies sake. I'm open to heaving them but I wanted the code to look familiar to what as already in SOTW

pkg/server/delta/v2/server.go
alecholmez: Ah sorry, I copied that comment from the SOTW implementation. Will clarify

pkg/server/delta/v2/server.go
```diff
@@ -0,0 +1,49 @@
...
+	OnStreamDeltaRequest(int64, *discovery.DeltaDiscoveryRequest) error
+	// OnStreamDelatResponse is called immediately prior to sending a response on a stream.
+	OnStreamDeltaResponse(int64, *discovery.DeltaDiscoveryRequest, *discovery.DeltaDiscoveryResponse)
+}
+
+type server struct {
+	cache     cache.ConfigWatcher
+	callbacks Callbacks
+
+	// streamCount for counting bi-di streams
```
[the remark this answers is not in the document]
alecholmez: Ah another SOTW copy pasta, this is total. It only gets modified when `process()` is called:
```go
// process handles a bi-di stream request
func (s *server) process(stream Stream, reqCh <-chan *discovery.DiscoveryRequest, defaultTypeURL string) error {
	// increment stream count
	streamID := atomic.AddInt64(&s.streamCount, 1)
```

pkg/cache/v2/cache.go
```diff
@@ -145,30 +210,113 @@ func (r *RawResponse) GetDiscoveryResponse() (*discovery.DiscoveryResponse, erro
...
+			}
+			version, err := HashResource(resource)
+			if err != nil {
+				panic(err)
+			}
+
+			name := GetResourceName(resource)
+			marshaledResources[i] = &discovery.Resource{
+				Name:    name,
+				Aliases: []string{name},
```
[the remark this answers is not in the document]
alecholmez: I went with name here due to the information stated within the discovery proto code:
```go
	// The aliases are a list of other names that this resource can go by.
	Aliases []string `protobuf:"bytes,4,rep,name=aliases,proto3" json:"aliases,omitempty"`
```
It seems like the literal meaning is name, I'm not sure what else we could use here? Unless we derived a name from particular information in the cache?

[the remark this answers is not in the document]
snowp: The key thing to notice from those docs is that it says "other", it will use these in addition to the Name so there is no reason to duplicate it

An example usage in Envoy looks like 

```
  std::set<std::string> added_resource_ids;
  for (const auto& resource : added_resources) {
    added_resource_ids.emplace(resource.get().name());
    std::copy(resource.get().aliases().begin(), resource.get().aliases().end(),
              std::inserter(added_resource_ids, added_resource_ids.end()));
...
```

which would just dedup it away, so there's no harm behind perhaps some confusion. I think we should just leave this empty.

cc @htuch in case we feel like the docs could be clearer here, maybe not clear that this is completely optional?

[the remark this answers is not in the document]
alecholmez: Gotcha. Yup we can probably pull that out it seems like it's not needed if that is the case. I misunderstood that this was an additional list of potential names + the original name. I'll make the change!

pkg/cache/v2/cache.go
```diff
@@ -72,6 +87,21 @@ type Response interface {
 	GetVersion() (string, error)
 }
 
+// DeltaResponse is a wrapper around Envoy's DeltaDiscoveryResponse
+type DeltaResponse interface {
+	// Get the constructed DeltaDiscoveryResponse
+	GetDeltaDiscoveryResponse() (*discovery.DeltaDiscoveryResponse, error)
+
+	// Get te original Request for the Response.
+	GetDeltaRequest() *discovery.DeltaDiscoveryRequest
```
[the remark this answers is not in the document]
alecholmez: Right, this should be treated as the most recent request. As far as my understanding of the protocol goes, the communication mechanism between the management server and the xDS client is still the same, but the management server is in charge of the delta's and spontaneously sending out the correct cache information based upon those deltas. And then of course the ability for clients to lazy load/request, etc...

But the reason this is here is because we still need the ability to track the the ACK/NACK i.e. the `response_nonce` and any error details that might come through. So this does become the most recent request on the stream

pkg/cache/v2/cache.go
```diff
@@ -72,6 +87,21 @@ type Response interface {
...
+
+	// Get the version in the Response.
```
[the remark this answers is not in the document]
alecholmez: Sure thing!

pkg/cache/v2/cache.go
```diff
@@ -105,7 +157,20 @@ type PassthroughResponse struct {
 	DiscoveryResponse *discovery.DiscoveryResponse
 }
 
+// DeltaPassthroughResponse is a pre constructed xDS response that need not go through marshalling transformations.
+type DeltaPassthroughResponse struct {
+	// Request is the original request
```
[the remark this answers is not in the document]
alecholmez: 👍 responded above

pkg/cache/v2/status.go
alecholmez: Responded above to this, I'll change the comment!

pkg/cache/v2/cache.go
alecholmez: Yes sorry that comment is misleading. Will change

pkg/cache/v2/cache.go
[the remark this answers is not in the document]
snowp: Yeah I think that's fine, just make sure we capture that in the comments etc. The protocol difference is that SotW there is a sense that "this is request that requested this response", but for delta the "requested state" is the sum of all the requests on the stream. Because of that the term "original" seems more ambiguous because it could also refer to the first request of the stream

pkg/cache/v2/cache.go
```diff
@@ -72,6 +87,21 @@ type Response interface {
...
+	GetDeltaDiscoveryResponse() (*discovery.DeltaDiscoveryResponse, error)
+
+	// Get te original Request for the Response.
+	GetDeltaRequest() *discovery.DeltaDiscoveryRequest
+
+	// Get the version in the Response.
+	GetSystemVersion() (string, error)
+
+	// Get the version map of the internal cache
+	GetDeltaVersionMap() map[string]string
```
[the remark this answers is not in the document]
alecholmez: This is here because we bubble up this generated version map into the server for it to use as a state tracker. For STOW it's similar to how `VersionInfo` was used, so `GetVersion()` equates to `GetVersionMap()`. They're used differently between the protocols but we still needed a way to pull version information out of the response in the server code since Incremental didn't have a mechanism like that (with the exception of the `system_version_info` debugging field but that is not a reliable source of version information and doesn't account for per resource version changes)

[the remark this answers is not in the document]
snowp: Yeah makes sense, especially after I looked back at the other PR how this is used.

How about we try to name it something that makes it a bit clearer that it's essentially the updated version mappings after this response is applied? 

As I understand it for a given delta stream we have version map at time T that we pass to CreateWatch, and it returns the version map at T+1, containing the resource versions after we apply the response. So maybe something like `NextVersionMap`, `UpdatedVersionMap`, etc?

I think the key difference between this and the SotW equivalent is that for SotW the VersionInfo doesn't depend on the input and captures the full version info of the response, which makes it a bit easier to grok.

pkg/cache/v2/cache.go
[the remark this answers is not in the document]
alecholmez: Yup understood, I'll make sure in the comments that is noted.

pkg/cache/v2/cache.go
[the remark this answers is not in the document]
alecholmez: Ah yes I like this. I think semantically `NextVersionMap` is my favorite. 

Right that was my understanding as well 👍

alecholmez:
I think most of this is addressed @snowp should anyone else look at this? Also I'm not entirely sure how to fix the build, I tried rebasing from master but it looks like that code is still broken

pkg/cache/v2/cache.go
```diff
@@ -72,6 +87,23 @@ type Response interface {
 	GetVersion() (string, error)
 }
 
+// DeltaResponse is a wrapper around Envoy's DeltaDiscoveryResponse
+type DeltaResponse interface {
+	// Get the constructed DeltaDiscoveryResponse
+	GetDeltaDiscoveryResponse() (*discovery.DeltaDiscoveryResponse, error)
+
+	// Get the latest DeltaRequest for the DeltaResponse.
+	// Delta's requesting mechanism
```
snowp, line 96: incomplete comment?

pkg/cache/v2/cache.go
```diff
@@ -94,7 +126,29 @@ type RawResponse struct {
 	marshaledResponse atomic.Value
 }
 
+// RawDeltaResponse is a pre-serialized xDS response that utilizes the delta discovery request/response objects.
+type RawDeltaResponse struct {
+	// Request is the latest delta request on the stream.
+	DeltaRequest *discovery.DeltaDiscoveryRequest
+
+	// SystemVersionInfo is present for holding the current applied response version but generally used debugging purposes only.
```
snowp, line 134: This sentence reads a bit weird? `SystemVersion holds the currently applied response system version and should be used for debugging purposes only`

pkg/cache/v2/cache.go
```diff
@@ -94,7 +126,29 @@ type RawResponse struct {
...
+	SystemVersionInfo string
+
+	// Resources to be included in the response.
+	Resources []types.Resource
+
+	// RemovedResources is a list of resource aliases which should be dropped by the client consuming.
```
snowp, line 140: `consuming client`

pkg/cache/v2/cache.go
```diff
@@ -94,7 +126,29 @@ type RawResponse struct {
...
+	RemovedResources []string
+
+	// VersionMap is a map of versions applied internally to the cache grouped by resource name.
+	VersionMap map[string]string
```
snowp, line 144: Same rename here to NextVersionMap?

pkg/cache/v2/cache.go
```diff
@@ -105,7 +159,20 @@ type PassthroughResponse struct {
 	DiscoveryResponse *discovery.DiscoveryResponse
 }
 
+// DeltaPassthroughResponse is a pre constructed xDS response that need not go through marshalling transformations.
+type DeltaPassthroughResponse struct {
+	// Request is the latest delta request on the stream
+	DeltaRequest *discovery.DeltaDiscoveryRequest
+
+	// VersionMap is a map of versions applied internally to the cache
+	VersionMap map[string]string
```
snowp, line 168: same + update comment

pkg/cache/v2/cache.go
```diff
@@ -145,30 +212,111 @@ func (r *RawResponse) GetDiscoveryResponse() (*discovery.DiscoveryResponse, erro
...
 func (r *RawResponse) GetVersion() (string, error) {
 	return r.Version, nil
 }
 
+// GetSystemVersion returns the raw SystemVersion
+func (r *RawDeltaResponse) GetSystemVersion() (string, error) {
+	return r.SystemVersionInfo, nil
+}
+
+// NextVersionMap returns the delta version map built internally by the cache for the state of a snapshot
```
snowp, line 278: update to talk about "next version". This is the downside of all these repeated comments :)

pkg/cache/v2/resource.go
```diff
@@ -138,3 +141,11 @@ func GetResourceReferences(resources map[string]types.ResourceWithTtl) map[strin
 	}
 	return out
 }
+
+// HashResource will take a resource and create a SHA256 hash sum out of the marshaled bytes
+func HashResource(resource []byte) (string, error) {
```
snowp, line 146: This never returns an error, so just return string?

snowp (changes requested):

snowp:
Seems like you recently included some proto changes? Would be good to keep those out. 

I think the build is broken on main due to https://github.com/envoyproxy/go-control-plane/issues/407, we probably need to fix that in order to proceed here

pkg/cache/v2/cache.go
alecholmez: I think if we call this `NextVersionMap` we'll get a naming collision, so I'd have to change the method to maybe `GetNextVersionMap()`?

pkg/cache/v2/cache.go
alecholmez: Ah oops

alecholmez: Does this make sense?
```go
	// Delta's requesting mechanism requires this to be the latest and not the original request
	// due to the protocol's knowledgement requirement of all past requests so the
	// management server can accurately create deltas on current resource maps
```

pkg/cache/v2/cache.go
alecholmez: Will do, I know sorry about that. I tried to comment as much as possible but the repetitive nature is no good

pkg/cache/v2/cache.go
snowp: I think whichever way we go it would be good be consistent with the naming

pkg/cache/v2/cache.go
snowp: How about

```
// The request that created the watch that we're now responding to. This is provided to allow the caller to correlate the
// response with a request. Generally this will be the latest request seen on the stream for the specific type.
```

I don't think we need to speak about the notion of original vs most recent, just speak to what it actually is and why it's provided

alecholmez:
@snowp I'm not sure I see those proto changes? Maybe my git fu is lacking

pkg/cache/v2/cache.go
alecholmez: Excellent I like that

pkg/cache/v2/cache.go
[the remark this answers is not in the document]
htuch: @dmitri-d for docs update as this is VHDS specific. We plan on deprecating aliases in the future, but for now they are essentially there for VHDS to fetch multiple virtual hosts in a single resource on-demand.

snowp:
Thanks, can you merge main to pick up the build fix?

alecholmez:
@snowp Builds all good!

snowp (approved):
Thanks!

alecholmez:
thanks for the reviews! Queuing up the implementations now

PR #409, opened 2021-04-02
ci: update to go 1.16 and fix go mod

mattklein123:
1) Update to go 1.16
2) Fix go mod issues
3) Switch to GH actions and build docker image inline
4) Other small fixes

Signed-off-by: Matt Klein <mklein@lyft.com>


mattklein123:
@snowp @kyessenov this should be ready for review. Once we review and merge this I will remove the circleci requirements from the build. I will follow up later with the TODO fixes.

Makefile
```diff
@@ -12,17 +12,19 @@ PKG 		:= github.com/envoyproxy/go-control-plane
...
 .PHONY: cover
 cover:
 	@build/coverage.sh
 
-.PHONY: format
-format:
-	@goimports -local $(PKG) -w -l pkg
+.PHONY: check_format
+check_format:
+	@gofmt -l $(shell go list -f '{{.Dir}}' ./...) | tee /dev/stderr | read && echo "Files failed gofmt" && exit 1 || true
```
snowp, line 27: Should we still be using goimports? iiuc it does what go fmt does but more

snowp (commented):
Thanks, this looks right to me. Seems like a nice cleanup!

Just one question

Makefile
mattklein123: I see let me check that. It didn't seem like goimports worked in the new container so I just made this what ratelimit was doing.

mattklein123:
@snowp OK I think this should be back to goimports

Makefile
```diff
@@ -12,9 +12,11 @@ PKG 		:= github.com/envoyproxy/go-control-plane
 build:
 	@go build ./pkg/... ./envoy/...
 
+# TODO(mattklein123): See the note in TestLinearConcurrentSetWatch() for why we set -parallel here
+# This should be
```
snowp, line 16: Partial comment

snowp (approved):
Thanks this looks good just the one comment nit. 

Might be worth moving towards using https://golangci-lint.run/ to mange running linters etc., but we can tackle that separately.

mattklein123:
@snowp updated

snowp (approved):
Thanks!

PR #411, opened 2021-04-06
Incremental Simple Cache Implementation

alecholmez:
@snowp here's the incremental implementation for the simple snapshot cache. There are a few functions in here to support the incoming server code, such as `GetVersionMap()` and `GetResponseTypeURL()`. If needed I can pull those out and add them to the server PR but I figured I'd group them with the cache since this logic will probably merge in first.

@snowp I didn't bring back your server linearization logic into this PR but I can create another one which will patch accordingly.

### Tests
```
./build/run_docker make test
```
All unit tests pass for V2 and V3 of incremental.


snowp:
@fxposter FYI

pkg/cache/v2/simple.go
```diff
@@ -366,8 +378,65 @@ func createResponse(request *Request, resources map[string]types.ResourceWithTtl
...
+	info.mu.Lock()
+	info.lastDeltaWatchRequestTime = time.Now()
+	info.mu.Unlock()
+
+	value := make(chan DeltaResponse, 1)
+
+	// find the current cache snapshot for the provided node
+	snapshot, exists := cache.snapshots[nodeID]
+
+	// if respondDelta returns nil, there is no change in any nested resource version against the previous snapshot
```
snowp, line 405: Does talking about the previous snapshot make sense in this context? We're just creating a watch against the current snapshot

I think this maybe leaks the fact that the incoming request might be happening in response to a snapshot change, but it requires quite a bit of mental acrobatics to get there.

Maybe just say something like `If respondDelta returns nil, there are no resources to send so we create a watch to wait for future updates`?

pkg/cache/v2/simple.go
```diff
@@ -366,8 +378,65 @@ func createResponse(request *Request, resources map[string]types.ResourceWithTtl
...
+	info, ok := cache.status[nodeID]
+	if !ok {
+		info = newStatusInfo(request.Node)
+		cache.status[nodeID] = info
+	}
+
+	// update last watch request times
+	info.mu.Lock()
+	info.lastDeltaWatchRequestTime = time.Now()
+	info.mu.Unlock()
```
snowp, line 398: it might be nice to have these simple getters use a helper function, e..g

```
func (i *info) SetLastDeltaWatchRequestTime(t time.Time) {
  i.Lock()
  defer i.Unlock()
  return i.lastDeltaRequestTime = t
}
```

as this avoids leaking the need for a lock to the caller

pkg/cache/v2/simple.go
```diff
@@ -366,8 +378,65 @@ func createResponse(request *Request, resources map[string]types.ResourceWithTtl
...
+}
+
+func (cache *snapshotCache) nextDeltaWatchID() int64 {
+	return atomic.AddInt64(&cache.deltaWatchCount, 1)
+}
+
+// cancellation function for cleaning stale delta watches
+func (cache *snapshotCache) cancelDeltaWatch(nodeID string, watchID int64) func() {
+	return func() {
+		// uses the cache mutex
```
snowp, line 431: why? clear from the code that we lock the cache mutex, so either elaborate or remove

pkg/cache/v2/snapshot.go
```diff
@@ -31,6 +31,21 @@ type Resources struct {
...
+	SystemVersion string
+
+	// Items in the group indexed by name
+	Items resourceItems
+}
+
+// resourceItems contain the lower level versioned resource map
+type resourceItems struct {
+	Version string
+	Items   map[string]types.Resource
```
snowp, line 46: Maybe use ResourceWithTTL here? I don't think we need to support TTL with Delta in this PR, but might as well use the right types if its not too much work

pkg/cache/v2/snapshot.go
```diff
@@ -40,6 +55,15 @@ func IndexResourcesByName(items []types.ResourceWithTtl) map[string]types.Resour
 	return indexed
 }
 
+// IndexRawResourcesByName creates a map from the resource name to the resource.
+func IndexRawResourcesByName(items []types.Resource) map[string]types.Resource {
```
snowp, line 59: probably won't need this if you switch to ResourceWithTTL

pkg/cache/v2/snapshot.go
```diff
@@ -194,3 +218,44 @@ func (s *Snapshot) GetVersion(typeURL string) string {
 	}
 	return s.Resources[typ].Version
 }
+
+// GetVersionMap will construct a verison map off the current state of a snapshot
```
snowp, line 222: off -> based of/on?

pkg/cache/v2/snapshot.go
```diff
@@ -194,3 +218,44 @@ func (s *Snapshot) GetVersion(typeURL string) string {
...
+	return versionMap, nil
+}
+
+// initializeVMap will build a nested map structure to hold all our version information
+func initializeVMap() map[string]map[string]string {
+	versionMap := make(map[string]map[string]string, types.UnknownType)
+
+	for i := 0; i < int(types.UnknownType); i++ {
+		versionMap[GetResponseTypeURL(types.ResponseType(i))] = make(map[string]string)
+	}
```
snowp, line 258: Is this going to work with custom resource types? Answer can be no, we can do in a follow up unless there are major blockers

snowp (changes requested):
Thanks for working on this! Got you started with some comments

pkg/cache/v2/resource.go
```diff
@@ -54,6 +54,25 @@ func GetResponseType(typeURL string) types.ResponseType {
...
+	case types.Route:
+		return resource.RouteType
+	case types.Listener:
+		return resource.ListenerType
+	case types.Secret:
+		return resource.SecretType
+	case types.Runtime:
+		return resource.RuntimeType
+	}
+	return resource.AnyType
```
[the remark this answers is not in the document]
alecholmez: Ah yes, so I didn't know what else to default to since there isn't really a direct mapping to `types.UnknownType`. Would it make sense to just return an empty string here? Callers would just have to check to make sure the conversion was successful

pkg/cache/v2/simple.go
```diff
@@ -72,6 +72,8 @@ type snapshotCache struct {
 	// 32-bit machines.
 	watchCount int64
 
+	deltaWatchCount int64
```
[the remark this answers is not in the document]
alecholmez: I don't think it'll have alignment issues. I'll add a comment

pkg/cache/v2/delta.go
```diff
@@ -0,0 +1,112 @@
...
+				request.GetNode().GetId(), st.ResourceVersions, resp.Resources, resp.NextVersionMap, resp.RemovedResources, st.IsWildcard)
+		}
+		value <- resp
+		return resp
+	}
+	return nil
+}
+
+func createDeltaResponse(request *DeltaRequest, st *stream.StreamState, resources map[string]types.Resource, systemVersion string) (*RawDeltaResponse, error) {
+	nextVersionMap := make(map[string]string)
```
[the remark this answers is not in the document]
alecholmez: I don't think that should cause any issues. It might cause empty entries if the new version map shrinks but we check for that so should be good 👍

pkg/cache/v2/snapshot.go
```diff
@@ -194,3 +218,44 @@ func (s *Snapshot) GetVersion(typeURL string) string {
 	}
 	return s.Resources[typ].Version
 }
+
+// GetVersionMap will construct a verison map off the current state of a snapshot
+func (s *Snapshot) GetVersionMap() (map[string]map[string]string, error) {
```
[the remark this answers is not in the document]
alecholmez: `GetVersionMap()` is only called in tests as the current code stands. That's an good suggestion though, but it might not be worth moving that computation into the snapshot creation if it doesn't get utilized anywhere else

pkg/cache/v2/simple.go
```diff
@@ -366,8 +378,65 @@ func createResponse(request *Request, resources map[string]types.ResourceWithTtl
 	}
 }
 
+// CreateDeltaWatch returns a watch for a delta xDS request which implements the Simple SnapshotCache.
```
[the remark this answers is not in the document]
alecholmez: Ah yea, i'll take off the part `which implements the Simple SnapshotCache`. Then it will read like the SOTW comment. Unless it's worth while differentiating the two here since the request cycle is a little different in delta

pkg/cache/v2/simple.go
alecholmez: Yea I see what you mean. I think it is a little hard to follow. If they want to investigate what respondDelta actually does they can go dig and should be able to figure out it just compares the resource version hashes

pkg/cache/v2/simple.go
alecholmez: Yup, I'll change that. I just noticed the code dupe

pkg/cache/v2/simple.go
alecholmez: Removed, comment wasn't helpful

pkg/cache/v2/snapshot.go
alecholmez: 👍 I'll move to the `ResourceWithTTL` here. When I wrote this code that wasn't in yet and I never updated so I'll change it now while we're working on it

pkg/cache/v2/snapshot.go
alecholmez: Cool I'll pull it out

pkg/cache/v2/delta.go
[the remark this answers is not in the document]
snowp: I was thinking about reserving capacity, not initializing to empty. Maybe maps don't have a reserve concept? If so ignore this suggestion

pkg/cache/v2/snapshot.go
alecholmez: I don't believe so no. Right now we only build the version map off the pre-defined resource types in the resource package so if a consumer were to use custom resource types I don't think it would work as of right now

pkg/cache/v2/delta.go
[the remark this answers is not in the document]
alecholmez: Ah, in go I believe there is no ability to allocate capacity in a map when using make:

make(T, n) for type map: "initial room for approximately n elements"

So i think if we create a size here it just allocates to that size and holds with empty resources

pkg/cache/v2/delta.go
```diff
@@ -44,17 +45,22 @@ func respondDelta(request *DeltaRequest, value chan DeltaResponse, st *stream.St
...
+	filtered := make([]types.ResourceWithTtl, 0)
 	toRemove := make([]string, 0)
 
 	// Wildcard can happen through CDS/LDS. If this is done we want to track all resources as well as track them in the version map
 	if st.IsWildcard {
-		for name, resource := range resources {
-			marshaledResource, err := MarshalResource(resource)
+		for name, r := range resources {
+			// Set the heartbeat here to false since I'm not entirely sure where to pull that from here?
+			maybeTtldResource, _, err := ttl.MaybeCreateTtlResourceIfSupported(r, name, request.TypeUrl, false)
```
snowp, line 57: For heartbeating to work with Delta we need to be pushing delta updates here: https://github.com/envoyproxy/go-control-plane/blob/35430f0740d5573e1ef9bf38599b2eb4d4e71612/pkg/cache/v2/simple.go#L156

As mentioned earlier we can handle TTL + heartbeating in delta separately and make sure we add proper test coverage if you feel like that would be easier, no need to block the immediate delta work. 

If we are going to add this we should have test coverage for TTL and heartbeats for delta APIs

For all other responses not triggered by the heartbeat routine we should set false here

pkg/cache/v2/simple.go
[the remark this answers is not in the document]
snowp: Maybe just combine the two comments into one? Right now its very verbose

pkg/cache/v2/simple.go
```diff
@@ -215,10 +218,14 @@ func (cache *snapshotCache) SetSnapshot(node string, snapshot Snapshot) error {
...
-				snapshot.GetVersion(watch.Request.TypeUrl), cache.log) != nil {
-
+			if respondDelta(
+				watch.Request,
+				watch.Response,
+				watch.StreamState,
+				snapshot.GetResourcesAndTtl(watch.Request.TypeUrl),
+				snapshot.GetVersion(watch.Request.TypeUrl),
+				cache.log,
+			) != nil {
```
snowp, line 228: slight preference to use a variable here, formatting makes this look funky

pkg/cache/v2/status.go
```diff
@@ -142,9 +151,28 @@ func (info *statusInfo) GetLastDeltaWatchRequestTime() time.Time {
...
+	info.lastDeltaWatchRequestTime = t
+}
+
 func (info *statusInfo) GetDeltaStreamState(watchID int64) *stream.StreamState {
 	info.mu.RLock()
 	defer info.mu.RUnlock()
 	return info.deltaWatches[watchID].StreamState
 }
+
+func (info *statusInfo) SetDeltaStreamState(watchID int64, st *stream.StreamState) bool {
```
snowp, line 166: Maybe I'm missing something but where is this used?

pkg/cache/v2/snapshot.go
snowp: Sure, let's file an issue to track?

pkg/cache/v2/resource.go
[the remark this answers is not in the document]
snowp: How about returning (string, error)? The behavior if there are multiple unknown types is unclear to me, maybe consider it an error for now?

pkg/cache/v2/snapshot.go
```diff
@@ -194,3 +209,44 @@ func (s *Snapshot) GetVersion(typeURL string) string {
...
+	if s == nil {
+		return nil, fmt.Errorf("missing snapshot")
+	}
+
+	// We need to initialize the map before we process anything since it is nested
+	versionMap := initializeVMap()
+	for i := 0; i < int(types.UnknownType); i++ {
+		typeURL := GetResponseTypeURL(types.ResponseType(i))
+		resources := s.GetResources(typeURL)
```
snowp, line 223: Can we just iterate over the keys in the snapshot items instead? Seems like this is a bit of a roundabout way of getting to the resource names

snowp, line 223: calling GetResourcesAndTtl is a bit cheaper since it avoids rebuilding the return list

pkg/cache/v2/simple.go
```diff
@@ -366,8 +385,61 @@ func createResponse(request *Request, resources map[string]types.ResourceWithTtl
...
+	if !exists || respondDelta(request, value, st, snapshot.GetResourcesAndTtl(t), snapshot.GetVersion(t), cache.log) == nil {
+		watchID := cache.nextDeltaWatchID()
+		if cache.log != nil {
+			cache.log.Infof("open delta watch ID:%d for %s Resources:%v from nodeID: %q, system version %q", watchID,
+				t, st.ResourceVersions, nodeID, snapshot.GetVersion(t))
+		}
+
+		info.mu.Lock()
+		info.deltaWatches[watchID] = DeltaResponseWatch{Request: request, Response: value, StreamState: st}
+		info.mu.Unlock()
```
snowp, line 420: helper function on info?

snowp (changes requested):

snowp:
FYI force pushing makes it harder to review iterative changes to a PR, so it would be great if you could avoid rebasing and force pushing :)

alecholmez:
oops sorry about that! I wanted to pull in the changes from the v2 removal, i'll make sure I don't do that anymore!

pkg/cache/v2/resource.go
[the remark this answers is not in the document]
alecholmez: Now that I'm thinking, this function might just be invalid when we go to support custom types. I think like you say later on, just going off the known types in the snapshot is the way to go instead of trying to do conversion here.

pkg/cache/v2/snapshot.go
alecholmez: Well wouldn't we still need to calculate the hash here? Which would require us to end up marshaling the resource anyways so we can rebuild the version map

alecholmez: Yup, just changed, good idea thank you for the catch

pkg/cache/v2/delta.go
alecholmez: Awesome, I have that written now just looking at the tests. I implemented the test for delta but it's getting thousands of responses which is obviously not write so something is messed up there

alecholmez:
Just pushed another change, sorry again for the force push. I'm getting TTL's working as quickly as possible, I'm noticing though the cancel functions seem to be `nil` which isn't right so it's not deleting old watches. Looking into that now

pkg/cache/v2/snapshot.go
alecholmez: Think I came up with something clever so we don't need to get the resources at all

snowp:
@alecholmez Did you push the failing tests? I can take a look at the code if you can share them

alecholmez:
@snowp yup I did, it’s in ‘delta_test.go’, you just need to remove the skip from the top. And if you remove the nil check from the cancel calls in the normal snapshot cache test, that also fails from a panic

snowp:
@alecholmez The problem is related to how we recompute the version: when producing a heartbeat resource the code hashes the heartbeat resource instead of the original resource, producing a different version than the original resource version:

```
			name := GetResourceName(resource.Resource)
			maybeTtldResource, resourceType, err := ttl.MaybeCreateTtlResourceIfSupported(resource, name, r.DeltaRequest.TypeUrl, r.Heartbeat)
			if err != nil {
				return nil, err
			}
			marshaledResource, err := MarshalResource(maybeTtldResource)
			if err != nil {
				return nil, err
			}
			version := HashResource(marshaledResource)
			if version == "" {
				return nil, errors.New("failed to create a resource hash")
			}
```

This can be solved locally by explicitly taking the hash of the original resource, but I think this ties into the idea of wanting to persist a version string alongside a resource to avoid repeated hashing, so I'll leave it up to you how to solve this.

snowp:
Oh and the tests don't maintain a version map either, which needs to be done, e.g. by doing 

```
@@ -71,7 +74,7 @@ func TestSnapshotCacheWithDeltaTtl(t *testing.T) {
        updatesByType := map[string]int{}
        for _, typ := range testTypes {
                wg.Add(1)
-               go func(typ string) {
+               go func(typ string, resourceVersions map[string]string) {
                        defer wg.Done()

                        end := time.After(5 * time.Second)
@@ -82,7 +85,7 @@ func TestSnapshotCacheWithDeltaTtl(t *testing.T) {
                                        },
                                        TypeUrl:                typ,
                                        ResourceNamesSubscribe: names[typ],
-                               }, &stream.StreamState{IsWildcard: true, ResourceVersions: nil})
+                               }, &stream.StreamState{IsWildcard: true, ResourceVersions: resourceVersions})

                                select {
                                case out := <-value:
@@ -94,17 +97,17 @@ func TestSnapshotCacheWithDeltaTtl(t *testing.T) {
                                                t.Errorf("get resources %v, want %v", out.(*cache.RawDeltaResponse).Resources, snapshotWithTtl.GetResources(typ))
                                        }

+                                       resourceVersions = out.GetNextVersionMap()
+
                                        mu.Lock()
                                        updatesByType[typ]++
                                        mu.Unlock()
                                case <-end:
-                                       if cancel != nil {
-                                               cancel()
-                                       }
+                                       cancel()
                                        return
                                }
                        }
-               }(typ)
+               }(typ, allResourceVersions[typ])
        }
```

alecholmez:
> @alecholmez The problem is related to how we recompute the version: when producing a heartbeat resource the code > hashes the heartbeat resource instead of the original resource, producing a different version than the original resource > version:

@snowp sweet thank you, I'll get to work on that right away. I didn't think about the version hashes differing from one another due to the new TTL logic. 

>This can be solved locally by explicitly taking the hash of the original resource, but I think this ties into the idea of wanting 
> to persist a version string alongside a resource to avoid repeated hashing, so I'll leave it up to you how to solve this.

I'll think about a good way we can persist a version string alongside a resource, I might run a few ideas by you if that's okay?

> Oh and the tests don't maintain a version map either, which needs to be done, e.g. by doing

Ah yup gotcha thank you, I need to do that.

alecholmez:
@snowp just pushed another commit, I added the version map tracking into the test, and just quickly patched the hash logic to create a version on the original resource. Haven't thought about moving that version creation logic farther up the chain yet. Will think about that now.

The test is still failing unfortunately, it seems like the TTLs aren't causing updates. Whenever you have some spare time can you take a look? I'm not sure I understand the TTL logic enough to figure that out but I'll keep digging in the meantime

snowp:
I would suggest reverting the TTL handling code and try to land this without TTLs at this point, I don't think it's worth stalling the Delta work on getting TTLs to work. 

I'll probably spend some time looking at your code, but I think we can proceed at a better pace here if skip TTL for now since nobody's really using Delta + TTL with go-control-plane.

alecholmez:
@snowp that sounds good, I'll file an issue and work the TTL stuff after this goes in. I'd rather get it in sooner than later so it doesn't sit and go stale.

alecholmez:
@snowp just pushed some new code. I addressed the issue of the versions being rehashed multiple times for comparison, now it's only calculated once when a user sets the snapshot. I also reverted the TTL logic so should be back to where we were. If you have any questions let me know!

pkg/cache/v3/delta.go
```diff
@@ -0,0 +1,103 @@
...
+				request.GetNode().GetId(), resp.Resources, resp.RemovedResources, st.IsWildcard)
+		}
+		value <- resp
+		return resp
+	}
+	return nil
+}
+
+// snapshot.GetResources(watch.Request.TypeUrl),
+// snapshot.GetVersion(watch.Request.TypeUrl),
```
snowp, line 48: Remove?

pkg/cache/v3/delta.go
```diff
@@ -0,0 +1,103 @@
...
+
+func createDeltaResponse(request *DeltaRequest, st *stream.StreamState, snapshot Snapshot) (*RawDeltaResponse, error) {
+	resources := snapshot.GetResources((request.TypeUrl))
+
+	// variables to build our response with
+	nextVersionMap := make(map[string]string)
+	filtered := make([]types.Resource, 0, len(resources))
+	toRemove := make([]string, 0)
+
+	// Wildcard can happen through CDS/LDS. If this is done we want to track all resources as well as track them in the version map
```
snowp, line 58: This sentence is a bit hard to parse, can you rephrase?

pkg/cache/v3/delta.go
```diff
@@ -0,0 +1,103 @@
...
+	if st.IsWildcard {
+		for name, r := range resources {
+			// Since we've already precomputed the version hashes of the new snapshot,
+			// we can just set it here to be used for comparison later
+			version := snapshot.GetVersionMap()[request.TypeUrl][name]
+			fmt.Println("new version " + version)
```
snowp, line 64: remove

pkg/cache/v3/delta.go
```diff
@@ -0,0 +1,103 @@
...
+		// on separate streams since requests do not share their response states.
+		for name, prevVersion := range st.ResourceVersions {
+			if r, ok := resources[name]; ok {
+				nextVersion := snapshot.GetVersionMap()[request.TypeUrl][name]
+				if prevVersion != nextVersion {
+					filtered = append(filtered, r)
+				}
+				nextVersionMap[name] = nextVersion
+			} else {
+				// if prevVersion == "" this means that the resourse was already removed or doesn't yet exist on the client.
```
snowp, line 83: How can this be "" if the resource haven't been sent yet? Wouldn't it not be part of st.ResourceVersions?

pkg/cache/v3/delta.go
```diff
@@ -0,0 +1,103 @@
...
+				if prevVersion != "" {
+					toRemove = append(toRemove, name)
+				}
+
+				// the resource is gone but we keep tracking it in the
+				// version map so we can detect an update if the resource comes back
```
snowp, line 89: Why does this matter? Why not just nil it out?

pkg/cache/v3/delta.go
```diff
@@ -0,0 +1,103 @@
...
+				nextVersionMap[name] = ""
+			}
+		}
+	}
+
+	// send through our version map
```
snowp, line 95: seems redundant, add more context or drop imo

pkg/cache/v3/simple.go
```diff
@@ -210,6 +211,28 @@ func (cache *snapshotCache) SetSnapshot(node string, snapshot Snapshot) error {
...
+		// process our delta watches
+		for id, watch := range info.deltaWatches {
+			res := respondDelta(
+				watch.Request,
+				watch.Response,
+				watch.StreamState,
+				snapshot,
+				cache.log,
+			)
+			if res != nil {
```
snowp, line 232: Maybe add a note why delta needs a nil check here while sotw does not

pkg/cache/v3/simple.go
```diff
@@ -366,8 +389,61 @@ func createResponse(request *Request, resources map[string]types.ResourceWithTtl
...
+	cache.mu.Lock()
+	defer cache.mu.Unlock()
+
+	info, ok := cache.status[nodeID]
+	if !ok {
+		info = newStatusInfo(request.Node)
+		cache.status[nodeID] = info
+	}
+
+	// update last watch request times
```
snowp, line 406: times -> time

pkg/cache/v3/simple.go
```diff
@@ -366,8 +389,61 @@ func createResponse(request *Request, resources map[string]types.ResourceWithTtl
...
+	info.SetLastDeltaWatchRequestTime(time.Now())
+
+	value := make(chan DeltaResponse, 1)
+
+	// find the current cache snapshot for the provided node
+	snapshot, exists := cache.snapshots[nodeID]
+
+	// if respondDelta returns nil, there is no change in any nested resource version against the previous snapshot
```
snowp, line 414: what does nested refer to this in case? Isn't there only one "level" of versions?

pkg/cache/v3/simple.go
```diff
@@ -366,8 +389,61 @@ func createResponse(request *Request, resources map[string]types.ResourceWithTtl
...
+	// if that is the case we want to trigger new watches
```
snowp, line 415: "trigger the watch immediately"?

pkg/cache/v3/snapshot.go
```diff
@@ -103,6 +130,11 @@ func NewSnapshotWithResources(version string, resources SnapshotResources) Snaps
 	out.Resources[types.Runtime] = NewResources(version, resources.Runtimes)
 	out.Resources[types.Secret] = NewResources(version, resources.Secrets)
 	out.Resources[types.ExtensionConfig] = NewResources(version, resources.ExtensionConfigs)
+
+	// Initialize the snapshot version map here
+	v, _ := initializeVMap(&out)
```
snowp, line 135: I would probably just do the initialization of the nested structures during `ConstructVersionMap`, this seems very decoupled from where it gets used, and we end up doing useless work in case the snapshot is never used with delta

snowp, line 135: handle err?

pkg/cache/v3/snapshot.go
```diff
@@ -194,3 +226,52 @@ func (s *Snapshot) GetVersion(typeURL string) string {
 	}
 	return s.Resources[typ].Version
 }
+
+// GetVersionMap will return the internal version map of the currently applied snapshot.
+func (s *Snapshot) GetVersionMap() map[string]map[string]string {
+	return s.VersionMap
+}
+
+// ConstructVersionMap will construct a verison map off the current state of a snapshot
```
snowp, line 235: `baed on`/`based of`

pkg/cache/v3/simple.go
```diff
@@ -210,6 +211,28 @@ func (cache *snapshotCache) SetSnapshot(node string, snapshot Snapshot) error {
 				delete(info.watches, id)
 			}
 		}
+
+		// this should always pass since we initialize on snapshot creation
```
snowp, line 215: Let's remove the check then? Avoids having to debug silent failures

pkg/cache/v3/simple.go
```diff
@@ -210,6 +211,28 @@ func (cache *snapshotCache) SetSnapshot(node string, snapshot Snapshot) error {
...
+		if snapshot.VersionMap != nil {
+			err := snapshot.ConstructVersionMap()
```
snowp, line 217: Maybe do this only if len(deltaWatches) > 0? That way we don't waste time for sotw computing the hashes

pkg/cache/v3/simple.go
```diff
@@ -366,8 +389,61 @@ func createResponse(request *Request, resources map[string]types.ResourceWithTtl
...
+}
+
+func (cache *snapshotCache) nextDeltaWatchID() int64 {
+	return atomic.AddInt64(&cache.deltaWatchCount, 1)
+}
+
+// cancellation function for cleaning stale delta watches
+func (cache *snapshotCache) cancelDeltaWatch(nodeID string, watchID int64) func() {
+	return func() {
+		// uses the cache mutex
```
snowp, line 438: comment seems redundant

pkg/server/stream/v3/stream.go
```diff
@@ -26,7 +26,8 @@ type StreamState struct {
 	// Indicates whether the original DeltaRequest was a wildcard LDS/RDS request.
 	IsWildcard bool
 
-	// ResourceVersions contain a hash of the resource as a version map.
+	// ResourceVersions contain a hash of the resource as the value of the map.
+	// The name is the version key.
```
snowp, line 30: This seems a bit confusing, this should probably speak about resource names somewhere?

snowp (changes requested):

pkg/cache/v3/simple.go
alecholmez: Ah yup sorry I was supposed to pull that out already 😄

pkg/cache/v3/simple.go
alecholmez: Cool I like that

pkg/cache/v3/snapshot.go
alecholmez: Oops sorry, thought I fixed that already.

pkg/cache/v3/cache.go
```diff
@@ -224,7 +225,7 @@ func (r *RawDeltaResponse) GetDeltaDiscoveryResponse() (*discovery.DeltaDiscover
 		marshaledResources := make([]*discovery.Resource, len(r.Resources))
 
 		for i, resource := range r.Resources {
-
+			name := GetResourceName(resource)
```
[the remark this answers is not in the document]
alecholmez: You mean pulling from `ResourceNamesSubscribed()`? I think that would work with the exception of wildcard because then no names would exist in that field

pkg/cache/v3/delta.go
alecholmez: Whoops

pkg/cache/v3/delta.go
alecholmez: Yup sorry, remnants of debugging ttl

pkg/cache/v3/cache.go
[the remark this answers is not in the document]
snowp: I think I got this confused with the type url, this is fine as is. Essentially I was looking for a way to get the resource name without having to know how to handle the specific type (to better support unknown resource types), but since we're tackling that in a follow up I think we can leave this

pkg/cache/v3/delta.go
alecholmez: Well the map is of type `map[string]string` so we can't nil it out or go will throw a type error:
```go
nextVersionMap[name] = nil
```
yields
```
cannot use nil (untyped nil value) as string value in assignmentcompilerIncompatibleAssign
```

But we could do something like:
```go
delete(nextVersionMap, name)
```
and that would nil out that entry. I guess we can just remove that because if the client does ever want that resource again it'll just ask for it correct?

pkg/cache/v3/delta.go
alecholmez: That's a useless comment, dropping

pkg/cache/v3/simple.go
alecholmez: Ah yea good point it's just resource version. Removing that

pkg/cache/v3/snapshot.go
alecholmez: Yup makes sense

pkg/cache/v3/cache.go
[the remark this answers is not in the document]
alecholmez: Cool. Yup I'll revisit this when we add support for custom resource types.

pkg/server/stream/v3/stream.go
alecholmez: Maybe this is a little clearer?
```go
	// ResourceVersions contains a hash of the resource as the value and the resource name as the key.
```

pkg/cache/v3/delta.go
alecholmez: I think you are correct but only for the initial request. Which is why we do a check:
```go
if prevVersion != "" {
	toRemove = append(toRemove, name)
}
```

alecholmez: Actually I think that comment is misleading. If `prevVersion == ""` then the client just isn't aware of it. That doesn't mean it doesn't exist. And if we are now just deleting the entry that's false entirely.

pkg/server/stream/v3/stream.go
snowp: Yep sounds better

alecholmez:
Oops looks like I broke something

pkg/cache/v3/snapshot.go
alecholmez: @snowp so it looks like this is what caused the test to blow up. When I move this into the `ConstructVersionMap` function, it somehow always overwrites and then the watches don't trigger properly. In the test we don't make a new snapshot twice we just overwrite the pre-existing snapshot. I even tried doing something like:
```go
	if len(s.VersionMap) == 0 {
		// initialize ...
		v, err := initializeVMap(s)
		if err != nil {
			return fmt.Errorf("failed to initialize verison map: %v", err)
		}
		s.VersionMap = v
	}
```
in ConstructVersionMap but that didn't work. I'm perplexed by this one

alecholmez:
@snowp so it looks like this is what caused the test to blow up. When I move this into the ConstructVersionMap function, it somehow always overwrites and then the watches don't trigger properly. In the test we don't make a new snapshot twice we just overwrite the pre-existing snapshot. I even tried doing something like:
```go
	if len(s.VersionMap) == 0 {
		// initialize ...
		v, err := initializeVMap(s)
		if err != nil {
			return fmt.Errorf("failed to initialize verison map: %v", err)
		}
		s.VersionMap = v
	}
```
in ConstructVersionMap but that didn't work. I'm perplexed by this one

snowp:
I took a look, you need to initialize the version map in CreateWatch for cases where the snapshot is set before any delta watches are added

I also did some refactoring, here's a commit against this PR that seem to resolve the test failures: https://github.com/envoyproxy/go-control-plane/commit/8997c5bf5ad54e57d645f4dd5f0f63d1135b5476

alecholmez:
Ah @snowp thank you, I guess I wasn't thinking broad enough I had a narrow scope for where to look. Do you mean CreateWatch or CreateDeltaWatch? CreateWatch shouldn't get called in a case where a client is using delta.

Awesome thank you! I'll give that commit a look

pkg/cache/v3/delta.go
```diff
@@ -0,0 +1,93 @@
...
+
+func createDeltaResponse(request *DeltaRequest, st *stream.StreamState, snapshot Snapshot) (*RawDeltaResponse, error) {
+	resources := snapshot.GetResources((request.TypeUrl))
+
+	// variables to build our response with
+	nextVersionMap := make(map[string]string)
+	filtered := make([]types.Resource, 0, len(resources))
+	toRemove := make([]string, 0)
+
+	// Wildcard requests originate from CDS or LDS. If detect one, we want to track all resources as well as track them in the version map
```
snowp, line 53: `If we are handling a wildcard request, we want to respond with all resources`?

I'm not sure what the "as well as track them in the version map" refers to

pkg/cache/v3/delta.go
```diff
@@ -0,0 +1,93 @@
...
+		for name, r := range resources {
+			// Since we've already precomputed the version hashes of the new snapshot,
+			// we can just set it here to be used for comparison later
+			version := snapshot.GetVersionMap()[request.TypeUrl][name]
+			nextVersionMap[name] = version
+			prevVersion, found := st.ResourceVersions[name]
+			if !found || (prevVersion != nextVersionMap[name]) {
+				filtered = append(filtered, r)
+			}
+		}
```
snowp, line 64: Don't we need to compute removed resources here as well?

pkg/cache/v3/delta.go
```diff
@@ -0,0 +1,93 @@
...
+	} else {
+		// Reply only with the requested resources. Envoy may ask each resource
+		// individually in a separate stream. It is ok to reply with the same version
+		// on separate streams since requests do not share their response states.
```
snowp, line 68: I'm not sure if I commented on this already but this comment seems kind of out of place, not sure how it's related so maybe rephrase remove?

pkg/cache/v3/delta.go
```diff
@@ -0,0 +1,93 @@
...
+			if r, ok := resources[name]; ok {
+				nextVersion := snapshot.GetVersionMap()[request.TypeUrl][name]
+				if prevVersion != nextVersion {
+					filtered = append(filtered, r)
+				}
+				nextVersionMap[name] = nextVersion
+			} else {
+				if prevVersion != "" {
+					toRemove = append(toRemove, name)
+				}
```
snowp, line 79: will this ever be true now that we don't leave the deleted entries in there with empty string? maybe we can remove if check and just assume that it should be removed?

pkg/cache/v3/delta_test.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	// set partially-versioned snapshot
+	snapshot2 := snapshot
+	snapshot2.Resources[types.Endpoint] = cache.NewResources(version2, []types.Resource{resource.MakeEndpoint(clusterName, 9090)})
+	if err := c.SetSnapshot(key, snapshot2); err != nil {
+		t.Fatal(err)
+	}
+	if count := c.GetStatusInfo(key).GetNumDeltaWatches(); count != len(testTypes)-1 {
+		t.Errorf("watches should be preserved for all but one: %d", count)
```
snowp, line 73: include some info about what %d is, e.g `number of watches: %d` to make debugging easier

pkg/cache/v3/delta_test.go
```diff
@@ -0,0 +1,148 @@
...
+			TypeUrl:                typ,
+			ResourceNamesSubscribe: names[typ],
+		}, &stream.StreamState{IsWildcard: true, ResourceVersions: nil})
+
+		// Cancel the watch
+		cancel()
+	}
+	// should be status info for the node
+	if keys := c.GetStatusKeys(); len(keys) == 0 {
+		t.Error("got 0, want status info for the node")
```
snowp, line 136: this error message could be more useful, maybe `expected to see a status info registered for watch, saw %d entries`

pkg/cache/v3/delta_test.go
```diff
@@ -0,0 +1,148 @@
...
+				Id: key,
+			},
+			TypeUrl:                typ,
+			ResourceNamesSubscribe: names[typ],
+		}, &stream.StreamState{IsWildcard: true, ResourceVersions: nil})
+
+		// Cancel the watch
+		cancel()
+	}
+	// should be status info for the node
```
snowp, line 134: incomplete sentence, can you rephrase?

pkg/cache/v3/delta_test.go
```diff
@@ -0,0 +1,148 @@
...
+		t.Error("got 0, want status info for the node")
+	}
+
+	for _, typ := range testTypes {
+		if count := c.GetStatusInfo(key).GetNumWatches(); count > 0 {
+			t.Errorf("watches should be released for %s", typ)
+		}
+	}
+
+	if empty := c.GetStatusInfo("missing"); empty != nil {
```
snowp, line 145: `empty` seems a bit odd name, maybe jsut `s` or `statusInfo`?

snowp, line 145: seems a bit odd to test this in this function? seems unrelated to detla?

pkg/cache/v3/simple.go
```diff
@@ -210,6 +211,30 @@ func (cache *snapshotCache) SetSnapshot(node string, snapshot Snapshot) error {
 				delete(info.watches, id)
 			}
 		}
+
+		// We only calculate version hashes when using delta. We don't want to do this when using SOTW.
```
snowp, line 215: note that we avoid this due to computational cost of calculating hashes

pkg/cache/v3/snapshot.go
```diff
@@ -62,6 +86,9 @@ func NewResourcesWithTtl(version string, items []types.ResourceWithTtl) Resource
 // from the snapshot may be delivered to the proxy in arbitrary order.
 type Snapshot struct {
 	Resources [types.UnknownType]Resources
+
+	// VersionMap holds the current hash map of all resources in the snapshot. This is only used in delta xDS
```
snowp, line 90: Maybe note that this should remain nil until it's used, at which point it should be instantiated with <name of func>

snowp (changes requested):

pkg/cache/v3/delta.go
alecholmez: Ah I guess that comment is redundant. I'll remove

pkg/cache/v3/snapshot.go
alecholmez: Sure thing!

alecholmez: Maybe something like this?
```go
	// VersionMap holds the current hash map of all resources in the snapshot.
	// This field should remain nil until it is used, at which point should be
	// instantiated by setting a snapshot while using the delta xDS protocol.
	// VersionMap is only to be used with delta xDS.
```

pkg/cache/v3/simple.go
alecholmez: 👍 Will provide a note there

pkg/cache/v3/delta_test.go
alecholmez: Well we do touch the StatusInfo in places like setting a snapshot and creating delta watches, I did find this to be incorrect though: 
```go
        for _, typ := range testTypes {
		if count := c.GetStatusInfo(key).GetNumWatches(); count > 0 {
			t.Errorf("watches should be released for %s", typ)
		}
	}
```
I went ahead and changed that to `c.GetStatusinfo(key).GetNumDeltaWatches()`

pkg/cache/v3/delta_test.go
alecholmez: Something like this?
```go
// c.GetStatusKeys() should return at least 1 because we register a node ID with the above watch creations
```

pkg/cache/v3/delta_test.go
alecholmez: Yup agreed, will change that.

pkg/cache/v3/delta_test.go
alecholmez: This is a bit verbose but i think provides some good contextual information around the number of watches:
```go
	if count := c.GetStatusInfo(key).GetNumDeltaWatches(); count != len(testTypes)-1 {
		t.Errorf("watches should be preserved for all but one, got: %d open watches instead of the expected %d open watches", count, len(testTypes)-1)
	}
```

pkg/cache/v3/delta.go
alecholmez: Oh hmm that's actually a good point. That entry won't ever exist since we delete so that check would never pass. That check is mute so I'll remove.

pkg/cache/v3/delta.go
alecholmez: Agreed, sorry I thought I addressed that. I think this is just a SOTW remnant again.

pkg/cache/v3/delta.go
alecholmez: Ah yes we do, wildcard is no different besides the fact we just track the entire state of the stream rather than specific resources.

alecholmez: Here we could do something like this since both paths need to compute removed resources:
```go
	// If we are handling a wildcard request, we want to respond with all resources
	if st.IsWildcard {
		for name, r := range resources {
			// Since we've already precomputed the version hashes of the new snapshot,
			// we can just set it here to be used for comparison later
			version := snapshot.GetVersionMap()[req.TypeUrl][name]
			nextVersionMap[name] = version
			prevVersion, found := st.ResourceVersions[name]
			if !found || (prevVersion != nextVersionMap[name]) {
			}
		}
	} else {
		// Reply only with the requested resources
		for name, prevVersion := range st.ResourceVersions {
			if r, ok := resources[name]; ok {
				nextVersion := snapshot.GetVersionMap()[req.TypeUrl][name]
				if prevVersion != nextVersion {
				}
			}
		}
	}

	// Compute resources for removal regardless of the request type
	for name := range st.ResourceVersions {
		if _, ok := resources[name]; !ok {
			toRemove = append(toRemove, name)
			delete(nextVersionMap, name)
		}
	}
```

alecholmez:
@snowp sorry for the delay! Just pushed a commit and left some comments/questions.

pkg/cache/v3/snapshot.go
```diff
@@ -87,7 +87,10 @@ func NewResourcesWithTtl(version string, items []types.ResourceWithTtl) Resource
 type Snapshot struct {
 	Resources [types.UnknownType]Resources
 
-	// VersionMap holds the current hash map of all resources in the snapshot. This is only used in delta xDS
+	// VersionMap holds the current hash map of all resources in the snapshot.
+	// This field should remain nil until it is used, at which point should be
+	// instantiated by setting a snapshot while, or creating delta xDS resource watches.
```
snowp, line 92: it would be instantiated by calling ComputeVersionMap or whatever it was called right?

pkg/cache/v3/delta.go
```diff
@@ -42,52 +42,52 @@ func respondDelta(request *DeltaRequest, value chan DeltaResponse, st *stream.St
...
 	}
 
+	// Compute resources for removal regardless of the request type
+	for name := range st.ResourceVersions {
+		if _, ok := resources[name]; !ok {
+			toRemove = append(toRemove, name)
+			delete(nextVersionMap, name)
+		}
+	}
+
```
snowp, line 85: Can we make sure we have test coverage of wildcard removals?

snowp (changes requested):
Thanks this is looking great, jus a few more comments

pkg/cache/v3/snapshot.go
alecholmez: Yes that is correct. That `ConstructVersionMap` function gets called when creating watches or setting a snapshot thought it's probably better to say directly what instantiates that version map instead.

pkg/cache/v3/delta.go
alecholmez: Ah yes certainly! I'll write a test for that now

alecholmez: Just pushed that test!

alecholmez:
@snowp think this should be good to go now

pkg/cache/v3/delta.go
```diff
@@ -0,0 +1,93 @@
...
+				nextVersionMap[name] = nextVersion
+			}
+		}
+	}
+
+	// Compute resources for removal regardless of the request type
+	for name := range st.ResourceVersions {
+		if _, ok := resources[name]; !ok {
+			toRemove = append(toRemove, name)
+			delete(nextVersionMap, name)
```
snowp, line 82: Since we compute `nextVersionMap` from resources already, won't this already be unset? Do we need this delete?

pkg/cache/v3/delta_test.go
```diff
@@ -0,0 +1,222 @@
...
+		t.Fatal("failed to receive snapshot response")
+	}
+}
+
+func TestDeltaRemoveResources(t *testing.T) {
+	c := cache.NewSnapshotCache(false, group{}, logger{t: t})
+	watches := make(map[string]chan cache.DeltaResponse)
+
+	for _, typ := range testTypes {
+		// We don't specify any resource name subscriptions here because we want to make sure we test wildcard
```
snowp, line 94: Do we have test coverage for removals for non wildcard streams as well?

pkg/cache/v3/delta_test.go
```diff
@@ -0,0 +1,222 @@
...
+				}
+				vMap := out.GetNextVersionMap()
+				vm[typ] = vMap
+			case <-time.After(time.Second):
+				t.Fatal("failed to receive snapshot response")
+			}
+		})
+	}
+
+	for _, typ := range testTypes {
```
snowp, line 52: Maybe add a comment making it clearer that we're testing both wildcard and not wildcard?

snowp (changes requested):
Thanks this looks basically ready, just a couple small comments

pkg/cache/v3/delta_test.go
alecholmez: We should yes. The first delta test does both so we create an initial request with wildcard, and then don't use wildcard when we re create the watches. This should go down the logic path of not using a wildcard built response in the delta code.

pkg/cache/v3/delta_test.go
alecholmez: Sure thing! I'll add that above

pkg/cache/v3/delta.go
alecholmez: Ah that's a good point. I guess the `delete` is not necessary. We just still need to find out what resources are slated for `RemovedResources`. So I'll drop the delete but keep the `toRemove` slice

snowp (approved):
LGTM, thanks!

alecholmez:
@snowp awesome! Thank you for the reviews!

PR #420, opened 2021-04-08
Revert back to V3 file structure

alecholmez:

@snowp @mattklein123 this should revert everything back to the previous file structure, as well as fix all the import statements. I still kept out all the previous v2 to v3 scripting and such. 

@phlax this should resolve the issues you're seeing


mattklein123 (approved):
Thanks!

PR #428, opened 2021-05-07
cleanup some staticchecks.

JaredTan95:
Signed-off-by: JaredTan95 <jian.tan@daocloud.io>


alecholmez:
@JaredTan95 thanks for this! Would you mind checking the build? It seems that this does not compile:
```
# github.com/envoyproxy/go-control-plane/pkg/ttl/v3
pkg/ttl/v3/ttl.go:12:93: cannot use &envoy_service_discovery_v3.Resource{} (type *envoy_service_discovery_v3.Resource) as type protoreflect.MessageDescriptor in argument to protoreflect.MessageDescriptor.FullName:
	*envoy_service_discovery_v3.Resource does not implement protoreflect.MessageDescriptor (missing Enums method)
pkg/ttl/v3/ttl.go:34:3: cannot use deltaResourceTypeURL (type protoreflect.FullName) as type string in return argument
```

alecholmez (changes requested):
If you can fix the build this should be good

alecholmez (dismissed):

pkg/server/v3/server.go
```diff
@@ -28,7 +28,6 @@ import (
 
 	clusterservice "github.com/envoyproxy/go-control-plane/envoy/service/cluster/v3"
 	discovery "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
-	discoverygrpc "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
```
kyessenov, line 31: Note: this is an artifact of google import. Google tooling puts gRPC and protos into separate packages.

alecholmez: It seems all the builds passed, would it be worth keeping them separate for the sake of distinguishing between the two? Or since it seemed to build fine merge and let staticcheck be happy?

JaredTan95: If we can keep`discoverygrpc`  import and delete `discovery `?

Issue #433, opened 2021-05-20
Cache for custom resources without individual versioning


jakubdyszkiewicz:
Hey, [Kuma](https://kuma.io) maintainer here 👋 

We reuse XDS protocol with custom resources across different parts of the code
* Exchange Kuma policies between control planes in multizone deployment
* Provide information about Envoys to Prometheus
* Implement HDS

We loved that you introduced `Server#StreamHandler` that can take any resource. It [greatly simplified](https://github.com/kumahq/kuma/blob/master/pkg/mads/v1/service/mads.go) our implementation of servers.

The problem is that `Snapshot` does not support custom resources.
Our solution for the time when we introduced this was to have a copy of SimpleCache with [Snapshot as an interface](https://github.com/kumahq/kuma/blob/master/pkg/util/xds/v3/cache.go#L30)

We would like to get rid of this code and contribute to the project. Since the time we implement this, you introduced `LinearCache + Mux`. This _almost_ solves our issue, but not quite. We don't use `resourceNames` in our custom XDS therefore we don't need versioning on the individual resources.
LinearCache has `UpdateResource` and `DeleteResource` which is inconvenient in our case when compared to `SetSnapshot`. Technically, we could do delta on resources our side and call individual methods, but then every call triggers `watchAll` watches. This means If at a given moment we have 10 resources to update, we will trigger 9 updates for no good reason.

I see two options
1) Create `LinearCache#SetResources` that replaces `resources`, triggers individual watches by name (in our case no watches) including deleted resources, and only then trigger `watchAll` watches.
2) Create a separate cache for custom resources that do not version individual resources

I think the first one is better although I want to ask beforehand so I don't break the intention of `LinearCache`. WDYT?

Thanks!

alecholmez:
@jakubdyszkiewicz Hey there!

Some of these problems you face might actually get solved with Incremental xDS going in :)
I'll shortly be opening PRs for Linear and Mux caching which should solve the issue you brought to light regarding excessive watch updates. 

I think the simple cache implementation could certainly use some massaging here. I actually like the idea of a `Snapshot` as an interface.

@snowp what do you think about that? I think doing so would easily enable delta xDS to do custom resources with the simple cache implementation. Especially if we follow the pattern they used above where you can query for the list of supported types, whatever those may be.

[...]

PR #437, opened 2021-05-24
cache: add SetSnapshot and GetSnapshot to LinearCache

jakubdyszkiewicz:
The implementation discussed in #433

`SetSnapshot` and `GetSnapshot` enables users to use `LinearCache` in a more convenient way if the major use case is sending `DiscoveryRequest` with empty `resourceNames`.


jakubdyszkiewicz:
Sorry about the force push 🙏  I started the work from `master` instead of `main`.

alecholmez:
Thanks for opening this! I'll review

pkg/cache/v3/linear.go
```diff
@@ -239,6 +239,45 @@ func (cache *LinearCache) CreateWatch(request *Request) (chan Response, func())
 	}
 }
 
+// SetResources replaces current resources with a new set of resources
+//
+// If we have individual watches for two resources (there are 2 DiscoveryRequest with resourceNames=res1 and resourceNames=res2)
+// Then if you set new resources, res1 changed but res2 is not, we assume that all resources are changed so both watches will be triggered.
+// That's why this function is well suited for a use case when resourceNames of DiscoveryRequest is empty.
+// This way you don't have to call UpdateResource for every resource which will trigger watch of DiscoveryRequest(resourceNames="") twice,
+// but it will only happen once regardless of how many resources are changed.
```
alecholmez, line 248: I think this comment might be a little much for this function. If I understand this correctly, this is a good hook to use wildcarding when requesting resources (resourceName being empty), maybe we should just allude to that here instead?

pkg/cache/v3/linear.go
```diff
@@ -239,6 +239,45 @@ func (cache *LinearCache) CreateWatch(request *Request) (chan Response, func())
...
+	cache.notifyAll(modified)
+}
+
+// GetResources returns current resources stored in the cache
+func (cache *LinearCache) GetResources() map[string]types.Resource {
+	cache.mu.Lock()
+	defer cache.mu.Unlock()
+	return cache.resources
+}
+
```
alecholmez, line 280: Just a nit pick but can we move both these functions above the `CreateWatch()`? I'd like to keep them bundled with Update/DeleteResource().

pkg/cache/v3/linear.go
```diff
@@ -239,6 +239,45 @@ func (cache *LinearCache) CreateWatch(request *Request) (chan Response, func())
...
+		cache.versionVector[name] = cache.version
+		modified[name] = struct{}{}
+	}
+
+	cache.notifyAll(modified)
+}
+
+// GetResources returns current resources stored in the cache
+func (cache *LinearCache) GetResources() map[string]types.Resource {
+	cache.mu.Lock()
```
alecholmez, line 276: This is just a read right? Could we change this to an `RLock()`?

alecholmez (changes requested):
Just a few things so far

pkg/cache/v3/linear.go
jakubdyszkiewicz: I rewrote the comment. Please take a look if it's ok, if not I could use some guidance. Thanks!

pkg/cache/v3/linear.go
jakubdyszkiewicz: Right, changed to RWLock and changed `NumWatces` to also only use `RLock()`

jakubdyszkiewicz:
hey, @alecholmez do you have any more suggestions?

alecholmez:
@jakubdyszkiewicz I think this is good from my end, can you address the merge conflicts?

@snowp you have any thoughts?

pkg/cache/v3/linear.go
```diff
@@ -165,6 +165,41 @@ func (cache *LinearCache) DeleteResource(name string) error {
...
+// This function is useful for use cases when DiscoveryRequest#resourceNames is empty.
+// This way watches that follow all resources are triggered only once regardless of how many resources are changed.
+func (cache *LinearCache) SetResources(resources map[string]types.Resource) {
+	cache.mu.Lock()
+	defer cache.mu.Unlock()
+
+	cache.version += 1
+
+	modified := map[string]struct{}{}
+	for name := range cache.resources {
```
snowp, line 178: Some comments would be make this easier to follow, like `// First account for removed resources.` and a similar one for the other for loop

pkg/cache/v3/linear.go
```diff
@@ -165,6 +165,41 @@ func (cache *LinearCache) DeleteResource(name string) error {
...
+		if _, found := resources[name]; !found {
+			delete(cache.versionVector, name)
+			modified[name] = struct{}{}
+		}
+	}
+
+	cache.resources = resources
+	for name := range resources {
+		// We assume all resources passed to SetResources are changed.
+		// Otherwise we would have to do proto.Equal on resources which is pretty expensive operation
```
snowp, line 188: There are other solutions to this (like per version resource values like how Delta is modeled), but it's probably fine to leave it as this for now.

pkg/cache/v3/linear_test.go
```diff
@@ -129,6 +130,64 @@ func TestLinearBasic(t *testing.T) {
 	verifyResponse(t, w, "3", 2)
 }
 
+func TestLinearSetResources(t *testing.T) {
+	c := NewLinearCache(testType)
+
+	// Create new resources
+	w1, _ := c.CreateWatch(&Request{ResourceNames: []string{"a"}, TypeUrl: testType, VersionInfo: "0"})
+	mustBlock(t, w1)
+	w, _ := c.CreateWatch(&Request{TypeUrl: testType, VersionInfo: "0"})
```
snowp, line 139: nit: name this variable w2 for consistency

snowp (commented):
Thanks this looks good to me for the most part, just a few minor nits

jakubdyszkiewicz:
@snowp fixed comments and conflicts

snowp (approved):
Thanks!

snowp:
@alecholmez for merge

alecholmez (approved):
This looks good @jakubdyszkiewicz, thanks for the contribution!

PR #438, opened 2021-05-25
sotw code cleanup

alecholmez:
- moves server unit tests to `assert`
- removes code duplication inside of the watches


alecholmez:
@snowp I ported the logic over from delta to sotw so we cleanup the existing code a bit 😄

snowp:
There is a slight perf concern (additional goroutines per stream) with making this change, I wonder if we want to wait until we have mem/speed benchmarks until we make a change like this to SotW since it's in active use?

alecholmez:
@snowp I'm not opposed to that. go routines are relatively lightweight but I think it's a good idea to measure the current performance first.

PR #440, opened 2021-05-26
fixed integration test for ads

alecholmez:

### DIFF
--- a/pkg/server/v3/delta_test.go
+++ b/pkg/server/v3/delta_test.go
@@ -414,9 +414,8 @@ func TestDeltaOpaqueRequestsChannelMuxing(t *testing.T) {
 	resp := makeMockDeltaStream(t)
 	for i := 0; i < 10; i++ {
 		resp.recv <- &discovery.DeltaDiscoveryRequest{
-			Node:    node,
-			TypeUrl: fmt.Sprintf("%s%d", opaqueType, i%2),
-			// each subsequent request is assumed to supercede the previous request
+			Node:                   node,
+			TypeUrl:                fmt.Sprintf("%s%d", opaqueType, i%2),
 			ResourceNamesSubscribe: []string{fmt.Sprintf("%d", i)},
 		}
 	}
--- a/pkg/server/v3/delta_test.go
+++ b/pkg/server/v3/delta_test.go
@@ -379,3 +379,82 @@ func TestDeltaAggregatedHandlers(t *testing.T) {
 		}
 	}
 }
+
+func TestDeltaAggregateRequestType(t *testing.T) {
+	config := makeMockConfigWatcher()
+	s := server.NewServer(context.Background(), config, server.CallbackFuncs{})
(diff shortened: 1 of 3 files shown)


pkg/server/v3/delta_test.go
```diff
@@ -379,3 +379,82 @@ func TestDeltaAggregatedHandlers(t *testing.T) {
...
+}
+
+func TestDeltaOpaqueRequestsChannelMuxing(t *testing.T) {
+	config := makeMockConfigWatcher()
+	resp := makeMockDeltaStream(t)
+	for i := 0; i < 10; i++ {
+		resp.recv <- &discovery.DeltaDiscoveryRequest{
+			Node:    node,
+			TypeUrl: fmt.Sprintf("%s%d", opaqueType, i%2),
+			// each subsequent request is assumed to supercede the previous request
```
snowp, line 419: supercede seems like the wrong term here since each request is for a different type?

sample/bootstrap-delta-ads.yaml
```diff
@@ -35,18 +35,4 @@ static_resources:
...
-        rtds_config:
-          resource_api_version: V3
-          ads: {}
-        name: runtime-0
-    - name: runtime-1
-      rtds_layer:
-        rtds_config:
-          resource_api_version: V3
-          ads: {}
-        name: runtime-1
```
snowp, line 52: Can you file an issue to look into this? I don't see why this shouldn't work

snowp (changes requested):
Thanks, few comments

sample/bootstrap-delta-ads.yaml
alecholmez: @snowp sure thing, do you want an issue in Envoy as well? Or just in the control-plane code

snowp: Let's keep it here for now until we can figure out what's wrong. Would be great if you could include the logs you got earlier

pkg/server/v3/delta_test.go
alecholmez: Maybe? I think they're meaning to say each request takes places of the previous one. Could also change that to `succeeds`?

sample/bootstrap-delta-ads.yaml
alecholmez: Okay cool I'll file an issue

pkg/server/v3/delta_test.go
snowp: If we don't know what it means maybe just remove it :)

alecholmez: Ha works for me

sample/bootstrap-delta-ads.yaml
alecholmez: https://github.com/envoyproxy/go-control-plane/issues/442

snowp (approved):
Thanks!

PR #376, opened 2020-12-06
server api without linearization

fxposter:
@jyotimahapatra @kyessenov Followup of #375 

Do you envision this to have any problems for clients to migrate to? Let's put linearization aside for now.

PR #443, opened 2021-05-31
server api without linearization

fxposter:
This is the rebased version of https://github.com/envoyproxy/go-control-plane/pull/376.

@kyessenov @jessicayuen @snowp @alecholmez

### DIFF
--- a/pkg/cache/v3/cache.go
+++ b/pkg/cache/v3/cache.go
@@ -54,7 +54,7 @@ type ConfigWatcher interface {
 	//
 	// Cancel is an optional function to release resources in the producer. If
 	// provided, the consumer may call this function multiple times.
-	CreateWatch(*Request) (value chan Response, cancel func())
+	CreateWatch(*Request, chan Response) (cancel func())
 
 	// CreateDeltaWatch returns a new open incremental xDS watch.
 	//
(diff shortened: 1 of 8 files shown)


fxposter:
This will conflict with https://github.com/envoyproxy/go-control-plane/pull/438

fxposter:
For those who does not want to dig deep into all previous issues: this changes the API between cache and server packages to allow to maintain the order of responses of different types from the cache side.
Basically, when you do ADS and send many changes to the client at once for different types - you lose their ordering, ie: listeners should go before routes, but if your cache implementation sends both listeners and routes - you end up in large [select statement](https://github.com/envoyproxy/go-control-plane/blob/main/pkg/server/sotw/v3/server.go#L198-L269), and select statement basically lose order of the changes that you sent to different channels. So the idea is to be able to unify those channels and simplify implementation (subject to future PRs, which would mane #438 not needed as well).

fxposter:
I was pretty busy last 2 months, but previously it was tested and just needed to rebase to merge: https://github.com/envoyproxy/go-control-plane/pull/376#issuecomment-818926309. Since there were many changes since then - maybe we should do another round of tests, but from code perspective - nothing significant has changed since previous PR.

alecholmez (approved):
This looks great, I won't stand in the way since this was already good to merge previously. 

No worries on the conflict we aren't sure if we want to go that route (waiting on some benchmarks first) so I can just address that when I need to.

I'll probably open a PR that follows this for Incremental soon, it would be helpful if you could put your eyes on that when it's opened to make sure I don't miss anything.

PR #447, opened 2021-06-11
move delta to new linearization API

### LINKED ISSUE #446: incrementalXDS does not cancel existing watch
Cancel on watch object is called when watch does not exist in the map (https://github.com/envoyproxy/go-control-plane/blob/main/pkg/server/delta/v3/server.go#L168).  (It is currently no-op since when 'ok' is false, watch will have the zero value. )
Cancel should be called when 'ok' is true since we expect cancel to be called when we find watch in the map. 

```
// cancel existing watch to (re-)request a newer version
watch, ok := watches.deltaWatches[typeURL]
if !ok {
   ...
   // We must signal goroutine termination to prevent a race between the cancel closing the watch
   // and the producer closing the watch.
   watch.Cancel()
```

alecholmez:

### DIFF
--- a/pkg/cache/v3/cache.go
+++ b/pkg/cache/v3/cache.go
@@ -47,10 +47,7 @@ type ConfigWatcher interface {
 	// CreateWatch returns a new open watch from a non-empty request.
 	// An individual consumer normally issues a single open watch by each type URL.
 	//
-	// Value channel produces requested resources, once they are available.  If
-	// the channel is closed prior to cancellation of the watch, an unrecoverable
-	// error has occurred in the producer, and the consumer should close the
-	// corresponding stream.
+	// Value channel produces requested resources, once they are available.
 	//
 	// Cancel is an optional function to release resources in the producer. If
 	// provided, the consumer may call this function multiple times.
@@ -59,9 +56,7 @@ type ConfigWatcher interface {
 	// CreateDeltaWatch returns a new open incremental xDS watch.
 	//
 	// Value channel produces requested resources, or spontaneous updates in accordance
-	// with the incremental xDS specification. If the channel is closed
-	// prior to cancellation of the watch, an unrecoverable error has occurred in the producer,
-	// and the consumer should close the corresponding stream.
+	// with the incremental xDS specification.
 	//
 	// Cancel is an optional function to release resources in the producer. If
 	// provided, the consumer may call this function multiple times.
--- a/pkg/cache/v3/cache.go
+++ b/pkg/cache/v3/cache.go
@@ -47,7 +47,7 @@ type ConfigWatcher interface {
 	// CreateWatch returns a new open watch from a non-empty request.
 	// An individual consumer normally issues a single open watch by each type URL.
 	//
-	// Value channel produces requested resources, once they are available.
+	// The provided channel produces requested resources as responses, once they are available.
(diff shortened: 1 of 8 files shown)


snowp (approved):
Great, thanks!

pkg/cache/v3/cache.go
```diff
@@ -65,7 +65,7 @@ type ConfigWatcher interface {
 	//
```
snowp, line 65: Should the comments above be updated? They seem to talk about the old channel behavior

pkg/server/delta/v3/server.go
```diff
@@ -170,33 +171,17 @@ func (s *server) processDelta(str stream.DeltaStream, reqCh <-chan *discovery.De
...
-						break
-					case <-terminate:
-						break
+					resp, more := <-watch.responses
+					if more {
+						watches.deltaMuxedResponses <- resp
 					}
 				}()
+			} else {
+				watch.Cancel()
```
snowp, line 184: Is this new behavior?

snowp (commented):
Actually had a few comments :) Looks good just some questions

pkg/cache/v3/cache.go
alecholmez: Oh good point, I guess we can simply everything regarding the cancellation behavior. Something like:
```go
type ConfigWatcher interface {
	// CreateWatch returns a new open watch from a non-empty request.
	//
	// Value channel produces requested resources, once they are available.
	//
	CreateWatch(*Request, chan Response) (cancel func())

	// CreateDeltaWatch returns a new open incremental xDS watch.
	//
	// Value channel produces requested resources, or spontaneous updates in accordance
	// with the incremental xDS specification.
	//
	CreateDeltaWatch(*DeltaRequest, stream.StreamState, chan DeltaResponse) (cancel func())
}
```

pkg/server/delta/v3/server.go
alecholmez: Ah yes I thought I removed that. It was to address: https://github.com/envoyproxy/go-control-plane/issues/446 but it seems https://github.com/envoyproxy/go-control-plane/issues/448 has outlined some more work to be done. I can remove this since it's out of scope for this PR.

alecholmez:
Should be good now @snowp

pkg/cache/v3/cache.go
```diff
@@ -47,10 +47,7 @@ type ConfigWatcher interface {
 	// CreateWatch returns a new open watch from a non-empty request.
 	// An individual consumer normally issues a single open watch by each type URL.
 	//
-	// Value channel produces requested resources, once they are available.  If
-	// the channel is closed prior to cancellation of the watch, an unrecoverable
-	// error has occurred in the producer, and the consumer should close the
-	// corresponding stream.
+	// Value channel produces requested resources, once they are available.
```
snowp, line 50: I think Value used to be the parameter name, so now this reads a bit odd. Maybe name the function parameters or reword this a bit?

alecholmez: Ah yea thats a good point, I'll change that so it makes more sense

alecholmez:
@snowp is this good to merge?

snowp (approved):
LGTM, thanks!

PR #451, opened 2021-06-21
server: sotw dedupe and linearization

alecholmez:

I closed my previous sotw dedupe PR in favor of this one since we merged in the linearization API code which this now adheres too. Still need to benchmark


dontlaugh:
@alecholmez you want to do benchmarks for this, right

alecholmez:
@dontlaugh yup, we'd like to see how this compares with the extra go routine overhead to the current system

pkg/server/sotw/v3/watches.go
```diff
@@ -0,0 +1,75 @@
...
+		watch.index = index
+		index++
+	}
+
+	w.cases = newCases
+}
+
+// watch contains the necessary modifiables for receiving resource responses
+type watch struct {
+	selectCase reflect.SelectCase
```
snowp, line 60: Do we need to track it here in addition to cases?

pkg/server/sotw/v3/watches.go
```diff
@@ -0,0 +1,75 @@
...
+			Dir:  reflect.SelectRecv,
+			Chan: reflect.ValueOf(ctx.Done()),
+		},
+		{
+			Dir:  reflect.SelectRecv,
+			Chan: reflect.ValueOf(reqCh),
+		},
+	}
+
+	index := 2
```
snowp, line 48: len(newCases) might seem less magical

pkg/server/sotw/v3/watches.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// Cancel all watches
+func (w *watches) Cancel() {
+	for _, watch := range w.responders {
+		watch.Cancel()
+	}
+}
+
+// recomputeWatches will analyze the currently typed list of the known watches and increase the known list of dynamic channels if needed
```
snowp, line 35: I would just say rebuilds the list of dynamic channels, I think it could go both up and down

pkg/server/sotw/v3/watches.go
```diff
@@ -0,0 +1,75 @@
...
+type watches struct {
+	responders map[string]*watch
+
+	// indexes is a list of indexes for each dynamic select case which match to a watch
+	cases []reflect.SelectCase
+}
+
+// newWatches creates and initializes watches.
+func newWatches() watches {
+	// deltaMuxedResponses needs a buffer to release go-routines populating it
```
snowp, line 21: This comment seems out of date?

pkg/server/sotw/v3/watches.go
```diff
@@ -0,0 +1,75 @@
...
+	return watches{
+		responders: make(map[string]*watch, int(types.UnknownType)),
+		cases:      make([]reflect.SelectCase, 2), // We use 2 for the default computation here: ctx.Done() + reqCh.Recv()
```
snowp, line 24: Don't we want this to be properly initialized? I assume we'll call RecompueWatches before this ever get read, so maybe just call it directly in newWatches or avoid this logic since it doesn't matter?

pkg/server/sotw/v3/watches.go
```diff
@@ -0,0 +1,75 @@
...
+// watch contains the necessary modifiables for receiving resource responses
+type watch struct {
+	selectCase reflect.SelectCase
+	cancel     func()
+	nonce      string
+
+	// Index is used to track the location of this channel in watches. This allows us
+	// to update the channel used at this slot without recomputing the entire list of select
+	// statements.
+	index int
```
snowp, line 67: When do we use this?

pkg/server/v3/server_test.go
```diff
@@ -378,99 +370,73 @@ func TestFetch(t *testing.T) {
...
-		t.Errorf("unexpected empty or error for secrets: %v", err)
-	}
-	if out, err := s.FetchRuntime(context.Background(), &discovery.DiscoveryRequest{Node: node}); out == nil || err != nil {
-		t.Errorf("unexpected empty or error for runtime: %v", err)
-	}
-	if out, err := s.FetchExtensionConfigs(context.Background(), &discovery.DiscoveryRequest{Node: node}); out == nil || err != nil {
-		t.Errorf("unexpected empty or error for extensionConfigs: %v", err)
-	}
+	out, err := s.FetchEndpoints(context.Background(), &discovery.DiscoveryRequest{Node: node})
+	assert.False(t, out == nil || err != nil)
```
snowp, line 374: Seems like it would be nicer to have these be
```
assert.NotNil(t, out)
assert.NoError(t, err)
```
?

pkg/server/v3/server_test.go
```diff
@@ -378,99 +370,73 @@ func TestFetch(t *testing.T) {
...
+	assert.False(t, out != nil || err == nil)
 
 	// verify fetch callbacks
-	if want := 13; requestCount != want {
-		t.Errorf("unexpected number of fetch requests: got %d, want %d", requestCount, want)
-	}
-	if want := 8; responseCount != want {
-		t.Errorf("unexpected number of fetch responses: got %d, want %d", responseCount, want)
-	}
+	assert.True(t, requestCount == 10)
```
snowp, line 438: assert.Equal

pkg/server/v3/server_test.go
```diff
@@ -625,12 +588,9 @@ func TestCancellations(t *testing.T) {
...
 	s := server.NewServer(context.Background(), config, server.CallbackFuncs{})
-	if err := s.StreamAggregatedResources(resp); err != nil {
-		t.Errorf("StreamAggregatedResources() => got %v, want no error", err)
-	}
-	if config.watches != 0 {
-		t.Errorf("Expect all watches canceled, got %q", config.watches)
-	}
+	err := s.StreamAggregatedResources(resp)
+	assert.NoError(t, err)
+	assert.True(t, config.watches == 0)
```
snowp, line 593: Equal

pkg/server/v3/server_test.go
```diff
@@ -646,12 +606,9 @@ func TestOpaqueRequestsChannelMuxing(t *testing.T) {
...
```
snowp, line 611: Equal

snowp (commented):
Thanks this looks good for the most part, just a few comments

snowp (dismissed):
Seems like CI is red?

pkg/server/sotw/v3/server.go
```diff
@@ -359,88 +190,46 @@ func (s *server) process(stream Stream, reqCh <-chan *discovery.DiscoveryRequest
...
+			typeURL := req.GetTypeUrl()
+
+			w, ok := watches.responders[typeURL]
+			if !ok {
+				watches.responders[typeURL] = &watch{}
+
+				w = watches.responders[typeURL]
+				watches.RecomputeWatches(s.ctx, reqCh)
+			}
+			if w.nonce == "" || w.nonce == nonce {
```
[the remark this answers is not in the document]
alecholmez: Ah yea that is quite a bit more readable I'll move to that.

pkg/server/sotw/v3/watches.go
```diff
@@ -0,0 +1,73 @@
...
+			Dir:  reflect.SelectRecv,
+			Chan: reflect.ValueOf(req),
+		},
+	}
+
+	index := len(cases)
+	for _, watch := range w.responders {
+		cases = append(cases, w.cases[watch.index])
+		watch.index = index
+		index++
```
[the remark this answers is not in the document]
alecholmez: Do you imagine that being too expensive at scale? I don't really ever imagine that slice becoming too large to the point where it's particularly an issue recalculating it, but when 10k+ streams are open and watches are being opened and canceled a lot that might be a problem. Istio has an issue like this right now where they see massive traffic and data movement (lots of config updates) on startup but once the mesh stabilizes it doesn't really become an expensive operation anymore. Unless things come up and down of course. What are your thoughts on that? I'd imagine this recompute would get called a lot in that scenario. I'm more inclined to deal with a bit more complexity on the code side for a little edge in performance

pkg/server/sotw/v3/watches.go
```diff
@@ -0,0 +1,73 @@
...
+
+// Cancel all watches
+func (w *watches) Cancel() {
+	for _, watch := range w.responders {
+		watch.Cancel()
+	}
+}
+
+// recomputeWatches rebuilds the known list of dynamic channels if needed
+func (w *watches) RecomputeWatches(ctx context.Context, req <-chan *discovery.DiscoveryRequest) {
```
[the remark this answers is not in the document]
alecholmez: I think a `watch` in this particular server code sense is more of an internal data behavior. In the previous code, what the `watch` and `watches` struct replace are overly repeated bits that are for moving data around (more or less cache interaction).

If you do have suggestions for how we can streamline that I'd love to hear them. I mostly just refactored that old code to these new structs simply to reduce repetitive lines with similar/same functionality (nothing too special on that front :joy:).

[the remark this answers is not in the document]
jpeach: Yeh, I'd suggest renaming `RecomputeWatches()` to `recompute()`, and `Cancel()` to `cancel()` on both `watches` and `watch`. To me, that makes it really clear that this is all internal implementation details,

pkg/server/sotw/v3/watches.go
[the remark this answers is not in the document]
jpeach: I tried this out and IMHO it seems cleaner ([gist](https://gist.github.com/jpeach/128b5199a564928d396545f92f7fc3a8)).

There's one watch per typeURL (so maybe a dozen would be a lot?), so it doesn't seem likely that this sort of micro-optimization would have a big impact.

pkg/server/sotw/v3/watches.go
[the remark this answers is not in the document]
alecholmez: That makes sense. I think as the code stands the `cancel()` method on watch will cause a name collision with the field `cancel func()`. Maybe rename those to close instead?

pkg/server/sotw/v3/watches.go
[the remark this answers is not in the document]
alecholmez: Awesome thanks for that diff :) I just made the changes and I'll push

alecholmez:
@jpeach can you give this another pass when you get some free time?

pkg/server/sotw/v3/watches.go
```diff
@@ -0,0 +1,75 @@
...
+	discovery "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
+	"github.com/envoyproxy/go-control-plane/pkg/cache/types"
+	"github.com/envoyproxy/go-control-plane/pkg/cache/v3"
+)
+
+// watches for all xDS resource types
+type watches struct {
+	responders map[string]*watch
+
+	// indexes is a list of indexes for each dynamic select case which match to a watch
```
jpeach, line 16:
```suggestion
	// cases is a dynamic select case which matches the watches.
```

jpeach (dismissed):
:+1:

jpeach (approved):

alecholmez:
@snowp want to give the final stamp on this one?

rueian:
Hi @alecholmez, @jpeach,

I based on this PR to solve another deadlock in sotw server (#530, #531). Please take a look.

alecholmez:
@jpeach is this good to merge? @rueian once this is merged can you rebase your branch off main instead?

jpeach:
> @jpeach is this good to merge?

Yup!

jpeach (approved):

PR #452, opened 2021-06-24
benchmarks: micro benchmarking

alecholmez:


alecholmez:
@snowp sorry about this, somehow my fork diverged (I guess since I've been working out of here now and haven't updated it) so I went and cherry picked and addressed your latest comments on this branch. Major headache but I think this should be in a good state now

PR #454, opened 2021-06-28
Add basic golangci-lint integration.

jpeach:
Add a "lint" target to the build that runs golangci-lint. This needs to
be run manually for now. CI can be wired up later once the issues are
fixed and there is consensus on the set of linters to run.

This updates #414.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -26,3 +26,12 @@ linters-settings:
 issues:
   exclude:
   exclude-rules:
+  - text: "SA1019: package github.com/golang/protobuf/jsonpb is deprecated"
+    linters:
+    - staticcheck
+  - text: "SA1019: package github.com/golang/protobuf/proto is deprecated"
+    linters:
+    - staticcheck
+  - text: "SA1019: proto.MessageName is deprecated"
+    linters:
+    - staticcheck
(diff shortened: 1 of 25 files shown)


alecholmez:
@mattklein123 You might want to take a quick look at this since you reworked the CI

alecholmez:
I think excluding these:
```
issues:
  exclude:
  exclude-rules:
  - text: "SA1019: package github.com/golang/protobuf/jsonpb is deprecated"
    linters:
    - staticcheck
  - text: "SA1019: package github.com/golang/protobuf/proto is deprecated"
    linters:
    - staticcheck
  - text: "SA1019: proto.MessageName is deprecated"
    linters:
    - staticcheck
```
might not be what we want. We might as well fix them if at all possible

jpeach:
> might not be what we want. We might as well fix them if at all possible. 

Migrating to the new proto API has effects on downstream consumers and I suspect is probably worth a separate discussion.

> I know we copy around a lock quite a lot in the SOTW code so that should also reduce the lint issues a bit

Yeh that gets picked up. Not fixed here yet.

jpeach:
Fixed the remaining lint issues. I'll rebase and squash before marking this ready.

jpeach:
Closing in favor of #455 because I messed up the branching in my fork :-/

PR #455, opened 2021-06-30
Add golangci lint integration.

jpeach:
Add golangci-lint integration and fix or ignore all the resulting issues. In general, the things I ignored were names that can't be changed without a deprecation and migration.

This updates #414.

### DIFF
--- a/internal/example/main/main.go
+++ b/internal/example/main/main.go
@@ -25,12 +25,8 @@ import (
 )
 
 var (
-	l example.Logger
-
-	port     uint
-	basePort uint
-	mode     string
-
+	l      example.Logger
+	port   uint
 	nodeID string
 )
 
--- a/pkg/cache/v3/cache.go
+++ b/pkg/cache/v3/cache.go
@@ -121,7 +121,7 @@ type RawResponse struct {
 	Version string
 
 	// Resources to be included in the response.
-	Resources []types.ResourceWithTtl
+	Resources []types.ResourceWithTTL
 
 	// Whether this is a heartbeat response. For xDS versions that support TTL, this
 	// will be converted into a response that doesn't contain the actual resource protobuf.
@@ -306,11 +306,11 @@ func (r *RawDeltaResponse) GetContext() context.Context {
 
 var deltaResourceTypeURL = "type.googleapis.com/" + proto.MessageName(&discovery.Resource{})
 
-func (r *RawResponse) maybeCreateTTLResource(resource types.ResourceWithTtl) (types.Resource, string, error) {
-	if resource.Ttl != nil {
+func (r *RawResponse) maybeCreateTTLResource(resource types.ResourceWithTTL) (types.Resource, string, error) {
+	if resource.TTL != nil {
 		wrappedResource := &discovery.Resource{
 			Name: GetResourceName(resource.Resource),
-			Ttl:  ptypes.DurationProto(*resource.Ttl),
+			Ttl:  ptypes.DurationProto(*resource.TTL),
 		}
 
 		if !r.Heartbeat {
--- a/pkg/cache/v3/cache.go
+++ b/pkg/cache/v3/cache.go
@@ -40,7 +40,7 @@ type DeltaRequest = discovery.DeltaDiscoveryRequest
 
 // ConfigWatcher requests watches for configuration resources by a node, last
 // applied version identifier, and resource names hint. The watch should send
-// the responses when they are ready. The watch can be cancelled by the
+// the responses when they are ready. The watch can be canceled by the
 // consumer, in effect terminating the watch for the request.
 // ConfigWatcher implementation must be thread-safe.
 type ConfigWatcher interface {
@@ -164,18 +164,18 @@ type RawDeltaResponse struct {
 var _ Response = &RawResponse{}
 var _ DeltaResponse = &RawDeltaResponse{}
 
-// PassthroughResponse is a pre constructed xDS response that need not go through marshalling transformations.
+// PassthroughResponse is a pre constructed xDS response that need not go through marshaling transformations.
 type PassthroughResponse struct {
 	// Request is the original request.
 	Request *discovery.DiscoveryRequest
 
-	// The discovery response that needs to be sent as is, without any marshalling transformations.
+	// The discovery response that needs to be sent as is, without any marshaling transformations.
 	DiscoveryResponse *discovery.DiscoveryResponse
 
 	ctx context.Context
 }
 
-// DeltaPassthroughResponse is a pre constructed xDS response that need not go through marshalling transformations.
+// DeltaPassthroughResponse is a pre constructed xDS response that need not go through marshaling transformations.
 type DeltaPassthroughResponse struct {
 	// Request is the latest delta request on the stream
 	DeltaRequest *discovery.DeltaDiscoveryRequest
@@ -183,7 +183,7 @@ type DeltaPassthroughResponse struct {
 	// NextVersionMap consists of updated version mappings after this response is applied
 	NextVersionMap map[string]string
 
-	// This discovery response that needs to be sent as is, without any marshalling transformations
+	// This discovery response that needs to be sent as is, without any marshaling transformations
 	DeltaDiscoveryResponse *discovery.DeltaDiscoveryResponse
 
 	ctx context.Context
@@ -192,9 +192,9 @@ type DeltaPassthroughResponse struct {
 var _ Response = &PassthroughResponse{}
 var _ DeltaResponse = &DeltaPassthroughResponse{}
 
-// GetDiscoveryResponse performs the marshalling the first time its called and uses the cached response subsequently.
-// This is necessary because the marshalled response does not change across the calls.
-// This caching behavior is important in high throughput scenarios because grpc marshalling has a cost and it drives the cpu utilization under load.
+// GetDiscoveryResponse performs the marshaling the first time its called and uses the cached response subsequently.
+// This is necessary because the marshaled response does not change across the calls.
+// This caching behavior is important in high throughput scenarios because grpc marshaling has a cost and it drives the cpu utilization under load.
 func (r *RawResponse) GetDiscoveryResponse() (*discovery.DiscoveryResponse, error) {
 
 	marshaledResponse := r.marshaledResponse.Load()
@@ -204,7 +204,7 @@ func (r *RawResponse) GetDiscoveryResponse() (*discovery.DiscoveryResponse, erro
 		marshaledResources := make([]*any.Any, len(r.Resources))
 
 		for i, resource := range r.Resources {
-			maybeTtldResource, resourceType, err := r.maybeCreateTtlResource(resource)
+			maybeTtldResource, resourceType, err := r.maybeCreateTTLResource(resource)
 			if err != nil {
 				return nil, err
 			}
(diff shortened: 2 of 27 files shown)


jpeach:
Rebased.

internal/example/main/main.go
```diff
@@ -25,12 +25,8 @@ import (
 )
 
 var (
-	l example.Logger
-
-	port     uint
-	basePort uint
-	mode     string
```
alecholmez, line 32: was this unused?

pkg/cache/v3/cache.go
```diff
@@ -186,7 +186,7 @@ func (r *RawResponse) GetDiscoveryResponse() (*discovery.DiscoveryResponse, erro
 		marshaledResources := make([]*any.Any, len(r.Resources))
 
 		for i, resource := range r.Resources {
-			maybeTtldResource, resourceType, err := r.maybeCreateTtlResource(resource)
+			maybeTtldResource, resourceType, err := r.maybeCreateTTLResource(resource)
```
alecholmez, line 189: I'd rather keep the TTL stuff consistent. Maybe change `ResourceWithTtl` to `ResourceWithTTL`?

pkg/cache/v3/delta_test.go
```diff
@@ -175,24 +175,21 @@ func TestConcurrentSetDeltaWatch(t *testing.T) {
...
-					cancel = c.CreateDeltaWatch(&discovery.DeltaDiscoveryRequest{
+					cancel := c.CreateDeltaWatch(&discovery.DeltaDiscoveryRequest{
 						Node: &core.Node{
 							Id: id,
 						},
 						TypeUrl:                rsrc.EndpointType,
 						ResourceNamesSubscribe: []string{clusterName},
 					}, stream.NewStreamState(false, make(map[string]string)), responses)
+
+					defer cancel()
```
alecholmez, line 192: Did this not cause test failures?

pkg/cache/v3/mux.go
```diff
@@ -37,6 +37,16 @@ type MuxCache struct {
 var _ Cache = &MuxCache{}
 
 func (mux *MuxCache) CreateWatch(request *Request, value chan Response) func() {
+	// Passing a Request by value to Classify triggers a govet copylocks
+	// error. This is because protobuf messages specifically embed
+	// a Mutex in order to trigger this check (i.e. not for actually
+	// locking fields).
+	//
+	// So in this specific case, it happens to be safe to copy the
```
alecholmez, line 45: I would not be opposed to changing the API to not copy lock values. We did this for delta.

pkg/cache/v3/simple.go
```diff
@@ -154,7 +150,7 @@ func NewSnapshotCacheWithHeartbeating(ctx context.Context, ads bool, hash NodeHa
 	return cache
 }
 
-func (cache *snapshotCache) sendHeartbeats(ctx context.Context, node string) {
+func (cache *snapshotCache) sendHeartbeats(_ context.Context, node string) {
```
alecholmez, line 153: Were there linters complaining about this? I generally think this is bad to have in param naming schemes

pkg/cache/v3/snapshot.go
```diff
@@ -66,15 +66,15 @@ func IndexRawResourcesByName(items []types.Resource) map[string]types.Resource {
...
-		itemsWithTtl = append(itemsWithTtl, types.ResourceWithTtl{Resource: item})
+		itemsWithTTL = append(itemsWithTTL, types.ResourceWithTtl{Resource: item})
 	}
-	return NewResourcesWithTtl(version, itemsWithTtl)
+	return NewResourcesWithTtl(version, itemsWithTTL)
 }
 
 // NewResources creates a new resource group.
-func NewResourcesWithTtl(version string, items []types.ResourceWithTtl) Resources {
+func NewResourcesWithTtl(version string, items []types.ResourceWithTtl) Resources { // nolint:golint,revive
```
alecholmez, line 77: Consistency with the `TTL` casing

pkg/cache/v3/snapshot.go
```diff
@@ -137,9 +137,9 @@ func NewSnapshotWithResources(version string, resources SnapshotResources) Snaps
 	return out
 }
 
-type ResourceWithTtl struct {
+type ResourceWithTtl struct { // nolint:golint,revive
```
alecholmez, line 140: Same here

pkg/resource/v3/resource.go
```diff
@@ -43,7 +43,7 @@ func GetHTTPConnectionManager(filter *listener.Filter) *hcm.HttpConnectionManage
 
 	// use typed config if available
 	if typedConfig := filter.GetTypedConfig(); typedConfig != nil {
-		ptypes.UnmarshalAny(typedConfig, config)
+		_ = ptypes.UnmarshalAny(typedConfig, config)
```
alecholmez, line 46: Why the value ignore?

pkg/server/v3/gateway_test.go
```diff
@@ -100,7 +100,7 @@ func TestGateway(t *testing.T) {
 		if err != nil {
 			t.Fatal(err)
 		}
-		resp, code, err := gtw.ServeHTTP(req)
+		resp, code, _ := gtw.ServeHTTP(req)
```
alecholmez, line 103: Let's not ignore the errors here, I'd rather add an error check

pkg/server/v3/gateway_test.go
```diff
@@ -114,7 +114,7 @@ func TestGateway(t *testing.T) {
...
```
alecholmez, line 117: Same

pkg/test/main/main.go
```diff
@@ -307,7 +311,7 @@ func callEcho() (int, int) {
 				ch <- err
 				return
 			}
-			defer req.Body.Close()
+			defer func() { _ = req.Body.Close() }()
```
alecholmez, line 314: More unnamed values

pkg/test/server.go
```diff
@@ -111,7 +111,7 @@ func RunManagementGateway(ctx context.Context, srv server.Server, port uint, lg
 	<-ctx.Done()
 
 	// Cleanup our gateway if we receive a shutdown
-	server.Shutdown(ctx)
+	_ = server.Shutdown(ctx)
```
alecholmez, line 114: Unnamed value

alecholmez (changes requested):
Thanks for the changes and fixing the lint issues! Just a few comments for you, I'd rather stay away from the unnamed values and params especially in situations where theres no need for a partial value read or something of that nature.

alecholmez:
I think @snowp or @mattklein123 will need to change CI to run the linter. I don't have those permissions

pkg/resource/v3/resource.go
jpeach: IIRC this is errcheck again :)

pkg/cache/v3/snapshot.go
jpeach: So Go naming conventions really want this to be `TTL`, but I didn't want to break any compatibility in this PR. Since this is a public function, I just annotated it.

Maybe rename and deprecate?

```Go
// Deprecated 
func NewResourcesWithTtl(version string, items []types.ResourceWithTtl) Resources {
        return NewResourcesWithTTL(...)
}

func NewResourcesWithTTL(version string, items []types.ResourceWithTtl) Resources {
...
}
```

pkg/cache/v3/simple.go
jpeach: IIRC this is an unused variable warning. Perhaps I can just remove the parameter altogether.

pkg/cache/v3/delta_test.go
jpeach: Nope, it was fine :)

pkg/cache/v3/cache.go
jpeach: This would be `TTL` according to Go's [initialism convention](https://github.com/golang/go/wiki/CodeReviewComments#initialisms). I updated the internal names, but I didn't want to change the API in this PR.

internal/example/main/main.go
jpeach: Yup.

pkg/cache/v3/cache.go
alecholmez: Yup I agree. I honestly am okay with changing the API. If that breaks users they can pin, and we also haven't hit `v1.0` yet.

pkg/cache/v3/delta_test.go
alecholmez: Nice, that was causing issues for me a while ago but I guess it's fixed

pkg/cache/v3/simple.go
alecholmez: I'd remove the param

pkg/cache/v3/snapshot.go
alecholmez: I'm all for the rename. Lets make it consistent throughout

pkg/resource/v3/resource.go
alecholmez: Hm, we probably shouldn't ignore this error. Maybe we should alter the function signature and bubble up that err

alecholmez:
Just FYI I'm more in favor of adding extra error checks and changing things around if needed to support that. I'd rather not ignore things

jpeach:
@alecholmez  this dropped off my list a bit. Any objections to me rebasing while I address your review feedback?

pkg/cache/v3/snapshot.go
jpeach: AFAIK we can't add compatibility aliases since we need to rename the `Ttl` field in `ResourceWithTtl` and there's no way to alias that.

pkg/cache/v3/snapshot.go
jpeach: Removed this, since it is unused in this repository.

pkg/resource/v3/resource.go
jpeach: Updated to return nil which I this is OK for this API.

pkg/test/server.go
jpeach: Added a log to match the error on listen.

pkg/cache/v3/cache.go
jpeach: Ok, changing it. AFAIK we can't add a compatibility alias to a field so there's no way to rename without a breaking change.

pkg/cache/v3/mux.go
jpeach: OK, then I'll change this to take a pointer.

jpeach:
Addresses all the review feedback. Note that this includes incompatible changes, so users of some APIs will need to make minor changes when they upgrade.

alecholmez:
@jpeach sorry for this, can you rebase? I'd like to get this in before we attempt to cut a `v1` release. Want to make sure the code quality is up to par

jpeach:
> @jpeach sorry for this, can you rebase? I'd like to get this in before we attempt to cut a `v1` release. Want to make sure the code quality is up to par

Rebased

If you are planning a v1 release, I'd recommend you make a pass and s/Ttl/TTL/ in all the public names. I fixed some that linters picked up, but better to make it all consistent now (and in a separate PR).

alecholmez (approved):
Thanks for this! I will try and knock off some more lint issues as we move towards `v1`

PR #457, opened 2021-07-01
integration tests: add option to use pprof profiler

dougfort:
Signed-off-by: dougfort <doug.fort@greymatter.io>

This patch enables the use of the go language pprof profiler in integration tests.

### DIFF
--- a/pkg/test/main/README.md
+++ b/pkg/test/main/README.md
@@ -37,3 +37,16 @@ eventually converges to use the latest pushed configuration) for each run.
 
 You can run ```bin/test -help``` to get a list of the cli flags that
 the test program accepts.  There are also comments in ```main.go```.
+
+## Using the pprof profiler
+
+One customization is to run the go language profiler [pprof](https://github.com/DataDog/go-profiler-notes/blob/main/pprof.md). See also <https://golang.org/pkg/runtime/pprof/>.
+
+The profiler is normally off because it adds overhead to the tests. You can turn 
+it on with the command line option `--pprof`. There is an environment variable 
+`PPROF` for the `make` commands shown above. For example:
+
+    (export PPROF=true; make integration.xds)
+
+The test will then write files of the form `block_profile_xds.pb.gz`. 
+You can use `go tool pprof bin/test <file name>` to analyze the profile data. 
--- a/pkg/test/main/README.md
+++ b/pkg/test/main/README.md
@@ -48,5 +48,7 @@ it on with the command line option `--pprof`. There is an environment variable
 
     (export PPROF=true; make integration.xds)
 
-The test will then write files of the form `block_profile_xds.pb.gz`. 
+The test will then write files of the form `block_profile_xds.pb.gz`. The files
+get written to the root of the project, in the same place as the envoy logs.
+
 You can use `go tool pprof bin/test <file name>` to analyze the profile data. 
--- a/pkg/test/main/main.go
+++ b/pkg/test/main/main.go
@@ -24,6 +24,8 @@ import (
 	"log"
 	"net/http"
 	"os"
+	"runtime"
+	"runtime/pprof"
 	"time"
 
 	"github.com/envoyproxy/go-control-plane/pkg/cache/v3"
@@ -56,6 +58,8 @@ var (
 	mux           bool
 
 	nodeID string
+
+	pprof_enabled bool
 )
 
 func init() {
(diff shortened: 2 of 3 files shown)


pkg/test/main/README.md
```diff
@@ -37,3 +37,16 @@ eventually converges to use the latest pushed configuration) for each run.
...
+
+One customization is to run the go language profiler [pprof](https://github.com/DataDog/go-profiler-notes/blob/main/pprof.md). See also <https://golang.org/pkg/runtime/pprof/>.
+
+The profiler is normally off because it adds overhead to the tests. You can turn 
+it on with the command line option `--pprof`. There is an environment variable 
+`PPROF` for the `make` commands shown above. For example:
+
+    (export PPROF=true; make integration.xds)
+
+The test will then write files of the form `block_profile_xds.pb.gz`.
```
alecholmez, line 51: I'm guessing this just gets written to the root of the project?

pkg/test/main/README.md
```diff
@@ -37,3 +37,16 @@ eventually converges to use the latest pushed configuration) for each run.
...
+One customization is to run the go language profiler [pprof](https://github.com/DataDog/go-profiler-notes/blob/main/pprof.md). See also <https://golang.org/pkg/runtime/pprof/>.
+
+The profiler is normally off because it adds overhead to the tests. You can turn 
+it on with the command line option `--pprof`. There is an environment variable 
+`PPROF` for the `make` commands shown above. For example:
+
+    (export PPROF=true; make integration.xds)
+
+The test will then write files of the form `block_profile_xds.pb.gz`. 
+You can use `go tool pprof bin/test <file name>` to analyze the profile data.
```
alecholmez, line 52: Maybe we can have a make target so this is easy?

```make
make analyze
```

pkg/test/main/main.go
```diff
@@ -133,13 +137,40 @@ func init() {
 
 	// Enable a muxed cache with partial snapshots
 	flag.BoolVar(&mux, "mux", false, "Enable muxed linear cache for EDS")
+
+	//
+	// These parameters control the the use of the pprof profiler
+	//
+
+	// Enable use of the pprof profiler
+	flag.BoolVar(&pprof_enabled, "pprof", false, "Enable use of the pprof profiler")
```
alecholmez, line 146: I'd follow the previous naming patterns and do camelCase instead of snake case. 

```suggestion
flag.BoolVar(&pprofEnabled, "pprof", false, "Enable use of the pprof profiler")
```

pkg/test/main/main.go
```diff
@@ -243,6 +274,21 @@ func main() {
 		}
 	}
 
+	if pprof_enabled {
+
```
alecholmez, line 278: maybe remove this whitespace?

pkg/test/main/main.go
```diff
@@ -243,6 +274,21 @@ func main() {
...
+		for _, prof := range []string{"block", "goroutine", "mutex"} {
+			p := pprof.Lookup(prof)
+			filePath := fmt.Sprintf("%s_profile_%s.pb.gz", prof, mode)
+			log.Printf("storing %s profile for %s in %s", prof, mode, filePath)
+			f, err := os.Create(filePath)
```
alecholmez, line 283: Do you think it makes sense to have multiple files created out of this? I'm not entirely sure how pprof generates its artifacts but would it be possible to have everything written to one file?

alecholmez (changes requested):
Just some minor comments, I think this is a great start to the full e2e benchmark. Now we've got some sort of profiling data so we can start generating artifacts based on difference scenarios we create

pkg/test/main/main.go
dougfort: roger

pkg/test/main/README.md
dougfort: I don't want to mess with the makefile yet. 

Note that they have multiple targets:
```
integration.ads: $(BINDIR)/test $(BINDIR)/upstream
	env XDS=ads build/integration.sh

integration.xds: $(BINDIR)/test $(BINDIR)/upstream
	env XDS=xds build/integration.sh

etc
```
I agree that we need a more convenient way to run this.

pkg/test/main/main.go
dougfort: The [pprof](https://github.com/google/pprof) tool really wants this. If this gets used we might do some intermediate software to manage files.

pkg/test/main/README.md
dougfort: added some test to the README to make it clearer

dontlaugh:
Seems low-risk. Can we merge it? Unless there's some code style stuff we should be aware of.

alecholmez:
@snowp any thoughts on this as a first step?

pkg/test/main/main.go
```diff
@@ -133,13 +137,40 @@ func init() {
...
 // main returns code 1 if any of the batches failed to pass all requests
 func main() {
 	flag.Parse()
 	ctx := context.Background()
 
+	if pprofEnabled {
+		// turn on the block profiler
+		log.Println("turn on pprof block profiler")
+		runtime.SetBlockProfileRate(1)
+		if pprof.Lookup("block") == nil {
```
snowp, line 159: Can we use the same pattern here as below where we iterate over the profile types? I would also simplify the output here to just say something like `Turning on block, goroutine and mutex profiling`

snowp (approved):
Just one non-blocking hit, this LGTM

alecholmez:
Thanks for this @dougfort, just approved the CI process to run so once that passes I'll merge

alecholmez (approved):

PR #458, opened 2021-07-02
Fix bug in delta xds server

s-matyukevich:
Here is the description of the request flow that illustrates the bug
1. Envoy sends a DeltaServiceDiscoveryRequest and subscribes for an **existing** resource.
2. [here](https://github.com/envoyproxy/go-control-plane/blob/main/pkg/server/delta/v3/server.go#L160) we check that there is no registered delta watch yet and we call `cache.CreateDeltaWatch`
3. `CreateDeltaWatch` function checks that resource already exists and responds immediately. [delayedResponse](https://github.com/envoyproxy/go-control-plane/blob/main/pkg/cache/v3/simple.go#L427) is false so the new watch isn't created.
4. Envoy receives the response and responds with another DeltaServiceDiscoveryRequest with the right nonce set.
5. This time, when the server receives the request there will be an object in the `deltaWatches` map [here](https://github.com/envoyproxy/go-control-plane/blob/main/pkg/server/delta/v3/server.go#L159) So the if statement below won't be executed, and `cache.CreateDeltaWatch` won't be called for the second time.
6. We end up in a situation when there is no watcher registered in the cache so envoy won't receive any updates anymore.

You can check this by simply running integration tests. Before this fix integration tests don't fail, but from the output, you can see that no updates are being sent to the client after the test runner prints`update snapshot` message. 

P.S: I think [this](https://github.com/envoyproxy/go-control-plane/blob/main/pkg/server/delta/v3/server.go#L180-L199) function always puts a single object into `watches.deltaMuxedResponses` channel (because the cache always closes the watcher after sending a single response) so maybe, instead terminating this function by writing into `terminate ` channels, we can simply get rid of the for loop and terminate the function after the first response is forwarded to the `deltaMuxedResponses ` channel?


alecholmez:
@s-matyukevich thanks for opening this bug fix! This looks good, you can ignore the nonce set in delta right now because it's not really used in envoy. But this definitely addresses the issues opened earlier with delta failing to respond.

With regards to the termination channel, which `for` loop are you referring to? That termination go routine doesn't currently have one. Also note in this branch we now do this: https://github.com/envoyproxy/go-control-plane/blob/75140082ac745dc9f996e7bf2e4e6f2665ceea01/pkg/server/delta/v3/server.go#L177-L183

Since we linearize the responses coming through

s-matyukevich:
Thanks for the quick review @alecholmez 

> With regards to the termination channel, which for loop are you referring to? That termination go routine doesn't currently have one. Also note in this branch we now do this:

The way how you handle this in the branch is exactly what I meant.

s-matyukevich:
Another problem that I see is that neither unit tests nor integration tests capture this. In our own control plane, we use "integration style" unit tests: we run the control plane server using `google.golang.org/grpc/test/bufconn` then in the test we create and send different DeltaDiscoveryRequests to the server and verify the responses. I personally find those tests very useful because they allow testing the whole control plane (server + cache) but at the same time they are implemented as normal unit tests, so I can debug them + it is much easier to add new tests and assertions than in the normal integration tests, which rely on envoy and docker. 

I can copy come of our test to this repo if you like. They actually captured this bug.

alecholmez:
@s-matyukevich can you fix the conflicts? I'd like to get this in

s-matyukevich:
> @s-matyukevich can you fix the conflicts? I'd like to get this in

Sure, done!

pkg/server/delta/v3/server.go
```diff
@@ -162,25 +162,25 @@ func (s *server) processDelta(str stream.DeltaStream, reqCh <-chan *discovery.De
...
+			}
 
-				s.subscribe(req.GetResourceNamesSubscribe(), watch.state.ResourceVersions)
-				s.unsubscribe(req.GetResourceNamesUnsubscribe(), watch.state.ResourceVersions)
+			s.subscribe(req.GetResourceNamesSubscribe(), watch.state.ResourceVersions)
+			s.unsubscribe(req.GetResourceNamesUnsubscribe(), watch.state.ResourceVersions)
 
-				watch.cancel = s.cache.CreateDeltaWatch(req, watch.state, watch.responses)
-				watches.deltaWatches[typeURL] = watch
+			watch.responses = make(chan cache.DeltaResponse, 1)
```
s-matyukevich, line 174: One thining I am not completely certain is whether we should set `watch.responses` here or in the if above (after we set `watch.state`)

s-matyukevich: I think this only matters in the case when in a single stream we receive a request for a particular type while a watch for the same type is open. My understanding is that this should never happen because envoy always waits for the response to the previous request before sending another request. 

However, if I am wrong, we should probably fix the following scenario:
1. The control plane receives a request for some typeUrl T, there are no open watchers for this type, so we create new state and open a watcher [here](https://github.com/s-matyukevich/go-control-plane/blob/fix-delta-watchers/pkg/server/delta/v3/server.go#L175)
2. The control plane doesn't respond immediately.
3. The control plane receives one more request for the same type T.
4. We cancel the existing watch [here](https://github.com/s-matyukevich/go-control-plane/blob/fix-delta-watchers/pkg/server/delta/v3/server.go#L168) fun [this](https://github.com/s-matyukevich/go-control-plane/blob/fix-delta-watchers/pkg/server/delta/v3/server.go#L178-L183) goroutine is still running because, as far as I see from [this](https://github.com/s-matyukevich/go-control-plane/blob/fix-delta-watchers/pkg/cache/v3/simple.go#L447-L453) code, canceling a delta watch doesn't result in anything being written to the `watch.responses` channel.
5. We are left with an orphaned goroutine that is going to run forever.

alecholmez: Well delta might function differently, I think it's no longer a 1 -> 1 with the request cycle anymore. I think one of the bugs we were fixing was canceling the watch outside of the if. That will fix the case you have but we do need a way to clean up the go routine.

s-matyukevich: @alecholmez  I think the right fix would be to close the responses channel [here](https://github.com/envoyproxy/go-control-plane/blob/main/pkg/cache/v3/simple.go#L447) and make sure we do the same in the linear cache implementation. This should clean up the goroutine. Do you agree? Should I add this additional fix to the existing PR?

alecholmez: Well the responses channel is now propagated from the server since we changed the API for the cache. Should we just close it [here](https://github.com/envoyproxy/go-control-plane/blob/main/pkg/server/delta/v3/watches.go#L27) instead?

s-matyukevich: Yep, that's even better. I'll update the code.

alecholmez: Oops sorry I meant here: https://github.com/envoyproxy/go-control-plane/blob/main/pkg/server/delta/v3/watches.go#L45

@s-matyukevich

pkg/server/delta/v3/server.go
```diff
@@ -162,25 +162,25 @@ func (s *server) processDelta(str stream.DeltaStream, reqCh <-chan *discovery.De
 				// Since there was no previous state, we know we're handling the first request of this type
 				// so we set the initial resource versions if we have any, and also signal if this stream is in wildcard mode.
 				watch.state = stream.NewStreamState(len(req.GetResourceNamesSubscribe()) == 0, req.GetInitialResourceVersions())
-				watch.responses = make(chan cache.DeltaResponse, 1)
-
+			} else {
 				// We must signal goroutine termination to prevent a race between the cancel closing the watch
```
alecholmez, line 166: I think we can actually do away with this comment as well. Or maybe note that we now close the responses channel when canceling an old watch.

alecholmez (changes requested):
I'll approve once we get that channel close in

pkg/server/delta/v3/server.go
s-matyukevich: ok, done

pkg/server/delta/v3/server.go
s-matyukevich: I deleted the comment.

pkg/server/delta/v3/server.go
alecholmez: Great thanks!

alecholmez:
This looks great thanks for working this!

alecholmez (approved):

PR #459, opened 2021-07-02
incremental: support for linear and mux caches

s-matyukevich:


pkg/cache/v3/delta.go
```diff
@@ -20,8 +20,16 @@ import (
 	"github.com/envoyproxy/go-control-plane/pkg/server/stream/v3"
 )
 
+type ResourceContainer interface {
+	GetResources(typeURL string) map[string]types.Resource
+
+	GetVersionMap() map[string]map[string]string
+
+	GetVersion(typeURL string) string
+}
```
s-matyukevich, line 29: Here is the only tricky thing about this PR. Instead of passing Snapshot to the `respondDelta ` function we now pass the instance of the `ResourceContainer`, which snapshot implements. This allows to easily reuse the function for linear cache. Originally I was planning to make the linear cache to implement the interface, but this doesn't work well (it doesn't make sense for the cache methods to accept typeURL, also GetResources method should return `types.Resource` instead of `map[string]types.Resource`) So I end up adding a wrapper class `cacheWrapper `, which adopts linear cache to the ResourceContainer interface.

pkg/cache/v3/linear.go
```diff
@@ -42,13 +44,42 @@ type LinearCache struct {
...
 	versionVector map[string]uint64
-	mu            sync.RWMutex
+
+	log log.Logger
+
+	mu sync.RWMutex
+}
+
+// This struct exist only because we want to reuse respondDelta function
+type cacheWrapper struct {
```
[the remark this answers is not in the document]
alecholmez: I'm just curious as to what you think @s-matyukevich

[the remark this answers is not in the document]
s-matyukevich: > We lose things like Update, Set, and Delete 

Not sure I follow this, why do we lose them? They are not present in the wrapper interface, but this interface is internal and should never be exposed. We still have them in the linear cache interface, and their implementation relies on cacheWrapper.

> it creates a rather large divergence with the linear SOTW implementation

Agree. But the difference in my opinion is that delta has a lot more things in common between simple and linear caches (It includes how we treat wildcard, calculate hashes, prepare RemovedResources list, and so on. Basically - all stuff we have in the respondDelta function)

>  How difficult would it be if we were to write linear from scratch with delta? I'd be willing to help, but it might require some heavy lifting. I'm inclined to keep the snapshot/resource wrapper stuff separated with the simple cache

I don't think it will be difficult and I considered this as an option. We could get rid of the cacheWrapper interface, copy respondDelta function to the linear.go file and modify it slightly. The only problem with this approach is that duplicates some logic from the respondDelta function (and in my opinion, this logic is similar/same for both delta and linear) If we duplicate this logic it might be harder to support in the future. If we decide to reuse this logic we must change the respondDelta function signature, as it currently accepts Snapshot, which is specific to the simple cache implementation (and that is exactly what I attempted to do)

s-matyukevich:
@alecholmez I did some refactoring in the `delta.go` file and created 2 separate functions: `respondDeltaSnapshot` and `respondDeltaLinear`, while still reusing all code from `createDeltaResponse` Let me know what do you think, maybe it addresses some of your concerns.

pkg/cache/v3/delta.go
alecholmez: @s-matyukevich would it be useful to you if I were to add something like:
```go
// List returns a list of resources not indexed by name.
// This is useful when dealing with the linear cache implementation.
func (r Resources) List() []types.Resource {
	resources := []types.Resource{}
	for _, item := range r.Items {
		resources = append(resources, item.Resource)
	}
	return resources
}
```
in the new snapshot API? So now we have a slice when dealing with `types.Resources` instead of having to pass around the map in the linear cache? I think if I understand this right you'd be able to remove the cacheWrapper that way? I guess my goal here is to reduce the amount of boilerplate needed to get the linear cache to work. 

What could we do to the API to remove the need of the `resourceContainer`? If we're going to `1.0` here now is the time to design an API that will stand the test of other cache implementations

pkg/cache/v3/delta.go
```diff
@@ -22,15 +22,35 @@ import (
...
+	GetResources(typeURL string) map[string]types.Resource
+
+	GetVersionMap(typeURL string) map[string]string
+
+	GetVersion(typeURL string) string
+}
+
+// This struct exist only because we want to reuse createDeltaResponse function from linear cache code
+// so we create a wrapper, which adopts LinearCche to satisfy ResourceContainer interface
+type cacheWrapper struct {
```
alecholmez, line 35: What can we do to get rid of this?

pkg/cache/v3/linear.go
```diff
@@ -277,9 +308,71 @@ func (cache *LinearCache) CreateWatch(request *Request, value chan Response) fun
 }
 
 func (cache *LinearCache) CreateDeltaWatch(request *DeltaRequest, state stream.StreamState, value chan DeltaResponse) func() {
+	cache.mu.Lock()
+	defer cache.mu.Unlock()
+
+	response := respondDeltaLinear(request, value, state, cache, cache.log)
```
alecholmez, line 314: Nice, maybe it makes a bit more sense to move that respondDeltaLinear function into this life? What do you think

alecholmez (commented):
I had a few questions here, if we're moving to fix up the snapshot API and revisit some of the old stuff, what can we do here to require the least amount of boilerplate to be added

pkg/cache/v3/delta.go
s-matyukevich: I think we have 2 options:
1. Make Linear cache to implement the following methods
```
	GetResources(typeURL string) map[string]types.Resource
	GetVersionMap(typeURL string) map[string]string
	GetVersion(typeURL string) string
```
The only problem with this is that typeURL parameter is redundant for linear cache.
2. Change `createDeltaResponse` signature to something like this.
```
func createDeltaResponse(ctx context.Context, req *DeltaRequest, state stream.StreamState, resources map[string]types.Resource, versionMap map[string]string, systemVersion string, log log.Logger) *RawDeltaResponse
```
then in the `respondDeltaLinear ` and `respondDeltaSnapshot` function get the needed parameters either from snapshot or from the cache. 

I think both options are ok.

pkg/cache/v3/delta.go
s-matyukevich: Adding just this method won't solve the problem as we need also methods to get the version map and system version. Basically, we need something that implements the full interfaces https://github.com/envoyproxy/go-control-plane/pull/459/files#diff-a54688e00c731dac28aa2b2faab04dbd4dffd0aaac139cf0abf98b55efb31963R25-R31

pkg/cache/v3/linear.go
s-matyukevich: Yep, I can move both `respondDeltaLinear` and `respondDeltaSnapshot` to the corresponding files after we agree on remaining implementation details.

pkg/cache/v3/delta.go
s-matyukevich: We only use it [here](https://github.com/s-matyukevich/go-control-plane/blob/incremental-linear/pkg/cache/v3/delta.go#L134) and yes, it is used only for debugging purposes (we only need it to send it with the DeltaDiscoveryResponse to envoy)

alecholmez: > We only use it here and yes, it is used only for debugging purposes (we only need it to send it with the DeltaDiscoveryResponse to envoy)

Okay great I just wanted to make sure we didn't do anything special in linear

alecholmez: I'm sort of inclined to have the Linear Cache implement the three methods above so we can remove the cache wrapper, and then just ignore the type url params

s-matyukevich: > You mean they just won't be used right since the linear cache is instantiated on a per resource type basis

Yes, exactly

> I'm sort of inclined to have the Linear Cache implement the three methods above so we can remove the cache wrapper, and then just ignore the type url params. Do you think that would reduce the boilerplate?

ok, I can do that.

s-matyukevich: Actually this caused problems: linear cache already has method `GetResources`, which doesn't accept typeUrl parameter. Of cause, we can add another method `GetResourcesDelte(typeUrl string)` but I don't think this is acceptable for a public facing API. 

I actually implemented a little bit different solution, which still allows us to get rid of the cacheWrapper. Let me know what do you think.

pkg/cache/v3/linear.go
```diff
@@ -131,6 +153,35 @@ func (cache *LinearCache) notifyAll(modified map[string]struct{}) {
...
+		}
+	}
+}
+
+func (cache *LinearCache) respondDelta(request *DeltaRequest, value chan DeltaResponse, state stream.StreamState) *RawDeltaResponse {
+	resources := &resourceContainer{
+		resourceMap:   cache.resources,
+		versionMap:    cache.versionMap,
+		systemVersion: cache.getVersion(),
+	}
```
s-matyukevich, line 172: @alecholmez in order to get rid of the cacheWrapper, we now initialize resourceContainer struct separately for the simple and linear caches. And we still reuse createDeltaResponse function, which now take resourceContainer struct as an argument.

pkg/cache/v3/delta.go
```diff
@@ -22,50 +22,27 @@ import (
...
 	toRemove := make([]string, 0)
 
 	// If we are handling a wildcard request, we want to respond with all resources
 	switch {
 	case state.IsWildcard():
-		for name, r := range resources {
+		for name, r := range resources.resourceMap {
 			// Since we've already precomputed the version hashes of the new snapshot,
 			// we can just set it here to be used for comparison later
-			version := snapshot.GetVersionMap()[req.TypeUrl][name]
```
alecholmez, line 68: I'm guessing the resource container is going to be created on a per type basis now? Since we don't key off the type anymore from the snapshot

pkg/cache/v3/snapshot.go
```diff
@@ -223,8 +226,8 @@ func (s *Snapshot) GetVersion(typeURL string) string {
 }
 
 // GetVersionMap will return the internal version map of the currently applied snapshot.
-func (s *Snapshot) GetVersionMap() map[string]map[string]string {
-	return s.VersionMap
+func (s *Snapshot) GetVersionMap(typeUrl string) map[string]string {
+	return s.VersionMap[typeUrl]
```
alecholmez, line 230: Ah I see now, ignore my previous comment about scoping the resource container

pkg/cache/v3/snapshot.go
```diff
@@ -266,3 +269,30 @@ func (s *Snapshot) ConstructVersionMap() error {
 
 	return nil
 }
+
+// Respond to a delta watch with the provided snapshot value. If the response is nil, there has been no state change.
+func (s *Snapshot) respondDelta(ctx context.Context, request *DeltaRequest, value chan DeltaResponse, state stream.StreamState, log log.Logger) (*RawDeltaResponse, error) {
```
alecholmez, line 274: I'm cool with the code duplication. They're different cache implementations so it makes sense for them to be localized anyways. It seems a bit easier to understand instead of trying to tie the two together

pkg/cache/v3/linear.go
```diff
@@ -42,13 +44,25 @@ type LinearCache struct {
 	watches map[string]watches
 	// Set of watches for all resources in the collection
 	watchAll watches
+	// Set of delta watches. A delta watch always contain the list of subscribed resources
+	// together with its current version
+	// version and versionPrefix fields are ignored for delta watches, because we always generate the resource version.
+	deltaWatches map[int64]DeltaResponseWatch
+	// Continously incremented counter used to index delta watches.
+	deltaWatchCount int64
```
alecholmez, line 52: Do we want a watchCount for linear instead of a set of watches like sotw? I'm guessing it would be too difficult to keep that style of caching mechanism

pkg/cache/v3/linear.go
alecholmez: That sounds good to me

alecholmez (changes requested):
Just a few small things/nits. This is definitely starting to look good

pkg/cache/v3/delta.go
alecholmez: I like it! I think we can roll with what we've got

pkg/cache/v3/linear.go
s-matyukevich: I took a closer look at the corresponding sotw implementation. So we use [map[string]map[chan Response]struct{}](https://github.com/s-matyukevich/go-control-plane/blob/incremental-linear/pkg/cache/v3/linear.go#L44) to store watches in sotw vs in the delta variant we use [map[int64]DeltaResponseWatch](https://github.com/s-matyukevich/go-control-plane/blob/incremental-linear/pkg/cache/v3/linear.go#L50) As far as I understand, we need to group watchers by resource name in the sotw implementation because [here](https://github.com/s-matyukevich/go-control-plane/blob/incremental-linear/pkg/cache/v3/linear.go#L143-L148) we use resource names to track what resources the watcher is subscribed to. We don't need to do this in the delta varian because we track what resources we subscribed to using information from the stream state [here](https://github.com/s-matyukevich/go-control-plane/blob/incremental-linear/pkg/cache/v3/delta.go#L54) That's why we don't need to additionally group by resource name in the delta implementation.

Another difference is that sotw implementation uses  `chan Response]` as the key in the map and delta uses int64, and this is the reason why we have `deltaWatchCount int64` field, so we can generate unique watch id for every new watch. We can use the same approach as sotw and use `chan DeltaResponse` as the map key or even `DeltaResponseWatch` object. I am not sure it worth it though, because the current implementation doesn't add a lot of complexity and additionally we can log every new watch id [here](https://github.com/s-matyukevich/go-control-plane/blob/incremental-linear/pkg/cache/v3/linear.go#L341), which could be useful.

alecholmez: Okay cool, I was just curious on if it was even worth trying to follow the sotw linear implementation delta but it sounds like its not. 

I didn't think about how we use the stream state in delta vs sotw so we don't have to track the resource name in the watches set.

pkg/cache/v3/delta.go
```diff
@@ -22,50 +22,27 @@ import (
...
-// Respond to a delta watch with the provided snapshot value. If the response is nil, there has been no state change.
-func respondDelta(ctx context.Context, request *DeltaRequest, value chan DeltaResponse, state stream.StreamState, snapshot Snapshot, log log.Logger) (*RawDeltaResponse, error) {
-	resp, err := createDeltaResponse(ctx, request, state, snapshot)
-	if err != nil {
-		if log != nil {
-			log.Errorf("Error creating delta response: %v", err)
-		}
-		return nil, nil
-	}
+type resourceContainer struct {
```
alecholmez, line 25: Can you format this like so? No need for the extra white space
```go
type resourceContainer struct {
	resourceMap map[string]types.Resource
	versionMap map[string]string
	systemVersion string
}
```
Also might be worth adding a doc string on this object so people understand why we created this

pkg/cache/v3/snapshot.go
```diff
@@ -266,3 +269,30 @@ func (s *Snapshot) ConstructVersionMap() error {
...
+	resources := &resourceContainer{
+		resourceMap:   s.GetResources(request.TypeUrl),
+		versionMap:    s.GetVersionMap(request.TypeUrl),
+		systemVersion: s.GetVersion(request.TypeUrl),
+	}
+	resp := createDeltaResponse(ctx, request, state, resources, log)
```
alecholmez, line 280: I think now that we removed the logger from `createDeltaResponse` we need to remove it here too right?

pkg/cache/v3/snapshot.go
```diff
@@ -266,3 +269,30 @@ func (s *Snapshot) ConstructVersionMap() error {
 
 	return nil
 }
+
+// Respond to a delta watch with the provided snapshot value. If the response is nil, there has been no state change.
+func (s *Snapshot) respondDelta(ctx context.Context, request *DeltaRequest, value chan DeltaResponse, state stream.StreamState, log log.Logger) (*RawDeltaResponse, error) {
+	resources := &resourceContainer{
```
alecholmez, line 275: Does this need to be a pointer? Maps and slices are pointers under the hood anyways so this might be redundant

pkg/cache/v3/snapshot.go
alecholmez: Could maybe do something like:
```go
resp := createDeltaResponse(ctx, request, state, resourceContainer{
    resourceMap:   s.GetResources(request.TypeUrl),
    versionMap:    s.GetVersionMap(request.TypeUrl),
    systemVersion: s.GetVersion(request.TypeUrl),
})

File: pkg/cache/v3/snapshot.go (hunk 8 above)
Comment (alecholmez) [inline pkg/cache/v3/snapshot.go:274]: Sorry I missed this, can we actually move this func to `simple.go`? It's the simple cache delta response mechanism and even though it uses the snapshot API maybe it makes sense to not make this a method on a Snapshot? SOTW has their `respond()` function separate and localized to the cache

File: pkg/cache/v3/linear.go [hunk 13]
```diff
@@ -277,9 +328,71 @@ func (cache *LinearCache) CreateWatch(request *Request, value chan Response) fun
 }
 
 func (cache *LinearCache) CreateDeltaWatch(request *DeltaRequest, state stream.StreamState, value chan DeltaResponse) func() {
+	cache.mu.Lock()
+	defer cache.mu.Unlock()
+
+	response := cache.respondDelta(request, value, state)
+
+	// if respondDelta returns nil this means that there is no change in any resource version from the previous snapshot
```
alecholmez, line 336: Maybe don't say snapshot?

pkg/cache/v3/linear.go
```diff
@@ -277,9 +328,71 @@ func (cache *LinearCache) CreateWatch(request *Request, value chan Response) fun
...
+
+		return cache.cancelDeltaWatch(watchID)
+	}
+
 	return nil
 }
 
+func (cache *LinearCache) updateVersionMap(modified map[string]struct{}) error {
+	for name, r := range cache.resources {
+		//skip recalculating hash for the resoces that weren't modified
```
alecholmez, line 355: `// skip`

alecholmez (changes requested):
Thanks! A few more comments

pkg/cache/v3/delta.go
s-matyukevich: done

pkg/cache/v3/snapshot.go
s-matyukevich: done

pkg/cache/v3/snapshot.go
s-matyukevich: yep, removed pointer

pkg/cache/v3/snapshot.go
s-matyukevich: ok, moved

pkg/cache/v3/linear.go
s-matyukevich: done

pkg/cache/v3/linear.go
s-matyukevich: fixed

pkg/cache/v3/linear.go
```diff
@@ -131,6 +153,34 @@ func (cache *LinearCache) notifyAll(modified map[string]struct{}) {
...
+	resp := createDeltaResponse(context.Background(), request, state, resourceContainer{
+		resourceMap:   cache.resources,
+		versionMap:    cache.versionMap,
+		systemVersion: cache.getVersion(),
+	})
+
+	// Only send a response if there were changes
+	if len(resp.Resources) > 0 || len(resp.RemovedResources) > 0 {
+		if cache.log != nil {
+			cache.log.Debugf("node: %s, sending delta response with resources: %v removed resources %v wildcard: %t",
```
alecholmez, line 177: Maybe we should denote this log statements are coming from a linear cache? They look really similar to the simple cache statements

pkg/cache/v3/linear.go
```diff
@@ -277,9 +327,71 @@ func (cache *LinearCache) CreateWatch(request *Request, value chan Response) fun
...
+	defer cache.mu.Unlock()
+
+	response := cache.respondDelta(request, value, state)
+
+	// if respondDelta returns nil this means that there is no change in any resource version
+	// create a new watch accordingly
+	if response == nil {
+		watchID := cache.nextDeltaWatchID()
+		if cache.log != nil {
+			cache.log.Infof("open delta watch ID:%d for %s Resources:%v, system version %q", watchID,
```
alecholmez, line 340: Same thing here we should maybe note this is a watch from linear

pkg/cache/v3/simple.go
```diff
@@ -456,6 +455,32 @@ func (cache *snapshotCache) CreateDeltaWatch(request *DeltaRequest, state stream
...
+		versionMap:    snapshot.GetVersionMap(request.TypeUrl),
+		systemVersion: snapshot.GetVersion(request.TypeUrl),
+	})
+
+	// Only send a response if there were changes
+	// We want to respond immediately for the first wildcard request in a stream, even if the response is empty
+	// otherwise, envoy won't complete initialization
+	if len(resp.Resources) > 0 || len(resp.RemovedResources) > 0 || (state.IsWildcard() && state.IsFirst()) {
+		if cache.log != nil {
+			cache.log.Debugf("node: %s, sending delta response with resources: %v removed resources %v wildcard: %t",
```
alecholmez, line 471: ditto here

alecholmez (changes requested):
Just some log nits, then I think we can merge this

pkg/cache/v3/linear.go
s-matyukevich: I added `[linear cache]` prefix to all logging statements in the linear cache, so we no longer have identical messages from different components. I didn't touch any messages from `simple.go` (I don't want to modify only delta related log statements because I think we should add tags to log messages consistently). 

I believe this problem could be fixed in a generic ways similar to how all popular logging frameworks do that, here is one sample: https://pkg.go.dev/go.uber.org/zap#Logger.With We can use a different logger (which is a breaking change) or create custom functionality to support child loggers with custom tags, which are attached to every message automatically, without changing logger interface. Anyway, I think this is out of scope for this PR.

alecholmez: Ah awesome thank you. I have a PR up that I could probably introduce that into. It's probably worth adding tags to logs

alecholmez (approved):
This looks good! Thanks for the time and effort on this

PR #461, opened 2021-07-06
Draft: SOTW variants

fxposter:
This is basically a continuation of https://github.com/envoyproxy/go-control-plane/pull/357, which instead of replacing current `sotw.Server#process` implementation - adds 3 more implementations and ability to switch between them.

After https://github.com/envoyproxy/go-control-plane/pull/443 is merged - go-control-plane can support my original goals of being able not to have traffic drops when you have multiple resource types changed in a single "step", ie: if you change both CDS and EDS resource at the same time the only correct way to send these resources down the line is to send CDS first and EDS second (https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#eventual-consistency-considerations).

Currently the user of this library cannot do this at all, cause if they pushe both CDS and EDS responses into their channels on  the cache side in order - the select loop here (https://github.com/envoyproxy/go-control-plane/blob/main/pkg/server/sotw/v3/server.go#L195) does not guarantee that CDS will be selected first and EDS next.

Now, to actual implementations:
1. `server_default.go` - it is a copy of original `sotw.Server#process` implementation, to be able to "fall back to good state"
2. `server_xds.go` - works in the same way as the default one (responses channel is recreated on every valid incoming request), but only supports single resource type, which covers non-ADS protocol part. it is simpler than default one and (probably) a bit faster, given the fact that it just removes a bunch of stuff that non-ADS requests do not need.
3. `server_ordered_ads.go` - completely different implementation for ADS, that instead of creating a separate channel for each incoming request and abandoning the old one - uses (and reuses) single channel for all request types. the main benefit is that you no longer have a select that breaks ordering, so that if cache pushes CDS first and EDS second - this implementation guarantees that server will actually send updates to the wire in this order. Downside (as was discussed in previous PRs) - you no longer have backpressure per resource. ie: in default implementation if cache decides to send 2 CDS updates - the second one will be blocked, cause the channel has buffer of size 1 and second push to the channel will be blocked, but pushes to channels of other type urls can succeed as long as corresponding channels are empty. This implementation due to channel reuse and the fact that the channel is shared between type urls - if your cache implementation sends many EDS updates, for example - it might be blocked on channel push when there is a CDS update.
4. `server_unordered_ads.go` - reimplementation of the default behaviour, but based on reflect.Select, which eliminates duplication in original implementation and allows having any number of type urls supported without the need to fall back to either shared channel for non-LDS/RDS/CDS/EDS and does not require any additional goroutines if there is a need for same "backpressure semantics" for these non-main type urls.

### DIFF
--- a/pkg/server/sotw/v3/server.go
+++ b/pkg/server/sotw/v3/server.go
@@ -143,14 +143,47 @@ func (values *watches) Cancel() {
 	}
 }
 
-// process handles a bi-di stream request
-func (s *server) process(stream Stream, reqCh <-chan *discovery.DiscoveryRequest, defaultTypeURL string) error {
-	// increment stream count
-	streamID := atomic.AddInt64(&s.streamCount, 1)
+type streamWrapper struct {
+	stream    Stream
+	id        int64
+	callbacks Callbacks
 
 	// unique nonce generator for req-resp pairs per xDS stream; the server
 	// ignores stale nonces. nonce is only modified within send() function.
-	var streamNonce int64
+	nonce int64
+}
+
+func (sw *streamWrapper) nextNonce() string {
+	// increment nonce
+	sw.nonce += 1
+	return strconv.FormatInt(sw.nonce, 10)
+}
+
+// sends a response by serializing to protobuf Any
+func (sw *streamWrapper) send(resp cache.Response) (string, error) {
+	if resp == nil {
+		return "", errors.New("missing response")
+	}
+
+	out, err := resp.GetDiscoveryResponse()
+	if err != nil {
+		return "", err
+	}
+
+	out.Nonce = sw.nextNonce()
+	if sw.callbacks != nil {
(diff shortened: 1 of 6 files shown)


fxposter:
@snowp @alecholmez @kyessenov @jessicayuen

alecholmez:
@fxposter I'd like to shepherd this along to being merged in, would you be okay with splitting up each of the implementations into respective PRs? I think it'd be much easier to review

alecholmez:
Not stale

maplebeats:
[at pkg](https://github.com/envoyproxy/go-control-plane/blob/1bbf44f240d914329caf6a1bd76b863d16ee7a8a/pkg/cache/v3/simple.go#L202)
cache use a map store watchs and range it, it will push in random. we don't just  guarantee the order of the server , but also range watchs respond order.

Issue #465, opened 2021-07-23
Delta xDS doesn't respond with anything if the snapshot for a node is empty. This causes envoy to fail its healthcheck.


s-matyukevich:
We use a combination of static and dynamic resources for our envoys and sometimes we want envoy to start with an empty dynamic_resources list. This doesn't work with the current delta xDS implementation because [here](https://github.com/envoyproxy/go-control-plane/blob/main/pkg/cache/v3/delta.go#L34) we check whether there were any updates and if no, we just open a new watch and don't send DeltaDiscoveryResponse to the client. As I mentioned in the issue title, this causes envoy to stay in a non-ready state (and if you use `/ready` endpoint for the liveness probe in k8s, which is standard practice, your envoy pod will be terminated)

This could be easily fixed if we first check whether the [previous version](https://github.com/envoyproxy/go-control-plane/blob/main/pkg/cache/v3/delta.go#L61) is empty in the stream state and if yes - respond anyway, even if the resource list is empty. I can send a PR with the fix if you think this is a good idea.

cc @alecholmez

snowp:
I don't think this is something we'd want to enable by default as it's a pretty important feature for any system that relies on observing the incoming requests or pending watches in order to generate the initial configuration: the fact that no response is sent ensures that Envoy will block startup until the snapshot is set, which is preferable over Envoy coming up without resources in certain deployments.

I'd suggest setting an empty snapshot in this case, or if there is a need for more sophisticated default handling we can consider making that configurable

s-matyukevich:
> I'd suggest setting an empty snapshot

That's actually what we did. To clarify the point: I am not suggesting sending a response if no snapshot is set for the node, I want to send an empty response if an empty snapshot is used.

snowp:
Ah yea I misunderstood. I think this makes sense for wildcard mode (you subscribe for all resources but there are none) but not for regular subscription mode (you subscribed for resource X Y Z but none of them exist). I assume the resources blocking your startup is either clusters or listeners?

@htuch thoughts on when a Delta server should respond with an empty response?

s-matyukevich:
> I assume the resources blocking your startup is either clusters or listeners?

Exactly. Right now we create a dummy cluster as a workaround.

htuch:
@snowp I think it wouldn't be incorrect to have a `DeltaDiscoveryResponse` that had empty add/removes being used to trigger initialization if that's what's desired. CC @envoyproxy/api-shepherds for second opinion.

PR #473, opened 2021-07-29
Respond immediately for the first wildcard request in a delta stream, If the corresponding snapshot exists and the response is empty

s-matyukevich:
This PR fixes https://github.com/envoyproxy/go-control-plane/issues/465


pkg/cache/v3/delta.go
```diff
@@ -31,10 +31,13 @@ func respondDelta(request *DeltaRequest, value chan DeltaResponse, state stream.
 	}
 
 	// Only send a response if there were changes
-	if len(resp.Resources) > 0 || len(resp.RemovedResources) > 0 {
+	if len(resp.Resources) > 0 || len(resp.RemovedResources) > 0 ||
+		// we want to respond immediatly for the first wildcard request in a stream, even if the response is empty
+		// otherwise, envoy won't complete initialization
+		(state.IsWildcard() && state.IsFresh()) {
```
s-matyukevich, line 37: This is the only meaningful change in this PR. The rest is just refactoring: I added wrapper functions, such as `Get/SetResourceVersions` and `IsWildcard` to track whether the stream state "is fresh", which means that the state corresponds to the first request in a stream.

pkg/cache/v3/delta.go
```diff
@@ -31,10 +31,13 @@ func respondDelta(request *DeltaRequest, value chan DeltaResponse, state stream.
 	}
 
 	// Only send a response if there were changes
-	if len(resp.Resources) > 0 || len(resp.RemovedResources) > 0 {
+	if len(resp.Resources) > 0 || len(resp.RemovedResources) > 0 ||
+		// we want to respond immediatly for the first wildcard request in a stream, even if the response is empty
+		// otherwise, envoy won't complete initialization
```
alecholmez, line 36: Does this only apply for CDS/LDS since this is targeted for wildcard mode?

pkg/cache/v3/delta.go
alecholmez: Just a slight nit pick for this, can we move the `&&` block to the same line as the `if`? It's a bit hard to read like that. I'd just mention the comment is targeted at that section

alecholmez: So @snowp and I originally had a bool on that Watches object that was `IsFirst`, do we really need that? Can't we just determine if there are no open watches and there's no stream state we can just assume this is the first request? That's what we currently do

pkg/server/stream/v3/stream.go
```diff
@@ -24,22 +24,43 @@ type DeltaStream interface {
...
 	// ResourceVersions contains a hash of the resource as the value and the resource name as the key.
 	// This field stores the last state sent to the client.
-	ResourceVersions map[string]string
+	resourceVersions map[string]string
+
+	// indicates whether the object has beed modified since its creation
+	fresh bool
+}
+
+func (s *StreamState) GetResourceVersions() map[string]string {
```
alecholmez, line 37: Do we need these wrappers? I'd be all for it if we had a lock we needed to hide, but I think we don't since resourceVersions will only be modified by one thread at a time so it just seems like extra boilerplate.

pkg/cache/v3/delta_test.go
```diff
@@ -28,7 +28,7 @@ func TestSnapshotCacheDeltaWatch(t *testing.T) {
 			},
 			TypeUrl:                typ,
 			ResourceNamesSubscribe: names[typ],
-		}, stream.StreamState{ResourceVersions: nil, Wildcard: true}, watches[typ])
+		}, stream.NewStreamState(true, nil), watches[typ])
```
alecholmez, line 31: Ah thanks for this catch

alecholmez (changes requested):
Just a few questions, @snowp can you do a quick review too?

pkg/cache/v3/delta.go
s-matyukevich: Are you referring to [this](https://github.com/envoyproxy/go-control-plane/blob/main/pkg/server/delta/v3/server.go#L164) code? We can easily infer whether this is the first request in a stream from the server code, but we can't do the same in the cache code because we don't pass watches objects to the cache. Alternatively, we can pass  isFirst flag all the way from the server, starting with CreateDeltaWatch function. But I think that the stream state should aggregate all information about the stream state, so it probably makes sense to store this information there.

pkg/cache/v3/delta.go
s-matyukevich: Yes, I added the check for  wildcard based on this comment https://github.com/envoyproxy/go-control-plane/issues/465#issuecomment-885708998

pkg/server/stream/v3/stream.go
s-matyukevich: I would also like to avoid having the wrappers. I added the wrapper for the  `SetResourceVersions ` method because I also set isFresh flag to false inside this method. The assumption is that we change the initial ResourceVersions only after we got the second request. Alternatively, we could manually set isFresh flag from the server code and keep accessing StreamState fields directly - this should make this PR much smaller, but I am not sure this is a better design. Let me know what you think. 

One more option is to keep `SetResourceVersions` wrapper but delete all other wrappers, but I think we should be consistent in how we access struct fields.

pkg/cache/v3/delta.go
alecholmez: Okay great. Can we address the nit I had with the formatting?

s-matyukevich: Sure, done!

pkg/cache/v3/delta.go
alecholmez: That's fair i'm in agreement with that. I'd like the stream state to contain most of the knowledge about delta streams.

I had previously tried passing down an `IsFirst()` flag down into the cache but it was awkward. Can we change the name `IsFresh()`? That seems to read a bit awkward

s-matyukevich: Yeah, I like `IsFirst` better, renaming.

s-matyukevich: done

pkg/server/stream/v3/stream.go
alecholmez: Well from a straight performance perspective I think it would be faster to just access the fields directly. An exception applies to maps since those are always pointers but then we won't copy around bools and such. This is obviously minor so it might be worth trading off for readability. I'm good with leaving the wrappers for now

alecholmez (approved):

PR #474, opened 2021-07-29
add context.Context param to SetSnapshot

snowp:
This allows for setting a timeout on responding to watches, and allows us to pass additional information to OnStreamResponse, allowing for additional information about the response to be passed to the server callbacks.

### DIFF
--- a/pkg/cache/v3/simple.go
+++ b/pkg/cache/v3/simple.go
@@ -143,7 +143,7 @@ func NewSnapshotCacheWithHeartbeating(ctx context.Context, ads bool, hash NodeHa
 				cache.mu.Lock()
 				for node := range cache.status {
 					// TODO(snowp): Omit heartbeats if a real response has been sent recently.
-					cache.sendHeartbeats(node)
+					cache.sendHeartbeats(ctx, node)
 				}
 				cache.mu.Unlock()
 			case <-ctx.Done():
@@ -154,7 +154,7 @@ func NewSnapshotCacheWithHeartbeating(ctx context.Context, ads bool, hash NodeHa
 	return cache
 }
 
-func (cache *snapshotCache) sendHeartbeats(node string) {
+func (cache *snapshotCache) sendHeartbeats(ctx context.Context, node string) {
 	snapshot := cache.snapshots[node]
 	if info, ok := cache.status[node]; ok {
 		info.mu.Lock()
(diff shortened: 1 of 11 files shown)


snowp:
@alecholmez I added tests here to verify that resetting the same snapshot will complete the action. This works because we don't delete the watches unless we actually succeed in sending the response, leaving the old watches open until the next time the snapshot is set

alecholmez (approved):
Nice this looks great, just had one minor question

pkg/cache/v3/simple.go
```diff
@@ -178,7 +178,7 @@ func (cache *snapshotCache) sendHeartbeats(ctx context.Context, node string) {
 				cache.log.Debugf("respond open watch %d%v with heartbeat for version %q", id, watch.Request.ResourceNames, version)
 			}
 
-			cache.respond(watch.Request, watch.Response, resourcesWithTtl, version, true)
+			cache.respond(context.Background(), watch.Request, watch.Response, resourcesWithTtl, version, true)
```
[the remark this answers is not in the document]
snowp: The ctx that we passed here was the ctx given to NewSnapshotCacheWithHeartbeating, which was intended to allow proper shutdown of the heartbeating routine. I don't think it makes sense for this to be propagated to the server, wdyt?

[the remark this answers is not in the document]
snowp: Actually I retract that, it might be nice to let people add information to the context to indicate heartbeats, so ill fix that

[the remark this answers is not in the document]
alecholmez: Yea I was thinking it might be nice just for the flexibility

alecholmez (approved):
Nice this looks good, thanks for opening this

pkg/cache/v3/simple.go
```diff
@@ -419,7 +432,13 @@ func (cache *snapshotCache) CreateDeltaWatch(request *DeltaRequest, state stream
 			cache.log.Errorf("failed to compute version for snapshot resources inline, waiting for next snapshot update")
 			delayedResponse = true
 		}
-		delayedResponse = respondDelta(request, value, state, snapshot, cache.log) == nil
+		response, err := respondDelta(context.Background(), request, value, state, snapshot, cache.log)
+		if err != nil {
+			cache.log.Errorf("failed to respond with delta response, waiting for next snapshot update: %s", err)
```
s-matyukevich, line 437: We usually check for `cache.log == nil` in other places, this could result in a panic.

pkg/cache/v3/simple.go
```diff
@@ -419,7 +432,13 @@ func (cache *snapshotCache) CreateDeltaWatch(request *DeltaRequest, state stream
...
+			delayedResponse = true
```
s-matyukevich, line 438: I think this statement is useless because we override delayedResponse on line 441 unconditionally.

pkg/cache/v3/simple.go
snowp: Good call, let me address this when I get a chance (if nobody else gets to it, I'm about done for the week) and also add a test for a nil logger

alecholmez: I'll open this quickly, I missed that in the review sorry about that

pkg/cache/v3/simple.go
alecholmez: Oh I think you might be right. That's always going to execute regardless if we hit any of those errors

PR #476, opened 2021-07-30
benchmarks: profiling

alecholmez:


alecholmez:
Just FYI the large file numbers are because I check in the generated reports so we can track a diff of when things change and why

alecholmez:
This is the first step to running e2e benchmarks on this repo. I've integrated some profiling utilizing pprof so we can understand what the code is really doing and see where we can squeeze a bit more performance out of. The `README` in `./benchmarks` demonstrates how to run the benchmarks and generate the artifacts, but this is only the first step for the e2e stuff. I still need to create test clients that can push this repo to the limits rather than just using envoy and the integration tests.

PR #477, opened 2021-07-31
bug: fix possible panic if no cache logger is supplied

alecholmez:


snowp (approved):
Thanks, LG

Anything we can do to make this less risky in the future? Be that adding tests that cover this or improve the API?

alecholmez:
@snowp I think both honestly, It might be worth changing the API so if a logger isn't declared at start we can have a default, that way we don't need to check for nil logs everywhere

alecholmez:
sorry @snowp had to rebase, can you rereview?

snowp (approved):
LG. Are you planning on tackling the null logger or should I do that?

alecholmez:
I'll go ahead and do that, should be simple

PR #480, opened 2021-08-02
lint for CI

alecholmez:

### DIFF
--- a/.github/workflows/golangci-lint.yaml
+++ b/.github/workflows/golangci-lint.yaml
@@ -1,0 +1,25 @@
+name: golangci-lint
+
+on:
+  push:
+    tags:
+      - v*
+    branches:
+      - master
+      - main
+  pull_request:
+  
+jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v2
+        with:
+          version: latest
+          only-new-issues: true
+          
+          Optional: golangci-lint command line arguments.
+          args: --timeout=10m0s
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -6,7 +6,6 @@ linters:
   - bodyclose
   - gofmt
   - goimports
-  - golint
   - gosec
   - misspell
   - revive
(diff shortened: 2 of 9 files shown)


alecholmez:
This introduces a linter for CI utilizing the new beta Github Action annotations feature. The aim is to increase general code quality and make sure we maintain a high standard for contribution!

pkg/cache/v3/cache.go
```diff
@@ -21,15 +21,16 @@ import (
 	"fmt"
 	"sync/atomic"
 
-	discovery "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
+	"google.golang.org/protobuf/types/known/anypb"
 
 	"github.com/golang/protobuf/proto"
```
jpeach, line 26: If you are switching proto libraries, should you switch over fully everywhere?

alecholmez: Ah oops I just missed that one thank you

alecholmez: So for this specific case I wasn't able to switch over since the underlying proto files still use `github.com/golang/protobuf/proto`, and to do a message reflect you need to use the corresponding lib. I was at least able to move away from the deprecated func though

jpeach: Yeh, when you mix & match, you end up having to convert using [MessageV2](https://pkg.go.dev/github.com/golang/protobuf/proto#MessageV2). For the v1 release, I'd recommend making sure the public API is fully V1 or fully V2 :)

alecholmez: Ah okay, i'll revist this then

jpeach:
Ping @jyotimahapatra  @snowp. Do you have review cycles for this one?

jpeach:
@alecholmez needs a rebase?

alecholmez:
@jpeach just removed that last bit of `github.com/golang/protobuf` from the cache package. Can you look over this one more time?

@snowp can we merge this in before `v0.10.0` as well?

pkg/cache/v3/linear_test.go
```diff
@@ -34,7 +33,7 @@ const (
 )
 
 func testResource(s string) types.Resource {
-	return &wrappers.StringValue{Value: s}
+	return &wrapperspb.StringValue{Value: s}
```
jpeach, line 36:
```suggestion
	return wrapperspb.String(s)
```

.github/workflows/golangci-lint.yaml
```diff
@@ -0,0 +1,25 @@
+name: golangci-lint
+
+on:
+  push:
+    tags:
+      - v*
+    branches:
+      - master
+      - main
+  pull_request:
```
jpeach, line 10:
```suggestion
on: [push, pull_request]
```

Is there any reason this should not always always run?

.golangci.yml
```diff
@@ -22,16 +21,3 @@ linters-settings:
     check-exported: false
   goimports:
     local-prefixes: github.com/envoyproxy/go-control-plane
```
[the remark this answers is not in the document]
alecholmez: I thought about doing something like that in the `go.mod` file but there's still some generated proto stubs that require the old stuff so I want to make sure we still keep support for that. If this really ends up becoming a problem we can add this at a later time

alecholmez:
@snowp if you've got some time today to look this one over that'd be great

PR #483, opened 2021-08-03
add default logger

alecholmez:


alecholmez:
@snowp here's that default logger I promised, I pulled out the nil logger checks in the cache since we don't need them anymore

pkg/log/default.go
```diff
@@ -0,0 +1,41 @@
+package log
+
+import (
+	"go.uber.org/zap"
```
snowp, line 4: Any reason to bring in a zap dependency? I would probably just have this use `log`, `fmt` or something?

alecholmez: Ah I did this cause it maps nicely with the different levels and provides structured logs. I can move back to `fmt` if we want @snowp

alecholmez: Ah that's fair, I'll make the change then so we don't throw in some unexpected behavior where it's not needed

alecholmez:
@snowp should be good to go, default is just a noop that logs out nothing. This keeps current logging behavior

snowp (approved):
Thanks!

PR #484, opened 2021-08-04
cache: rework snapshot api

alecholmez:
reworks cache api so we no longer need to change function signatures when adding support for new xDS services

### DIFF
--- a/pkg/cache/v3/resources.go
+++ b/pkg/cache/v3/resources.go
@@ -38,8 +38,8 @@ func NewResources(version string, items []types.Resource) Resources {
 	return NewResourcesWithTTL(version, itemsWithTTL)
 }
 
-// NewResources creates a new resource group.
-func NewResourcesWithTTL(version string, items []types.ResourceWithTTL) Resources { // nolint:golint,revive
+// NewResourcesWithTTL creates a new resource group.
+func NewResourcesWithTTL(version string, items []types.ResourceWithTTL) Resources {
 	return Resources{
 		Version: version,
 		Items:   IndexResourcesByName(items),
--- a/pkg/cache/v3/resources.go
+++ b/pkg/cache/v3/resources.go
@@ -1,0 +1,47 @@
+package cache
+
+import "github.com/envoyproxy/go-control-plane/pkg/cache/types"
+
+// Resources is a versioned group of resources.
+type Resources struct {
+	// Version information.
+	Version string
+
+	// Items in the group indexed by name.
+	Items map[string]types.ResourceWithTTL
+}
+
+// IndexResourcesByName creates a map from the resource name to the resource.
+func IndexResourcesByName(items []types.ResourceWithTTL) map[string]types.ResourceWithTTL {
+	indexed := make(map[string]types.ResourceWithTTL)
+	for _, item := range items {
+		indexed[GetResourceName(item.Resource)] = item
+	}
+	return indexed
+}
+
+// IndexRawResourcesByName creates a map from the resource name to the resource.
+func IndexRawResourcesByName(items []types.Resource) map[string]types.Resource {
+	indexed := make(map[string]types.Resource)
+	for _, item := range items {
+		indexed[GetResourceName(item)] = item
+	}
+	return indexed
+}
+
+// NewResources creates a new resource group.
+func NewResources(version string, items []types.Resource) Resources {
+	itemsWithTTL := []types.ResourceWithTTL{}
+	for _, item := range items {
+		itemsWithTTL = append(itemsWithTTL, types.ResourceWithTTL{Resource: item})
+	}
+	return NewResourcesWithTTL(version, itemsWithTTL)
+}
+
+// NewResources creates a new resource group.
+func NewResourcesWithTTL(version string, items []types.ResourceWithTTL) Resources { // nolint:golint,revive
+	return Resources{
+		Version: version,
+		Items:   IndexResourcesByName(items),
+	}
+}
--- a/pkg/cache/v3/snapshot.go
+++ b/pkg/cache/v3/snapshot.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 
 	"github.com/envoyproxy/go-control-plane/pkg/cache/types"
+	"github.com/envoyproxy/go-control-plane/pkg/resource/v3"
 )
 
 // Snapshot is an internally consistent snapshot of xDS resources.
@@ -36,7 +37,7 @@ type Snapshot struct {
 
 // NewSnapshot creates a snapshot from response types and a version.
 // The resources map is keyed off the type URL of a resource, followed by the slice of resource objects.
-func NewSnapshot(version string, resources map[string][]types.Resource) Snapshot {
+func NewSnapshot(version string, resources map[resource.Type][]types.Resource) Snapshot {
 	out := Snapshot{}
 
 	for typ, resource := range resources {
@@ -51,7 +52,7 @@ func NewSnapshot(version string, resources map[string][]types.Resource) Snapshot
 
 // NewSnapshotWithTTLs creates a snapshot with time to live per resource.
 // The resources map is keyed off the type URL of a resource, followed by the slice of resource objects.
-func NewSnapshotWithTTLs(version string, resources map[string][]types.ResourceWithTTL) Snapshot {
+func NewSnapshotWithTTLs(version string, resources map[resource.Type][]types.ResourceWithTTL) Snapshot {
 	out := Snapshot{}
 
 	for typ, resource := range resources {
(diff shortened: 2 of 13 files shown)


pkg/cache/v3/snapshot.go
```diff
@@ -94,64 +35,30 @@ type Snapshot struct {
...
-	Routes           []types.Resource
-	Listeners        []types.Resource
-	Runtimes         []types.Resource
-	Secrets          []types.Resource
-	ExtensionConfigs []types.Resource
-}
-
-// NewSnapshotWithResources creates a snapshot from response types and a version.
-func NewSnapshotWithResources(version string, resources SnapshotResources) Snapshot {
+func NewSnapshot(version string, resources map[string][]types.Resource) Snapshot {
```
jpeach, line 38: So make it clear that this map is keyed on resource type, consider switching to a typedef. This would make it easier to keep straight which keys are resource type names, and which are resource object names.

```Go
type ResourceType string

const EndpointType = ResourceType(apiTypePrefix + "envoy.config.endpoint.v3.ClusterLoadAssignment")
```

alecholmez: Hm yea I like that. My only thought to not do something like that is to keep the generic ability of any type, not just what is defined in the resources file. But I think yea it's probably worth doing some sort of typedef and then users can just create their own custom types

alecholmez: I'll give it a shot and see how it goes

alecholmez: I think at the very least it's worth adding a doc string noting the key is the resource type

jpeach: Yeh if you make it a type, clients can still provide their own values of that type.

alecholmez: So this is kinda awkward since the type url in a `DiscoveryRequest` is of type `string`. If I were to require a type alias here we'd have to do a lot of type casting back and forth between string

jpeach: Maybe you can cast it once at the API boundary (and verify the type in the protobuf registry)?

alecholmez: That seemed to work out nicely @jpeach thanks :)

pkg/cache/v3/resources.go
```diff
@@ -0,0 +1,47 @@
...
+func NewResources(version string, items []types.Resource) Resources {
+	itemsWithTTL := []types.ResourceWithTTL{}
+	for _, item := range items {
+		itemsWithTTL = append(itemsWithTTL, types.ResourceWithTTL{Resource: item})
+	}
+	return NewResourcesWithTTL(version, itemsWithTTL)
+}
+
+// NewResources creates a new resource group.
+func NewResourcesWithTTL(version string, items []types.ResourceWithTTL) Resources { // nolint:golint,revive
```
jpeach, line 42: Nit: should be able to remove the `nolint`

pkg/cache/v3/resources.go
```diff
@@ -0,0 +1,47 @@
...
+// NewResources creates a new resource group.
+func NewResources(version string, items []types.Resource) Resources {
+	itemsWithTTL := []types.ResourceWithTTL{}
+	for _, item := range items {
+		itemsWithTTL = append(itemsWithTTL, types.ResourceWithTTL{Resource: item})
+	}
+	return NewResourcesWithTTL(version, itemsWithTTL)
+}
+
+// NewResources creates a new resource group.
```
jpeach, line 41:
```suggestion
// NewResourcesWithTTL creates a new resource group.
```

pkg/cache/v3/snapshot.go
```diff
@@ -199,7 +109,7 @@ func (s *Snapshot) GetResources(typeURL string) map[string]types.Resource {
 }
 
 // GetResourcesAndTtl selects snapshot resources by type, returning the map of resources and the associated TTL.
```
jpeach, line 111:
```suggestion
// GetResourcesAndTTL selects snapshot resources by type, returning the map of resources and the associated TTL.
```

pkg/resource/v3/resource.go
```diff
@@ -8,6 +8,8 @@ import (
 	hcm "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/network/http_connection_manager/v3"
 )
 
+type Type string
```
jpeach, line 11: Nit: add a godoc comment for this?

pkg/resource/v3/resource.go
```diff
@@ -8,6 +8,8 @@ import (
...
+
```
jpeach, line 12: Nit: WDYT about making all the FooType constants below be of type `Type`? Too intrusive?

pkg/cache/v3/snapshot.go
```diff
@@ -199,7 +109,7 @@ func (s *Snapshot) GetResources(typeURL string) map[string]types.Resource {
...
-func (s *Snapshot) GetResourcesAndTtl(typeURL string) map[string]types.ResourceWithTTL { // nolint:golint,revive
+func (s *Snapshot) GetResourcesAndTTL(typeURL string) map[string]types.ResourceWithTTL { // nolint:golint,revive
```
jpeach, line 112: Nit: remove the `nolint`?

jpeach: Should this, and the other APIs that take a typeURL argument use `resource.Type` for it?

pkg/resource/v3/resource.go
alecholmez: Yea unfortunately that was intrusive and what caused all the code changes originally.

pkg/cache/v3/snapshot.go
alecholmez: I just tried that and it seemed to also be too intrusive in the cache. It would require a lot of type casting again

alecholmez: Actually nevermind, just figured something out I will push a big change haha

alecholmez:
@jpeach ended up aliasing to string instead of creating a whole new type. Now a lot of the snapshot stuff asks for a `resource.Type` as a param

pkg/resource/v3/resource.go
```diff
@@ -8,7 +8,8 @@ import (
 	hcm "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/network/http_connection_manager/v3"
 )
 
-type Type string
+// Type is an alias to string which we expose to users of the snapshot API which accepts `resource.Type` resouurce URLs.
```
jpeach, line 11:
```suggestion
// Type is an alias to string which we expose to users of the snapshot API which accepts `resource.Type` resource URLs.
```

pkg/cache/v3/snapshot.go
```diff
@@ -94,64 +36,32 @@ type Snapshot struct {
...
 }
 
-func NewSnapshotWithTtls(version string,
-	endpoints []types.ResourceWithTTL,
-	clusters []types.ResourceWithTTL,
-	routes []types.ResourceWithTTL,
-	listeners []types.ResourceWithTTL,
-	runtimes []types.ResourceWithTTL,
-	secrets []types.ResourceWithTTL) Snapshot {
+// NewSnapshotWithTTLs creates a snapshot with time to live per resource.
```
snowp, line 53: I would just phrase this as `creates a snapshot of ResourceWithTTLs` or something, people can refer to ResourceWithTTL to understand what fields can be set

pkg/cache/v3/snapshot.go
```diff
@@ -94,64 +36,32 @@ type Snapshot struct {
...
-	out.Resources[types.Route] = NewResourcesWithTtl(version, routes)
-	out.Resources[types.Listener] = NewResourcesWithTtl(version, listeners)
-	out.Resources[types.Runtime] = NewResourcesWithTtl(version, runtimes)
-	out.Resources[types.Secret] = NewResourcesWithTtl(version, secrets)
+
+	for typ, resource := range resources {
+		index := GetResponseType(typ)
+		if index != types.UnknownType {
+			out.Resources[index] = NewResourcesWithTTL(version, resource)
+		}
```
snowp, line 62: If we hit this it seems like a programmer error, should we return an error?

snowp (commented):
Thanks this seems to be a great improvement, just some minor comments

pkg/cache/v3/snapshot.go
alecholmez: Cool that sounds good

pkg/cache/v3/snapshot.go
alecholmez: I think an error here would make sense. Right now it's a silent ignore which might be confusing to users

alecholmez:
@snowp just pushed that last change. `NewSnapshot` now returns an error

alecholmez:
@snowp can we merge this?

snowp (approved):
LGTM, thanks!

PR #486, opened 2021-08-11
Minor updates for the existing benchmark process

hiromis:
- `make docker_benchmarks` was throwing the following error, so added `mkdir` commands to the script as well as adding these folders to `.gitignore` and `make clean`:
```
+ pprof -text bin/test cpu.pprof
/go-control-plane/build/do_benchmarks.sh: line 25: benchmarks/reports/cpu_text.txt: No such file or directory
make: *** [docker_benchmarks] Error 1
```
- Added a paragraph to benchmarks/README.md about visualizing the results when running locally.
- Corrected default `MODE` value setting in `build/integration.sh`
- Removed redundant value setting (`MODE=$(MODE)`) from `make benchmark`


hiromis:
Sure. Updated.

alecholmez (approved):
Thanks this looks good

PR #487, opened 2021-08-17
Adding -sleep flag

hiromis:
Added `-sleep` flag to test driver which will not run integration tests but wait for a given duration so that a user can send custom requests.

## How to test changes:

Build test binary
```
make bin/test
```
Run test driver for 5 minutes:
```
SLEEP=5m make benchmark
```

The log should show something like:
```
sleeping for 5m0s
...
sleep period is complete
```


hiromis:
Closing this PR as I chose to take a different path.

PR #488, opened 2021-08-27
Don't delete non-existent resources in non-wildcard mode in delta xds

s-matyukevich:
Before this PR we used to always add resource to the `RemovedResources` field of the `DeltaDiscoveryResponse` if it doesn't exist on the server. This causes problems in the following scenario:
1. Control plane instructs envoy to create a cluster "foo" via CDS.
2. Envoy sends `DeltaDiscoveryRequest` with type `ClusterLoadAssignment` and `ResourceNamesSubscribe` field set to `[]string{"foo"}`
3. At the moment control plane doesn't have any endpoints for the cluster "foo", so it sends `DeltaDiscoveryResponse` with `RemovedResources` name set to `[]string{"foo"}`, which instructs envoy to unsubscribe from this resource.
4. Envoy sends ACK and control plane creates a delta watch that doesn't watch for changes in "foo".
5. After some time a few endpoints are registered in the control plane for the cluster "foo". Control plane updates corresponding cache (either simple or linear)
6. Envoy doesn't receive any updates.

This PR fixes the issue by preventing resources from being deleted for non-wildcard requests. Instead, now we are going to wait for the client to unsubscribe from resources by itself. This works fine if the client always uses wildcard requests to pull initial resources and it also uses non-wildcard requests only to pull dependent resources (This is the case with envoy: it always uses wildcard requests to pull clusters and listeners and then non-wildcard requests to pull all other dependent resources)


alecholmez:
@s-matyukevich I'm not sure if you've seen this PR: https://github.com/envoyproxy/go-control-plane/pull/461 but maybe the ordering of the resources issue can be fixed by porting this over to delta

alecholmez:
@snowp your input might be valuable here, I remember us having a conversation on this bit of the logic but I can't remember the decisions as to why we decided to not keep the blank string entry

s-matyukevich:
> @s-matyukevich I'm not sure if you've seen this PR: #461 but maybe the ordering of the resources issue can be fixed by porting this over to delta

Yeah, I saw this PR and implemented ordering. It didn't. help me to fix integration tests because I was also trying to implement [this](https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#id2) part of the spec (sent DeltaDiscoverResponse with empty Resource field for non-existent resources) In this case envoy indeed deleted the resource immediately, which was causing issue mo matter how we order resources. 

After I updated the code and stopped sending anything for non-existent resources requested in non-wildcard mode, everything started working as expected. (We just wait for envoy to unsubscribed from resources that are no longer references by any other resource) This also works ok with any resource ordering, so I deleted the ordering code to keep this PR focused. 

I am not sure whether and how we are supposed to implement the above mentioned part of the spec.

snowp:
>@snowp your input might be valuable here, I remember us having a conversation on this bit of the logic but I can't remember the decisions as to why we decided to not keep the blank string entry

I don't recall what the semantics of empty string used to mean, but we should definitely be keeping track of resources that are currently being subscribed to but they don't have a current value.

s-matyukevich:
> Haven't read the full thread but this doesn't seem right:
> 
> ```
> At the moment control plane doesn't have any endpoints for the cluster "foo", so it sends DeltaDiscoveryResponse with RemovedResources name set to []string{"foo"}, which instructs envoy to unsubscribe from this resource.
> 
> Envoy sends ACK and control plane creates a delta watch that doesn't watch for changes in "foo".
> ```
> 
> Whether or not a resource is subscribed to should not be up to the server at all, the client knows what resources it wants. `removed_resources` should be used to instruct the client that a resource that it is subscribed to and thinks exist no longer does, but once it's removed from the client the subscription should remain (unless the client includes it in a unsubscribe call).

I updated the code accordingly to this comment. Now we keep the subscription active for the non-existent resources while still populating `removed_resources` field. I'll test this in our staging env tomorrow.

s-matyukevich:
@snowp @alecholmez I tested the latest version of this PR in our staging environment and verified that it indeed fixes the issue.

alecholmez:
Cool I think this satisfies everything we need. @snowp this look good to you? I have nothing else to add

pkg/cache/v3/delta.go
```diff
@@ -56,13 +56,18 @@ func createDeltaResponse(ctx context.Context, req *DeltaRequest, state stream.St
 					filtered = append(filtered, r)
 				}
 				nextVersionMap[name] = nextVersion
+			} else {
+				// we keep traking non-existent resources for non-woldcard streams until the client explicitly unsubscribe from them.
```
alecholmez, line 60:
```suggestion
				// we track non-existent resources for non-wildcard streams until the client explicitly unsubscribes from them.

```

pkg/cache/v3/delta.go
```diff
@@ -56,13 +56,18 @@ func createDeltaResponse(ctx context.Context, req *DeltaRequest, state stream.St
...
+				nextVersionMap[name] = ""
 			}
 		}
 	}
 
 	// Compute resources for removal regardless of the request type
-	for name := range state.GetResourceVersions() {
-		if _, ok := resources.resourceMap[name]; !ok {
+	for name, prevVersion := range state.GetResourceVersions() {
+		// we need prevVersion != "" check to make sure we are only sending the update to the client once (right after it was removed)
```
alecholmez, line 68:
```suggestion
// The prevVersion != "" check is in place to make sure we are only sending an update to the client once right after it is removed.
```

pkg/cache/v3/delta.go
```diff
@@ -56,13 +56,18 @@ func createDeltaResponse(ctx context.Context, req *DeltaRequest, state stream.St
...
+		// if the client decides to keep the subscription for this resource we won't add it to every subsequest response.
```
alecholmez, line 69:
```suggestion
		// If the client decides to keep the subscription we skip the add for every subsequent response.
```

alecholmez (changes requested):
Just a few comment nitpicks. Everything else looks good though

alecholmez:
can you check the build @s-matyukevich

s-matyukevich:
> Hmm it seems like envoy is seg faulting..

https://github.com/envoyproxy/go-control-plane/pull/490#issuecomment-912211182

I am pretty sure something is wrong with the latest envoy dev build.

alecholmez:
I've just notified the Envoy team about the segfault

s-matyukevich:
> I've just notified the Envoy team about the segfault

Thanks! This  https://github.com/envoyproxy/go-control-plane/issues/493 also could be helpful (Basically, I did a binary search on recent envoy commits and figured out the one after which the build starts failing) I was also trying to use `stack_decode.py` to decode the stack trace, but I have some issues with building the envoy binary with debug symbols in using it in CI.

alecholmez:
Awesome thank you @s-matyukevich. Would you also mind alerting the security team as well? Here's a [doc](https://github.com/envoyproxy/envoy/security/policy) for how to do so

s-matyukevich:
> Awesome thank you @s-matyukevich. Would you also mind alerting the security team as well? Here's a [doc](https://github.com/envoyproxy/envoy/security/policy) for how to do so

Sure, done.

s-matyukevich:
@alecholmez This now passes CI, anything else needs to be done before we can merge?

pkg/cache/v3/delta.go
```diff
@@ -56,13 +56,19 @@ func createDeltaResponse(ctx context.Context, req *DeltaRequest, state stream.St
 					filtered = append(filtered, r)
 				}
 				nextVersionMap[name] = nextVersion
+			} else {
+				// we track non-existent resources for non-wildcard streams until the client explicitly unsubscribes from them.
+
```
alecholmez, line 61: can you remove this space?

pkg/cache/v3/delta.go
```diff
@@ -56,13 +56,19 @@ func createDeltaResponse(ctx context.Context, req *DeltaRequest, state stream.St
 					filtered = append(filtered, r)
 				}
 				nextVersionMap[name] = nextVersion
+			} else {
+				// we track non-existent resources for non-wildcard streams until the client explicitly unsubscribes from them.
```
alecholmez, line 60:
```suggestion
				// We track non-existent resources for non-wildcard streams until the client explicitly unsubscribes from them.

```

alecholmez (changes requested):
Last 2 things! I'll approve and merge once these are addressed

s-matyukevich:
> Last 2 things! I'll approve and merge once these are addressed

Sure, done.

build/integration.sh
```diff
@@ -39,6 +39,7 @@ ENVOY_PID=$!
 function cleanup() {
   kill ${ENVOY_PID} ${UPSTREAM_PID}
   wait ${ENVOY_PID} ${UPSTREAM_PID} 2> /dev/null || true
+  cat ${ENVOY_LOG}
```
alecholmez, line 42: Ah can you remove this cat? We don't want to pollute this log here

alecholmez (changes requested):

build/integration.sh
s-matyukevich: oops, I forgot about this, this is fixed now.

alecholmez: Awesome thank you!

alecholmez (approved):
Looks good to me! Thanks for all the cooperation @s-matyukevich :smile:

PR #490, opened 2021-09-01
Implement tracking of known resource names by caller stream. Fixes #399

### LINKED ISSUE #399: snapshot cache doesn't handle xDS request correctly
based on xDS protocol: [https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol]

> When the client sends a new request that changes the set of resources being requested, the server must resend any newly requested resources, even if it previously sent those resources without having been asked for them and the resources have not changed since that time. 

It basically means just check version_info is not enough to know if discovery server should send a response. so this code won't send response if the request just adding a new name with previous version_info:
```
	// if the requested version is up-to-date or missing a response, leave an open watch
	if !exists || request.VersionInfo == version {
```
It need check both version_info and resource_names, which is much more complicated. Because the request won't carry previous resource_names so we must store it somewhere in the cache server, but CreateWatch() interface is called per request level rather than per stream level, so there is no way to know which resource_names should be used. I think to fully resolve this issue correctly we need add the stream id information in the CreateWatch() function.

kkalin68:
The change is implemented similarly to Java implementation.
The control plane keeps track of resource names that a caller knows already.

Signed-off-by: Konstantin Kalin <konstantin.kalin@socotra.com>

### DIFF
--- a/pkg/cache/v3/simple.go
+++ b/pkg/cache/v3/simple.go
@@ -314,10 +314,10 @@ func (cache *snapshotCache) CreateWatch(request *Request, streamState stream.Str
 				diff = append(diff, r)
 			}
 		}
-		if cache.log != nil {
-			cache.log.Debugf("nodeID %q requested %s%v and known %v. Diff %v", nodeID,
-				request.TypeUrl, request.ResourceNames, knownResourceNames, diff)
-		}
+
+		cache.log.Debugf("nodeID %q requested %s%v and known %v. Diff %v", nodeID,
+			request.TypeUrl, request.ResourceNames, knownResourceNames, diff)
+
 		if len(diff) > 0 {
 			resources := snapshot.GetResourcesAndTTL(request.TypeUrl)
 			for _, name := range diff {


pkg/server/sotw/v3/server.go
```diff
@@ -176,6 +184,16 @@ func (s *server) process(stream Stream, reqCh <-chan *discovery.DiscoveryRequest
 		// increment nonce
 		streamNonce = streamNonce + 1
 		out.Nonce = strconv.FormatInt(streamNonce, 10)
+
+		lastResponse := latestDiscoveryResponse{
+			nonce:     out.Nonce,
+			resources: make(map[string]struct{}),
+		}
+		for _, r := range resp.(*cache.RawResponse).Resources {
```
zhishi, line 192: i think here we should use the ResourceNames from original request rather than the response resources list. because there is chance the response list contain more names than the request list (which although not make sense but still valid logic). in that case we still want resent the response if the new request adding more names already sent before.

kkalin68: I was thinking about it as well. But I believe the implemented logic has more sense (less confusing) since it implies the tracking of resource names which are sent back to the caller. Java control plane implementation implements exactly same logic as I did in PR.

zhishi: it does not fully follow the protocol, which means the java implementation also buggy. this is what the protocol defined:

> When the client sends a new request that changes the set of resources being requested, the server must resend any newly requested resources, even if it previously sent those resources without having been asked for them and the resources have not changed since that time.

kkalin68: I see. It is easy to change the logic but it will be hard to test in unit-test since the implementation sends only requested resources. Shall I change it ? Any other thoughts

zhishi: I think you can get the original request by RawResponse.Request object, and you only need compare the names.

kkalin68: Response has GetRequest method that gives the original request. As I said it is easy to change

kkalin68: Changed in https://github.com/envoyproxy/go-control-plane/pull/490/commits/1df2abd6aebfce09974db904e0ba47e6843f7c89

kkalin68:
It seems that CI failure is unrelated to my changes. `make docker_tests` did pass yesterday locally (before I opened PR) as well as CI passed when it was triggered.

To verify that I just reverted my changes locally and run `make docker_tests`. It failed exactly at same place. What should I do in this situation?

kkalin68:
Checking the test locally I think envoy crashes with Segmentation fault in XDS mode and I would say it is a reason of failed integration tests. Envoy crashes with and without my changes.

kkalin68:
@alecholmez @snowp Is there anything I can do to get PR merged ? 
I am currently using my fork to build the control plane because of this issue.

alecholmez:
@kkalin68 sorry for the delay I've had relatively low bandwidth recently. I'll make some time to review this. Thanks for your patience!

pkg/cache/v3/cache.go
```diff
@@ -48,10 +48,11 @@ type ConfigWatcher interface {
 	// An individual consumer normally issues a single open watch by each type URL.
 	//
 	// The provided channel produces requested resources as responses, once they are available.
+	// The provided map contains resources already known to the caller
 	//
 	// Cancel is an optional function to release resources in the producer. If
 	// provided, the consumer may call this function multiple times.
-	CreateWatch(*Request, chan Response) (cancel func())
+	CreateWatch(*Request, chan Response, map[string]struct{}) (cancel func())
```
alecholmez, line 55: How difficult would it be to follow the pattern laid in place by delta here (utilizing the streamState object)? While they achieve the same thing I'd rather unify the way that's done and we've already established that with delta. I was originally going to take this task and follow that pattern but you beat me to it :)

pkg/cache/v3/simple.go
```diff
@@ -306,6 +306,39 @@ func (cache *snapshotCache) CreateWatch(request *Request, value chan Response) f
 	snapshot, exists := cache.snapshots[nodeID]
 	version := snapshot.GetVersion(request.TypeUrl)
 
+	if exists && knownResourceNames != nil {
```
alecholmez, line 309: If this is the initial request and theres no previously known resources will this be guaranteed to be initialized?

pkg/cache/v3/simple.go
```diff
@@ -306,6 +306,39 @@ func (cache *snapshotCache) CreateWatch(request *Request, value chan Response) f
...
+		}
+		if len(diff) > 0 {
+			found := false
+			if cache.log != nil {
+				cache.log.Debugf("nodeID %q still needs %v", nodeID, diff)
+			}
+
+			resources := snapshot.GetResourcesAndTTL(request.TypeUrl)
+			for _, name := range diff {
+				if _, exists := resources[name]; exists {
```
alecholmez, line 328: So I think this will yield unwanted results if we enter a case where resources have the same name but different objects. That's one of the reasons we did the hash map in incremental so we handle that scenario. I'd be okay with moving a field inside of the streamState to something like `State map[string]struct{} // current state of resources in SOTW` where the `struct{}` is the value we can compare. That way we guarantee we won't falsely match

pkg/server/sotw/v3/server.go
```diff
@@ -102,6 +102,11 @@ type watches struct {
 	nonces        map[string]string
 }
 
+type latestDiscoveryResponse struct {
```
alecholmez, line 105: I think we could move things like this into the `StreamState`. That way we don't need to declare extra things just for SOTW

alecholmez (changes requested):
Just a few things to get started

pkg/cache/v3/cache.go
kkalin68: I will check. I think what happened my initial patch was based of 0.9.9 and I don't recall seeing streamState object.

pkg/cache/v3/simple.go
kkalin68: Not sure if I understand the question. `knownResourceNames` is initialized when a new stream is created in sowt/server.go Line 161. 
```
func (s *server) process(stream Stream, reqCh <-chan *discovery.DiscoveryRequest, defaultTypeURL string) error {
...
	knownResourceNames := map[string]map[string]struct{}{}
```

pkg/cache/v3/simple.go
kkalin68: Sorry, not sure if I follow again. What do you mean in `a case where resources have the same name but different objects.` ? 

The logic I added there checks if newly requested resource is already known to control plane.

pkg/cache/v3/cache.go
alecholmez: Okay thank you :) that would be great if we could migrate over to the same pattern

pkg/cache/v3/simple.go
alecholmez: Right I saw that, I only ask cause there may be a case when a consumer of this API only uses the cache and writes their own server implementation. If they don't initialize knownResourceNames I believe this will panic. Maybe add a nil check at the top?

pkg/cache/v3/simple.go
alecholmez: Ah sorry I think I misunderstood, just tracking the resource name is sufficient for creating a diff in SOTW. I was in the incremental xDS mindset where we'd also need to create versions for the new resource. Ignore my previous statement

pkg/cache/v3/simple.go
kkalin68: I see. It should not panic and it doesn't panic right now in Unit tests :). I added this check to avoid adding more changes in all unit tests. Majority of Unit tests passes `knownResourceNames=nil` . There are specific Unit tests for case when `knownResourceNames != nil` to check the logic.
Anyway I will see how the code is changing when streamState is in play.

pkg/cache/v3/cache.go
kkalin68: Coo. I should have something early of next week.

alecholmez: Awesome thank you so much :)

pkg/cache/v3/simple.go
alecholmez: Great that sounds good

kkalin68:
@alecholmez Got buried by urgent work, so I couldn't finished yet. I will continue on the change tomorrow or day after tomorrow. Sorry for a delay.

alecholmez:
> @alecholmez Got buried by urgent work, so I couldn't finished yet. I will continue working on the change tomorrow or day after tomorrow. Sorry for a delay.

@kkalin68 no worries! Take your time

alecholmez:
@kkalin68 ping, any news on this?

kkalin68:
@alecholmez I did a change and tested on my env. It works. Please check it out. I am thinking about if `latestDiscoveryResponse` should be part of StreamState as well and implement similar logic for delta. But I don't know how the delta should work thus it is hard for me to make a judgment call. Please advise.

pkg/server/stream/v3/stream.go
```diff
@@ -30,6 +30,9 @@ type StreamState struct { // nolint:golint,revive
 	// This field stores the last state sent to the client.
 	resourceVersions map[string]string
 
+	// knowResourceName contains resources names that a client has recieved previously
```
jpeach, line 33:
```suggestion
	// knownResourceNames contains resources names that a client has received previously.
```

pkg/server/sotw/v3/server.go
alecholmez: A comment here noting why we need this might be helpful

pkg/server/stream/v3/stream.go
```diff
@@ -30,6 +30,9 @@ type StreamState struct { // nolint:golint,revive
 	// This field stores the last state sent to the client.
 	resourceVersions map[string]string
 
+	// knownResourceName contains resources names that a client has received previously
```
alecholmez, line 33:
```suggestion
	// knownResourceNames contains resource names that a client has received previously
```

alecholmez (changes requested):
Just a few questions! Hopefully we can merge this soon

kkalin68:
> Hm super weird on the failing tests I can see what you're saying. In the PR that line is different than what the tests are running.
> 
> EDIT: I actually have no clue why CI is failing either. I cloned your fork and branch and it built/ran the integration tests perfectly fine.

Should I create a new branch and cherry-pick my changes ? I saw such errors before after a branch was rebased previously. Although another rebase normally fixes it which didn't happen this time :)

pkg/cache/v3/simple.go
```diff
@@ -306,6 +306,40 @@ func (cache *snapshotCache) CreateWatch(request *Request, value chan Response) f
...
+					found = true
+					break
+				}
+			}
+
+			// cache contains resources already, the watch may be responded immediately
+			if found {
+				_ = cache.respond(context.Background(), request, value, resources, version, false)
```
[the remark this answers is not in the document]
kkalin68: Will do

pkg/cache/v3/simple.go
```diff
@@ -306,6 +306,40 @@ func (cache *snapshotCache) CreateWatch(request *Request, value chan Response) f
...
+				diff = append(diff, r)
+			}
+		}
+		if cache.log != nil {
+			cache.log.Debugf("nodeID %q requested %s%v and known %v. Diff %v", nodeID,
+				request.TypeUrl, request.ResourceNames, knownResourceNames, diff)
+		}
+		if len(diff) > 0 {
+			found := false
+			if cache.log != nil {
```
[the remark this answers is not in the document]
kkalin68: Done

pkg/cache/v3/simple.go
```diff
@@ -306,6 +306,40 @@ func (cache *snapshotCache) CreateWatch(request *Request, value chan Response) f
...
+				cache.log.Debugf("nodeID %q still needs %v", nodeID, diff)
+			}
+
+			resources := snapshot.GetResourcesAndTTL(request.TypeUrl)
+			for _, name := range diff {
+				if _, exists := resources[name]; exists {
+					found = true
```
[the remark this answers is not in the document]
kkalin68: Done

pkg/server/stream/v3/stream.go
kkalin68: Done

pkg/cache/v3/simple_test.go
```diff
@@ -409,6 +410,76 @@ func TestSnapshotCacheWatchTimeout(t *testing.T) {
 	}
 }
 
+func TestSnapshotCreateWatchWithResourcePrevioslyNotRequested(t *testing.T) {
```
jpeach, line 413:
```suggestion
func TestSnapshotCreateWatchWithResourcePreviouslyNotRequested(t *testing.T) {
```

pkg/cache/v3/simple.go
```diff
@@ -286,7 +286,7 @@ func superset(names map[string]bool, resources map[string]types.ResourceWithTTL)
 }
 
 // CreateWatch returns a watch for an xDS request.
-func (cache *snapshotCache) CreateWatch(request *Request, value chan Response) func() {
+func (cache *snapshotCache) CreateWatch(request *Request, streamState *stream.StreamState, value chan Response) func() {
```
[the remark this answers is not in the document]
alecholmez: IMO I think it would be better if we didn't make this a pointer. I see your point but I'd like to match the external cache API to Delta so users don't get confused. `CreateDeltaWatch` doesn't take a pointer so it would make sense for this to not either.

pkg/cache/v3/simple.go
```diff
@@ -306,6 +306,34 @@ func (cache *snapshotCache) CreateWatch(request *Request, value chan Response) f
...
 
+	if exists {
+		knownResourceNames := streamState.GetKnownResourceNames(request.TypeUrl)
+		diff := []string{}
+		for _, r := range request.ResourceNames {
+			if _, ok := knownResourceNames[r]; !ok {
+				diff = append(diff, r)
+			}
+		}
+		if cache.log != nil {
```
alecholmez, line 317: Sorry you can remove this nil check here. We no longer require that

pkg/cache/v3/simple.go
```diff
@@ -306,6 +306,34 @@ func (cache *snapshotCache) CreateWatch(request *Request, value chan Response) f
...
+			cache.log.Debugf("nodeID %q requested %s%v and known %v. Diff %v", nodeID,
+				request.TypeUrl, request.ResourceNames, knownResourceNames, diff)
+		}
+		if len(diff) > 0 {
+			resources := snapshot.GetResourcesAndTTL(request.TypeUrl)
+			for _, name := range diff {
+				if _, exists := resources[name]; exists {
+					if err := cache.respond(context.Background(), request, value, resources, version, false); err != nil {
+						if cache.log != nil {
```
alecholmez, line 326: Same thing here, no more nil check needed we just merged a PR to fix this

pkg/cache/v3/simple.go
```diff
@@ -321,7 +349,12 @@ func (cache *snapshotCache) CreateWatch(request *Request, value chan Response) f
 
 	// otherwise, the watch may be responded immediately
 	resources := snapshot.GetResourcesAndTTL(request.TypeUrl)
-	_ = cache.respond(context.Background(), request, value, resources, version, false)
+	if err := cache.respond(context.Background(), request, value, resources, version, false); err != nil {
+		if cache.log != nil {
```
alecholmez, line 353: Ditto

alecholmez: And should we return nil in here?

alecholmez (changes requested):
Nice this is looking good, just one question and then I forgot to mention a few changes last time sorry about that

pkg/server/stream/v3/stream.go
```diff
@@ -51,12 +54,29 @@ func (s *StreamState) IsWildcard() bool {
 	return s.wildcard
 }
 
+func (s *StreamState) SetKnownResourceNames(url string, names map[string]struct{}) {
+	s.knownResourceNames[url] = names
+}
+
+func (s *StreamState) SetKnownResourceNamesAsList(url string, names []string) {
```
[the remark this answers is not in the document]
kkalin68: Right now it is used by tests only. But I have a feeling that the implementation of sotw/server.go may change once the delta support is added. So streamState may have common logic there. I hope it makes sense.

pkg/cache/v3/simple.go
kkalin68: It does return `nil` in line 359

pkg/cache/v3/simple.go
kkalin68: done

alecholmez:
@kkalin68 Nope I think it's time to open a new PR

PR #491, opened 2021-09-06
Introduce ResourceSnapshot interface.

### LINKED ISSUE #489: Add a snapshot interface for caching non-Envoy types
Kuma has some services that use xDS to transfer types that are not Envoy protobufs. The snapshot cache works well for this, but the snapshot type itself doesn't let you use arbitrary resource types.

If we use an interface in the snapshot cache, we can allow API clients to use it for non-envoy types too. for example:

```Go
type ResourceSnapshot interface {
       GetVersion(typeURL string) string

       ConstructVersionMap() error
       GetVersionMap(typeURL string) map[string]string

       GetResources(typeURL string) map[string]types.Resource
       GetResourcesAndTtl(typeURL string) map[string]types.ResourceWithTTL
}
```

jpeach:
Introduce a ResourceSnaphot interface to abstract the type of the snapshot
stored in a SnapshotCache. This allows applications to use SnapshotCache
without also having to use Snapshot. The usefulness of that is that it
is possible to use SnapshotCache with non-Envoy xDS applications.

This fixes #489.

### DIFF
--- a/pkg/cache/v3/simple.go
+++ b/pkg/cache/v3/simple.go
@@ -26,6 +26,35 @@ import (
 	"github.com/envoyproxy/go-control-plane/pkg/server/stream/v3"
 )
 
+// ResourceSnapshot is an abstract snapshot of a collection of resources that
+// can be stored in a SnapshotCache. This enables applications to use the
+// SnapshotCache watch machinery with their own resource types. Most
+// applications will use Snapshot.
+type ResourceSnapshot interface {
+	// GetVersion should return the current version of the resource indicated
+	// by typeURL. The version string that is returned is opaque and should
+	// only be compared for equality.
+	GetVersion(typeURL string) string
+
+	// GetResourcesAndTTL returns all resources of the type indicted by
+	// typeURL, together with their TTL.
+	GetResourcesAndTTL(typeURL string) map[string]types.ResourceWithTTL
+
+	// GetResources returns all resources of the type indicted by
+	// typeURL. This is identical to GetResourcesAndTTL, except that
+	// the TTL is omitted.
+	GetResources(typeURL string) map[string]types.Resource
+
+	// ConstructVersionMap is a hint that a delta watch will soon make a
+	// call to GetVersionMap. The snapshot should construct an internal
+	// opaque version string for each collection of resource types.
+	ConstructVersionMap() error
+
+	// GetVersionMap returns a map of resource name to resource version for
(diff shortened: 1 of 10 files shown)


pkg/cache/v3/simple.go
```diff
@@ -26,6 +26,35 @@ import (
 	"github.com/envoyproxy/go-control-plane/pkg/server/stream/v3"
 )
 
+// ResourceSnapshot is an abstract snapshot of a collection of resources that
+// can be stored in a SnapshotCache. This enables applications to use the
+// SnapshotCache watch machinery with their own resource types. Most
+// applications will use Snapshot.
+type ResourceSnapshot interface {
```
alecholmez, line 33: If this is an interface at this high of a level why not just call it a `Snapshot`? I think that would make more sense and then subsequent structs that implement can name themselves accordingly. I imagine the snapshot itself as the blueprint for a grouping of resources looks like.

Changing this would require we change the original snapshot struct though which I'm not opposed to

alecholmez (changes requested):

pkg/cache/v3/simple.go
jpeach: Yeh I didn't want to rename `Snapshot`, because it's pretty disruptive and I also didn't have a good suggestion for a new name :)

alecholmez: Hm, what are your thoughts on if we were to deal with that upfront cost? You think the API would be easier to use/consume?

jpeach: I'm not sure that there are big wins to the API from renaming. It's probably a bit nicer, but not sure it's really a big deal. If we have a a larger goal to separate out the envoy protobuf dependencies, maybe we would think of it as part of that.

jpeach: @alecholmez I'm  still inclined to keep the current `Snapshot` name as is for now.

akshaysngupta:
Any update on this ?

jpeach:
> Any update on this ?

I plan to rebase and update this in the next few days.

jpeach:
@alecholmez I rebased and added a basic test, but otherwise decided not to change anything. I'm thinking that we may want to remove `(ResourceSnapshot)ConstructGetVersionMap()` and just build it on demand, but it's probably better to do that in a separate PR.

alecholmez (approved):
I think this is good to go in

PR #495, opened 2021-09-15
server: provide Scoped Routes Discovery Service (SRDS) support

mtagstyle:
[server]
- Updated the server subsystem, to support opening SRDS streams.
- Updated associated test cases under the server subsystem.

[snapshot]
- Refactored Consistency() check to handle resource checking more generically, and support for ScopedRoutes.
- Updated and added test cases under the resource and snapshot subsystem.

[integration tests]
- Updated the integration test to use ScopedRoutes filter which points to a Route.


pkg/cache/v3/resource.go
```diff
@@ -111,57 +117,103 @@ func MarshalResource(resource types.Resource) (types.MarshaledResource, error) {
...
-			case *cluster.Cluster_Type:
-				if typ.Type == cluster.Cluster_EDS {
-					if v.EdsClusterConfig != nil && v.EdsClusterConfig.ServiceName != "" {
-						out[v.EdsClusterConfig.ServiceName] = true
-					} else {
-						out[v.Name] = true
-					}
-				}
-			}
+			getClusterReferences(v, out)
```
alecholmez, line 138: I'm a bit more of a fan of this function returning a map instead of taking an `out` param. That we can we can remove the nil checks and just always assume one will exist since we create it. 

Then here maybe do something like
```go
return getClusterReference(v)
```

pkg/cache/v3/resource.go
```diff
@@ -111,57 +117,103 @@ func MarshalResource(resource types.Resource) (types.MarshaledResource, error) {
...
+			} else {
+				out[resource.EndpointType][src.Name] = true
+			}
+		}
+	}
+}
+
+// HTTP listeners will either reference ScopedRoutes or Routes.
+func getListenerReferences(src *listener.Listener, out map[resource.Type]map[string]bool) {
+
```
alecholmez, line 173: extra `\n` here

pkg/test/resource/v3/resource.go
```diff
@@ -133,6 +133,28 @@ func MakeRoute(routeName, clusterName string) *route.RouteConfiguration {
 	}
 }
 
+// MakeScopedRoute creates an HTTP scoped route that routes to a given cluster.
+func MakeScopedRoute(scopedRouteName string, routeConfigurationName string, keyFragments []string) *route.ScopedRouteConfiguration {
+
```
alecholmez, line 138: extra `\n`

pkg/server/v3/server_test.go
```diff
@@ -154,7 +155,8 @@ var (
 	endpoint        = resource.MakeEndpoint(clusterName, 8080)
 	cluster         = resource.MakeCluster(resource.Ads, clusterName)
 	route           = resource.MakeRoute(routeName, clusterName)
-	listener        = resource.MakeHTTPListener(resource.Ads, listenerName, 80, routeName)
+	scopedRoute     = resource.MakeScopedRoute(scopedRouteName, routeName, []string{"127.0.0.1"})
+	listener        = resource.MakeScopedRouteHTTPListener(resource.Ads, listenerName, 80, scopedRouteName)
```
alecholmez, line 159: Do we no longer test a normal `HttpRouteListener`?

pkg/cache/v3/snapshot.go
```diff
@@ -81,19 +82,47 @@ func (s *Snapshot) Consistent() error {
...
-	routes := GetResourceReferences(s.Resources[types.Listener].Items)
-	if len(routes) != len(s.Resources[types.Route].Items) {
-		return fmt.Errorf("mismatched route reference and resource lengths: %v != %d", routes, len(s.Resources[types.Route].Items))
+	for idx, items := range s.Resources {
+
+		// We only want to check resource types that are expected to be referenced by another resource type.
+		// Basically, if the consistency relationship is modeled as a DAG, we only want
+		// to check nodes that are expected to have edges pointing to it.
+		responseType := types.ResponseType(idx)
+		if _, ok := referencedResponseTypes[responseType]; ok {
```
alecholmez, line 107: Don't you also need to check the returned value from this map access? The second value only checks to see if the key in the map exists so `types.Endpoint` could be `false` and `ok` will be true.

```
if yes, ok := referencedResponseTypes[responseType]; ok && yes {

}
```

pkg/cache/v3/snapshot.go
```diff
@@ -81,19 +82,47 @@ func (s *Snapshot) Consistent() error {
...
-	endpoints := GetResourceReferences(s.Resources[types.Cluster].Items)
-	if len(endpoints) != len(s.Resources[types.Endpoint].Items) {
-		return fmt.Errorf("mismatched endpoint reference and resource lengths: %v != %d", endpoints, len(s.Resources[types.Endpoint].Items))
-	}
-	if err := superset(endpoints, s.Resources[types.Endpoint].Items); err != nil {
-		return err
+
+	consolidated := make(map[resource.Type]map[string]bool)
+	// References made by CDS.
+	GetResourceReferencesByReference(s.Resources[types.Cluster].Items, consolidated)
```
alecholmez, line 88: If we changed this function to just check all references in the snapshot I don't think we'd need to make 3 separate calls here. Does the previous consistency check not perform this same functionality? I'm a bit confused on this change

alecholmez (changes requested):
Thanks for the changes! I think this is definitely on the right path I just had a few questions to get you started.

pkg/cache/v3/snapshot.go
mtagstyle: This map is supposed to be a "set", but I see the concern. I'm going to change the struct type to map[string]struct{} that way there is no confusion with the boolean values.

pkg/cache/v3/resource.go
mtagstyle: Yeah it was originally just returning a new param instead of taking a reference param, but there was some inefficiency with it later on where you would have to combine all the maps you made later on. Instead, I opted to just use 1 shared map.

I think I have a better solution which meets both of our concerns though, using what you recommended. (see my long winded comment below)

pkg/cache/v3/snapshot.go
mtagstyle: So the old consistency check would basically do the following:

1.a) For all clusters in the snapshot, check the contents of each cluster resource, to gather if there are any references to other resources. Assume here that clusters can ONLY reference endpoints.

1.b) For each ENDPOINT in the snapshot, check to see if it is in the referenced resources list.

2.a) For all listeners in the snapshot, check the contents of each listener resource, to gather if there are any references to other resources. Assume here that listeners can ONLY reference routes.

2.b) For each ROUTE in the snapshot, check to see if it is in the referenced resources list.

Check 1.b, and 2.b here is done counterintuitively (Checking resources in the snapshot AGAINST what was referenced) vs. (Checking what was referenced AGAINST what's in the snapshot) because of a time complexity optimization. (Looking it up 1 way is O(n) vs O(1)).

Now this worked fine without SRDS. But with SRDS, listeners can reference both Scoped Routes and Routes at the same time. Which ends up breaking the entire algorithm for the following reason:

1) The "GetResourceReferences" function used to just given a list of resources, return a flat list of resources. It was then up to the caller to assume what the returned resource type was. I have updated this to instead a map looking like the following...

```
GetResourceReferences(myResources)
```
| ResourceType | Resources By Name|
| ----------- | ----------- |
| Endpoint | "endpoint1", "endpoint2" ... "endpoint N" |
| Scoped Route | "scopedroute1", "scopedroute2", ... "scopedrouteN" |
| Route | "route1", "route2", ... "route N" |

mtagstyle: I think with the new solution, we can do what you recommended and just check all resources for any references. I.e.

```
// Goes through the snapshot and returns a map of all referenced resources in the snapshot, keyed by type.
func GetReferencedResources() map[resource.Type]map[string]bool
```

However, we may incur some runtime tradeoffs here if we naively check every resource for references. (For instance, there's no way a route can reference another resource, so why bother checking it at all).

mtagstyle: Let me know what your thoughts are on this.

pkg/server/v3/server_test.go
mtagstyle: I'll update the integration test again to make sure we test both, but yes you are correct I removed it thinking it didn't matter :)

pkg/cache/v3/resource.go
mtagstyle: Nvm, I get it.

pkg/cache/v3/snapshot.go
alecholmez: I think this makes sense thanks @mtagstyle for the little write up here. So since we don't want to waste compute cycles on checking if a route is consistent, we only really need to verify listeners and their associated routes/scoped routes, clusters and their endpoints, etc.. I think I'm picking up what you're saying.

Let me go look at this new algorithm

alecholmez: Okay I think this makes sense to me now

So I see why we went this way with the algorithm, I think my only hesitation at this point is it looks like the worst case scenario of time complexity here is `O(n^4)` since when we're doing the map merges we have quite a few nested loops. Are there any optimizations we can make here computation wise? I imagine at really large scale if a user checked a snapshots consistency, (10k+ configuration objects), this might cause some issues

mtagstyle: I'm drawing through the algorithm again, starting from "GetAllResourceReferences" and I can't see why we would have O(N^4). 

If we let N be the number of total resources in the snapshot, then I only see a static case of O(N) time complexity. To be more precise, it would be O(3*N). Refer to attached diagram for more details... Since I suck at communicating things through text.

image

As for optimizations, I think the biggest win would be to not bother checking every single resource, and only the ones we expect to have references (Refer to "resourceGroups" of Fig. 1)

Next, if we just pass the outMap all the way through to getXReferences (But add in the hacky nil checks), we would only have to pass through the referenced resources once, putting us at time complexity O(2*N).

mtagstyle: Let me know what you think on my recommendations for optimizations.

alecholmez: Ah yes you're totally right. I went a re-read it and its not quadratic. It's definitely linear time, thanks for the diagram that helped :) 

I think it would be nice if we just go ahead and make those optimizations you recommended. In my head I imagine people making these consistency checks often (at least once per snapshot update) which would happen a lot especially on the startup of a mesh or xDS system.

alecholmez (approved):
I think this is good now! Thanks for working back and forth on this.

PR #496, opened 2021-09-22
Actually wire up Envoy logging to ALS for integration test cases

unicell:


unicell:
@alecholmez Do you mind taking a look? The change is bit lengthier (i.g touching all the bootstrap yamls) than I'd like it to be. However, due to https://github.com/envoyproxy/envoy/issues/3660 ALS cluster seems can only be configured as static resource.

Let me know your thoughts. Thanks!

PR #501, opened 2021-10-14
WIP: Resource Snapshot Without New Type

alecholmez:

### DIFF
--- a/pkg/cache/v3/snapshot.go
+++ b/pkg/cache/v3/snapshot.go
@@ -17,15 +17,49 @@ package cache
 import (
 	"errors"
 	"fmt"
+	"sync"
 
 	"github.com/envoyproxy/go-control-plane/pkg/cache/types"
 	"github.com/envoyproxy/go-control-plane/pkg/resource/v3"
 )
 
 // Snapshot is an internally consistent snapshot of xDS resources.
+type Snapshot interface {
+	// GetVersion returns the current version of the resources indicated
+	// by typeURL. The version string that is returned is opaque and should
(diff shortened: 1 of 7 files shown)


pkg/cache/v3/snapshot.go
```diff
@@ -160,20 +234,29 @@ func (s *Snapshot) GetVersion(typeURL resource.Type) string {
 }
 
 // GetVersionMap will return the internal version map of the currently applied snapshot.
-func (s *Snapshot) GetVersionMap(typeUrl string) map[string]string {
-	return s.VersionMap[typeUrl]
+func (s *snapshot) GetVersionMap(typeURL string) map[string]string {
+	if s == nil {
+		return nil
+	}
```
jpeach, line 240: What is the use case for calling methods on a nil object?

pkg/cache/v3/snapshot.go
```diff
@@ -17,28 +17,70 @@ package cache
...
 	Resources [types.UnknownType]Resources
 
 	// VersionMap holds the current hash map of all resources in the snapshot.
 	// This field should remain nil until it is used, at which point should be
 	// instantiated by calling ConstructVersionMap().
 	// VersionMap is only to be used with delta xDS.
 	VersionMap map[string]map[string]string
+
+	// Lock for multi-thread processing
+	mu sync.RWMutex
```
jpeach, line 72: Is the lock really needed here?

alecholmez: Negatory it isn't was just experimenting with things

PR #511, opened 2021-11-10
xdsmatcher: add evaluation framework for xDS matching API

snowp:


snowp:
@alecholmez wanna give this one a look? LMK how I can help moving this forward

snowp:
Maybe @howardjohn if you're interested in this bit as well, I'd appreciate the eyes :)

alecholmez:
> @alecholmez wanna give this one a look? LMK how I can help moving this forward

Yup looking this over, sorry got a bit wrapped up

xdsmatcher/pkg/matcher/types/types.go
```diff
@@ -0,0 +1,33 @@
+package types
```
alecholmez, line 1: Maybe some comments in this file above the important types/interfaces might be useful? Like the `Matcher` interface for example. Just so it shows up nicely in godoc for people consuming the framework

alecholmez (commented):
I think overall this looks nice. The API is pretty friendly so long as you understand the original matching API. Overall nothing major from my end. Maybe just some more comments here and there and beefing up the docs for godoc but other than that it looks good!

xdsmatcher/README.md
```diff
@@ -0,0 +1,109 @@
...
+func fooInputFactory(m proto.Message) (interface{}, error) {
+	_, ok := m.(*pbtest.FooInput)
+	if !ok {
+		return nil, errors.New("unexpected proto")
+	}
+
+	return fooInput{}, nil
+}
+
+// Implement types.DataInput for the input type.
```
alecholmez, line 33: Just so I understand correctly, is the `InputType` the registered type URL in the `init()` function? I'd assume so since we are attempting a cast below that never fails

xdsmatcher/README.md
```diff
@@ -0,0 +1,109 @@
...
+	matcher := &pbmatcher.Matcher{}
+	_ = iproto.ProtoFromYaml([]byte(yaml), matcher)
+
+	m, _ := matcher.Create(matcher)
+
+	data := &pbtest.TestData{}
+	r, _ := m.Match(data)
+	if r.MatchResult != nil && r.MatchResult.Action != nil {
+		// resulting action
+		action, ok := r.MatchResult.Action.(action)
```
alecholmez, line 96: I saw that Actions are interfaces, what kind of things could we do on this resulting variable? Or would that be for the consumer of this API to decide?

xdsmatcher/pkg/matcher/matcher.go
```diff
@@ -0,0 +1,523 @@
...
+
+		// Otherwise pass through the result.
+		return result, nil
+	}
+
+	// We failed to arrive at a result due to the data not being available,
+	// report this back up.
+	if result.NeedMoreData {
+		m.logger.log("matcher tree requires more data")
+		return result, nil
```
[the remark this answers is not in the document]
sunjayBhatia: as i understand it a result with "needs more data" is valid, might just need a small comment adjustment

xdsmatcher/README.md
snowp: Yeah they are opaque objects that are defined using extension points in the config. You'd expect to check the type of these by casting to the type you might care about, then you can do whatever you want using the resulting casted type (which may have more interesting fields/functions defined)

xdsmatcher/README.md
snowp: To make things super generic, the entire tree acts against a specific InputType (simple case here is just a string, but can be a complex data structure). Once the tree is built, you pass a value of type InputType to it, which is then passed to all DataInputs (which tell us how to extract a specific value out of the complex data type).

So the InputType is never registered, but you must be defining DataInputs that all support the specific type that you provide to the matcher on evaluation time

xdsmatcher/pkg/matcher/matcher.go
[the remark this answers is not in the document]
snowp: Yeah this is not an error, it's the complex data type understanding that this data may change in the future, so it's unable to conclusively say that the data doesn't match.

An example here is how this is used for matching HTTP streams in Envoy: if you run into a matcher that tries to match against trailers before trailers are seen, we respond "NeedMoreData" so that the caller knows to call this again when trailers are made available.

snowp:
Updated, ptal @alecholmez

snowp:
ping @alecholmez

alecholmez (approved):
I think from me reading this it looks good to go. I might have more questions once I start using it but looks ready to merge on my end.

PR #515, opened 2021-11-18
Fix scoped route consistency check.

### LINKED ISSUE #514: Scoped routes consistency check assumes that SRDS is used
In the snapshot consistency checker, the listener references assumes that SRDS is being used when coped routes are being used.

https://github.com/envoyproxy/go-control-plane/blob/967983d384a303529d544e452540f00eb106453b/pkg/cache/v3/resource.go#L222-L225

However, when you configured scoped routes, you can either used SRDS or you can provide the mappings to RDS names in the configurations list. The consistency checker should add ScopedRoutesConfigurations in the former case, and RouteConfigurations in the latter.

jpeach:
Scoped routes can be specified with a SRDS resource, or they can be
inlined, in which case a RDS resource is needed. The snapshot consistency
checker was always assuming the former case.

This fixes #514.

### DIFF
--- a/pkg/cache/v3/resource.go
+++ b/pkg/cache/v3/resource.go
@@ -215,13 +215,27 @@ func getListenerReferences(src *listener.Listener, out map[resource.Type]map[str
 			routeSpecifier := config.RouteSpecifier
 			switch r := routeSpecifier.(type) {
 			case *hcm.HttpConnectionManager_Rds:
-				if r != nil && r.Rds != nil {
-					routes[r.Rds.RouteConfigName] = true
+				if r == nil || r.Rds == nil {
+					break
 				}
 
+				routes[r.Rds.RouteConfigName] = true
+
 			case *hcm.HttpConnectionManager_ScopedRoutes:
-				if r != nil && r.ScopedRoutes != nil {
+				if r == nil || r.ScopedRoutes == nil {
+					break
+				}
+
+				// If we are using SRDS add the scoped route
+				// resource.
+				if r.ScopedRoutes.GetScopedRds() != nil {
 					scopedRoutes[r.ScopedRoutes.Name] = true
+				}
+
+				// If the scoped route mapping is embedded, add
+				// the referenced route resource names.
+				for _, s := range r.ScopedRoutes.GetScopedRouteConfigurationsList().GetScopedRouteConfigurations() {
+					routes[s.RouteConfigurationName] = true
 				}
 			}
 		}
--- a/pkg/cache/v3/snapshot_test.go
+++ b/pkg/cache/v3/snapshot_test.go
@@ -112,6 +112,30 @@ func TestScopedRouteListenerWithScopedRouteAndRouteIsConsistent(t *testing.T) {
 	}
 }
 
+func TestScopedRouteListenerWithInlineScopedRouteAndRouteIsConsistent(t *testing.T) {
+	snap, _ := cache.NewSnapshot(version, map[rsrc.Type][]types.Resource{
+		rsrc.ListenerType: {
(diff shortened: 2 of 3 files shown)


alecholmez:
@mtagstyle can you take a quick look at this?

mtagstyle:
> @mtagstyle can you take a quick look at this?

On it.

pkg/cache/v3/resource.go
```diff
@@ -215,14 +215,28 @@ func getListenerReferences(src *listener.Listener, out map[resource.Type]map[str
 			routeSpecifier := config.RouteSpecifier
 			switch r := routeSpecifier.(type) {
 			case *hcm.HttpConnectionManager_Rds:
-				if r != nil && r.Rds != nil {
-					routes[r.Rds.RouteConfigName] = true
+				if r == nil || r.Rds == nil {
+					break
```
mtagstyle, line 219: I don't understand the break here. This will cause the function to return an incomplete list of resource references. Shouldn't it be a continue?

pkg/cache/v3/resource.go
```diff
@@ -215,14 +215,28 @@ func getListenerReferences(src *listener.Listener, out map[resource.Type]map[str
...
+				}
+
+				// If we are using SRDS add the scoped route
+				// resource.
+				if r.ScopedRoutes.GetScopedRds() != nil {
 					scopedRoutes[r.ScopedRoutes.Name] = true
 				}
+
+				// If the scoped route mapping is embedded, add
+				// the referenced route resource names.
```
mtagstyle, line 236: Thanks for adding this. I never even considered the case where scoped routes could have STATIC ROUTES instead of RDS.

mtagstyle (changes requested):

pkg/cache/v3/resource.go
jpeach: This breaks from the innermost switch scope

pkg/cache/v3/resource.go
jpeach: It's still RDS, but not SDRS in this case :)

pkg/cache/v3/resource.go
mtagstyle: Right.

pkg/cache/v3/snapshot_test.go
```diff
@@ -112,6 +112,30 @@ func TestScopedRouteListenerWithScopedRouteAndRouteIsConsistent(t *testing.T) {
...
+			resource.MakeScopedRouteHTTPListenerForRoute(resource.Xds, "listener0", 80, "scopedRoute0", "testRoute0"),
+		},
+		rsrc.RouteType: {
+			resource.MakeRoute("testRoute0", clusterName),
+		},
+	})
+
+	assert.NoError(t, snap.Consistent())
+
+	snap, _ = cache.NewSnapshot(version, map[rsrc.Type][]types.Resource{
```
mtagstyle, line 127: NIT: The title only describes the first test case here for a consistent snapshot, but the second case is not referenced. Could you either update the test title, or break this test case in 2?

mtagstyle (changes requested):

pkg/cache/v3/snapshot_test.go
jpeach: I split it into 2.

mtagstyle (approved):

mtagstyle:
@alecholmez I've taken a look at the added changes and they look correct to me. Ball is back in your court.

jpeach:
xref https://github.com/envoyproxy/envoy/issues/19075

alecholmez (approved):
This looks good to me thank you

PR #518, opened 2021-11-29
Fix scoped route references.

### LINKED ISSUE #517: scoped route consistency checking is wrong
I'm not convinced that I completely understand how scoped routes are put together, but I'm fairly sure that the consistency check is not correct. In `getListenerReferences`, we crack the connection manager and, if SRDS is being used, add the name of the scoped routes element to the set of resource references that we expect:

https://github.com/envoyproxy/go-control-plane/blob/d02f089307688a1fe99f1cdf9f66623f07707fa2/pkg/cache/v3/resource.go#L229-L234

However, this isn't how scoped routes work, because there should be a `ScopedRouteConfiguration` resource for every unique key. The name in the HCM `ScopedRoutes` fields is irrelevant and does not need to match any resources. IIUC, what happens is that Envoy collects all the `ScopedRouteConfiguration` resources, and indexes the correct one based on the results of the key builder match. There's no direct link between the HCM and the scoped route resources (seems like a problem with the API design).

So for consistency checking, I think that:
* `ScopedRouteConfiguration` resources should be checked to ensure they name a valid `RouteConfiguration`
* the name in the HCM `ScopedRoutes` field should not be checked
* if HCM enables SRDS, then at least one `ScopedRouteConfiguration` must be present
* additional `ScopedRouteConfiguration` resources are allowed

jpeach:
When setting the `ScopedRoutes` field in a `HttpConnectionManager`
resouce, the name provided in the `ScopedRoutes` field is not the
name of a `ScopedRouteConfiguration` xDS resource, it's just a name.
When scoped routing is being used, the control plane may provide any
number of `ScopedRouteConfiguration` resouces, which are not linked to
any specific `HttpConnectionManager`.

This fixes #517.

### DIFF
--- a/pkg/cache/v3/resource.go
+++ b/pkg/cache/v3/resource.go
@@ -26,7 +26,6 @@ import (
 	endpoint "github.com/envoyproxy/go-control-plane/envoy/config/endpoint/v3"
 	listener "github.com/envoyproxy/go-control-plane/envoy/config/listener/v3"
 	route "github.com/envoyproxy/go-control-plane/envoy/config/route/v3"
-	hcm "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/network/http_connection_manager/v3"
 	auth "github.com/envoyproxy/go-control-plane/envoy/extensions/transport_sockets/tls/v3"
 	runtime "github.com/envoyproxy/go-control-plane/envoy/service/runtime/v3"
 	"github.com/envoyproxy/go-control-plane/pkg/cache/types"
@@ -197,10 +196,9 @@ func getClusterReferences(src *cluster.Cluster, out map[resource.Type]map[string
 
 // HTTP listeners will either reference ScopedRoutes or Routes.
 func getListenerReferences(src *listener.Listener, out map[resource.Type]map[string]bool) {
-	scopedRoutes := map[string]bool{}
 	routes := map[string]bool{}
 
-	// extract route configuration names from HTTP connection manager
+	// Extract route configuration names from HTTP connection manager.
 	for _, chain := range src.FilterChains {
 		for _, filter := range chain.Filters {
 			if filter.Name != wellknown.HTTPConnectionManager {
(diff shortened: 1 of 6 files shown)


jpeach:
@mtagstyle Are you able to review this?

mtagstyle:
Thanks for the clarification in the other thread. I'll take a look at this tonight.

alecholmez:
This looks great, I'll wait for @mtagstyle's review but you have an approval from my end

pkg/cache/v3/resource.go
```diff
@@ -212,42 +210,18 @@ func getListenerReferences(src *listener.Listener, out map[resource.Type]map[str
...
+				routes[name] = true
 			}
-		}
-	}
 
-	if len(scopedRoutes) > 0 {
-		if _, ok := out[resource.ScopedRouteType]; !ok {
-			out[resource.ScopedRouteType] = map[string]bool{}
+			// If the scoped route mapping is embedded, add the referenced route resource names.
+			for _, s := range config.GetScopedRoutes().GetScopedRouteConfigurationsList().GetScopedRouteConfigurations() {
```
mtagstyle, line 219: Should we be nil checking each call? 

Is there potential for GetScopedRoutes() to return nil? 
Is there potential for GetScopedRouteConfigurationsList() to return nil?

pkg/cache/v3/resource_test.go
```diff
@@ -18,6 +18,8 @@ import (
 	"reflect"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
```
mtagstyle, line 21: Thanks for bringing this in. It will greatly simplify all the assertions

pkg/cache/v3/resource_test.go
```diff
@@ -129,8 +121,8 @@ func TestGetResourceReferences(t *testing.T) {
 			out: map[rsrc.Type]map[string]bool{rsrc.EndpointType: {"test": true}},
 		},
 		{
-			in:  resource.MakeScopedRouteHTTPListener(resource.Xds, listenerName, 80, scopedRouteName),
-			out: map[rsrc.Type]map[string]bool{rsrc.ScopedRouteType: {scopedRouteName: true}},
+			in:  resource.MakeScopedRouteHTTPListener(resource.Xds, listenerName, 80),
```
mtagstyle, line 124: Should we add a test case to check the embedded routes case? (I.e. MakeScopedRouteHTTPListenerForRoutes)

pkg/cache/v3/snapshot_test.go
```diff
@@ -98,7 +98,7 @@ func TestScopedRouteListenerWithScopedRouteOnlyIsInconsistent(t *testing.T) {
 func TestScopedRouteListenerWithScopedRouteAndRouteIsConsistent(t *testing.T) {
 	snap, _ := cache.NewSnapshot(version, map[rsrc.Type][]types.Resource{
 		rsrc.ListenerType: {
-			resource.MakeScopedRouteHTTPListener(resource.Xds, "listener0", 80, "scopedRoute0"),
+			resource.MakeScopedRouteHTTPListener(resource.Xds, "listener0", 80),
```
mtagstyle, line 101: I guess now, is it possible to have Scoped Routes. But have no HTTP Listeners in the filter chain using SRDS?

mtagstyle (changes requested):
Thanks for getting to this before I did. Added some comments.

pkg/cache/v3/resource.go
jpeach: protobuf generates the `Get` accessors to gracefully handle nil objects. That's convenient for this kind of code because if any of the fields in the chain of references is nil, we will just end up ranging over a nil slice (which is a no-op).

pkg/cache/v3/snapshot_test.go
jpeach: Yeh, the reverse is also possible. I thought about implementing that, but balked at having to iterate and unpack the filters a second time and make a special case check for scoped routes. WDYT @alecholmez, is it worth it?

pkg/cache/v3/resource_test.go
jpeach: Yeh we might as well. Added this.

mtagstyle (approved):
LGTM. Just wanted that missing test case in.

pkg/cache/v3/snapshot_test.go
alecholmez: This is probably good for now, let's definitely note and potentially make an issue if that's something worth implementing down the road

alecholmez (approved):

PR #520, opened 2021-12-03
docs: rework current repo docs

### LINKED ISSUE #513: Help with cache.NewSnapshot breaking change and outdated documentation
I'm currently on 0.9.9 and looking at going to 0.10.0 and it looks like a breaking change was introduced. It also appears that this change was accidentally omitted from the changelog, unless I missed it?

It appears that the cache.NewSnapshot function was changed to take a different set of arguments however all existing documentation still refers to the old function arguments. 

Can anyone help with an upgrade example?

alecholmez:

closes #513


alecholmez:
@snowp just a quick fix up for docs here

jpeach:
@alecholmez WDYT about moving this to the `doc.go` in the relevant packages, and moving the code snippets into [testable examples](https://go.dev/blog/examples)?

alecholmez:
@jpeach I think that would be a  great idea, it looks like most people are using those so maybe we can push the changes through but then also do that? Honestly I don't like these docs since they require updating often especially if we make API changes. IMO looking at the tests are better anyways :) but I do allude to the code example quite a lot to minimize how much I need to change these docs

jpeach (approved):
This looks pretty reasonable to me.

alecholmez:
@snowp can you give the stamp of approval over here? It looks like James doesn't have repo control yet so his approval didn't allow a merge

snowp (approved):

docs/README.md
```diff
@@ -1,80 +1,12 @@
...
 
-Below lies a list of resources that may be helpful to those looking to understand the go-control-plane API. The aim of these artifacts is to provide enough knowledge and understanding to newcomers and users who wish to use this API within their own codebases to implement an xDS compliant control-plane.
+Below lies a list of resources that may be helpful to those looking to understand the go-control-plane API.
 
-## Snapshot Cache
+## Implementations
 The following guides may be helpful on how to use go-control-plane's Snapshot Cache:
-- [Snapshot.md](cache/Snapshot.md)
-- [Server.md](cache/Server.md)
+- [cache.md](snapshot.md)
```
sunjayBhatia, line 7: nit: snapshot.md as link text?

alecholmez: Ah oops thank you, it should be `cache.md`

sunjayBhatia: I think this was merged without updating this change, submitted https://github.com/envoyproxy/go-control-plane/pull/535

PR #533, opened 2021-12-21
Update lint to use the bugs preset.

jpeach:
Update golangci-lint to the latest version and enable the bugs preset
to run the collection of linters that are intended to catch code bugs.


alecholmez (approved):
Thanks this looks great

PR #536, opened 2022-01-04
Update the Go version to 1.16.

### LINKED ISSUE #532: Update go build version to something current
In #272, `go.mod` was downgraded to Go 1.11. While that was an unsupported release at the time, it's still unsupported. Per the Go [release policy](https://go.dev/doc/devel/release), the last 2 major releases are supported, which at time of writing puts us on 1.17 or 1.16.

The downgrade was done to support [grpcurl](https://github.com/fullstorydev/grpcurl), which I guess ended up taking a go-control-plane dependency via gRPC. grpcurl is on Go 1.15 now, so that seems like a safe minimum. My preference would still be to use oldest supported Go release.

jpeach:
Update the Go version directive to 1.16 since that is currently the
oldest supported go release.

This fixes #532.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ feedback, we might decided to revisit this aspect at a later point in time.
 
 ## Requirements
 
-1. Go 1.11+
+1. Go 1.16+
 
 ## Quick start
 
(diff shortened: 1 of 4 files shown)


alecholmez (approved):
Thanks for this!

PR #537, opened 2022-01-07
Added envoy.filters.listener.original_src

hzxuzhonghu:


sunjayBhatia:
Seems reasonable but @alecholmez @jpeach do we have any policy on what goes in the well known filters list?

alecholmez:
@sunjayBhatia i think this is fine. As long as it’s an upstream filter I’m cool with it

sunjayBhatia (approved):

PR #543, opened 2022-03-04
linear: be aware of stream state when responding (#540)

### LINKED ISSUE #540: In LinearCache, respond can't tell the caller is from update or delete.
UpdateResource and DeleteResource will call notifyAll once it has changes.  And It respond the resource that has been changed.
https://github.com/envoyproxy/go-control-plane/blob/4455b6323de455adf8d0c8afeb3fa5a90652555c/pkg/cache/v3/linear.go#L140-L151

However, the xds-client get the response. It would remove the resources that aren't in the response from local cache.
https://github.com/grpc/grpc-go/blob/011544f72939c85397b0e24378280e6075061cb1/xds/internal/xdsclient/pubsub/update.go#L227-L240 

Thus, a DeleteResource has been called, the xds-server would respond the deleted resource. But the resource has been deleted, actually it's empty. The xds-client remove the other resource not in the response from its local cache. it causes the grpc-client is unable to connect to other endpoints that haven't been deleted.
https://github.com/envoyproxy/go-control-plane/blob/4455b6323de455adf8d0c8afeb3fa5a90652555c/pkg/cache/v3/linear.go#L126-L131

fscnick:
To resolve #540, keep the `StreamState` in `watches`.

The sotw could calculate the required resource to respond when there is a modified resource.

dependency: #531


fscnick:
@alecholmez Would you mind helping me to approve this workflow?

fscnick:
not stale

fscnick:
not stale

fscnick:
not stale

fscnick:
not stale

fscnick:
not stale

fscnick:
not stale

fscnick:
not stale

PR #544, opened 2022-03-07
sotw: Ordered ADS mode

alecholmez:
introduce optional resource ordering for guaranteed eventual consistency in ADS mode.

### DIFF
--- a/pkg/integration/ttl_integration_test.go
+++ b/pkg/integration/ttl_integration_test.go
@@ -31,14 +31,11 @@ func (log logger) Warnf(format string, args ...interface{})  { log.t.Logf(format
 func (log logger) Errorf(format string, args ...interface{}) { log.t.Logf(format, args...) }
 
 func TestTTLResponse(t *testing.T) {
-
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
 	snapshotCache := cache.NewSnapshotCacheWithHeartbeating(ctx, false, cache.IDHash{}, logger{t: t}, time.Second)
-
 	server := server.NewServer(ctx, snapshotCache, nil)
-
 	grpcServer := grpc.NewServer()
 	endpointservice.RegisterEndpointDiscoveryServiceServer(grpcServer, server)
 
@@ -52,8 +49,8 @@ func TestTTLResponse(t *testing.T) {
 
 	conn, err := grpc.Dial(":9999", grpc.WithInsecure())
 	assert.NoError(t, err)
-	client := endpointservice.NewEndpointDiscoveryServiceClient(conn)
 
+	client := endpointservice.NewEndpointDiscoveryServiceClient(conn)
 	sclient, err := client.StreamEndpoints(ctx)
 	assert.NoError(t, err)
 
@@ -74,16 +71,16 @@ func TestTTLResponse(t *testing.T) {
 			TTL:      &oneSecond,
 		}},
 	})
+
 	err = snapshotCache.SetSnapshot(context.Background(), "test", snap)
 	assert.NoError(t, err)
 
 	timeout := time.NewTimer(5 * time.Second)
-
 	awaitResponse := func() *envoy_service_discovery_v3.DiscoveryResponse {
 		t.Helper()
 		doneCh := make(chan *envoy_service_discovery_v3.DiscoveryResponse)
-		go func() {
 
+		go func() {
 			r, err := sclient.Recv()
 			assert.NoError(t, err)
 
--- a/pkg/server/config/config.go
+++ b/pkg/server/config/config.go
@@ -1,0 +1,25 @@
+package config
+
+// Opts for individual xDS implementations that can be
+// utilized through the functional opts pattern
+type Opts struct {
+	// If true respond to ADS requests with a guaranteed resource ordering
+	Ordered bool
+}
+
+func NewOpts() Opts {
+	return Opts{
+		Ordered: false,
+	}
+}
+
+// Each xDS implementation should implement their own functional opts.
+// It is recommended that config values be added in this package specifically,
+// but the individual opts functions should be in their respective
+// implementation package so the import looks like the following:
+//
+// `sotw.WithOrderedADS()`
+// `delta.WithOrderedADS()`
+//
+// this allows for easy inference as to which opt applies to what implementation.
+type XDSOption func(*Opts)
--- a/pkg/server/sotw/v3/ads.go
+++ b/pkg/server/sotw/v3/ads.go
@@ -1,0 +1,162 @@
+package sotw
+
+import (
+	"fmt"
+	"reflect"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
+	discovery "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
+	"github.com/envoyproxy/go-control-plane/pkg/cache/v3"
+	"github.com/envoyproxy/go-control-plane/pkg/resource/v3"
+	"github.com/envoyproxy/go-control-plane/pkg/server/stream/v3"
+)
+
+// process handles a bi-di stream request
+func (s *server) processADS(sw *streamWrapper, reqCh <-chan *discovery.DiscoveryRequest, defaultTypeURL string) error {
+	// process := func(resp cache.Response) error {
+	// 	nonce, err := send(resp)
+	// 	if err != nil {
+	// 		return err
+	// 	}
+	// 	typeUrl := resp.GetRequest().TypeUrl
+	// 	watches.nonces[typeUrl] = nonce
+	// 	return nil
+	// }
+
+	// processAllExcept := func(typeUrl string) error {
+	// 	for {
+	// 		select {
+	// 		case resp := <-values.responses:
+	// 			if resp.GetRequest().TypeUrl != typeUrl {
+	// 				if err := process(resp); err != nil {
+	// 					return err
+	// 				}
+	// 			}
+	// 		default:
+	// 			return nil
+	// 		}
+	// 	}
+	// }
+
+	// node may only be set on the first discovery request
+	var node = &core.Node{}
+
+	// This control loop strictly orders resources when running in ADS mode.
+	// It should be treated as a child process of the original process() loop
+	// and should return on close of stream or error. This will cause the cleanup routines
+	// in the parent process() loop to execute.
+	for {
+		fmt.Printf("%+v\n", sw.watches.cases)
+		index, value, ok := reflect.Select(sw.watches.cases)
+		fmt.Println(index)
+		fmt.Println(value)
+		fmt.Println(ok)
+
+		switch index {
+		case 0:
+			fmt.Println("recieved a Done() on the ADS code path.")
+			// ctx.Done() -> no further computation is needed
+			return nil
+		case 1:
+			fmt.Println("This is the ordered ADS code path request pipeline")
+			if !ok {
+				return nil
+			}
+
+			req := value.Interface().(*discovery.DiscoveryRequest)
+			if req == nil {
+				return status.Errorf(codes.Unavailable, "empty request")
+			}
+
+			// Only first request is guaranteed to hold node info so if it's missing, reassign.
+			if req.Node != nil {
+				node = req.Node
+			} else {
+				req.Node = node
+			}
+
+			// nonces can be reused across streams; we verify nonce only if nonce is not initialized
+			nonce := req.GetResponseNonce()
+
+			// type URL is required for ADS but is implicit for xDS
+			if defaultTypeURL == resource.AnyType {
+				if req.TypeUrl == "" {
+					return status.Errorf(codes.InvalidArgument, "type URL is required for ADS")
+				}
+			} else if req.TypeUrl == "" {
+				req.TypeUrl = defaultTypeURL
+			}
+
+			if s.callbacks != nil {
+				if err := s.callbacks.OnStreamRequest(sw.ID, req); err != nil {
+					return err
+				}
+			}
+
+			// if lastResponse, ok := lastDiscoveryResponses[req.TypeUrl]; ok {
+			// 	if lastResponse.nonce == "" || lastResponse.nonce == nonce {
+			// 		// Let's record Resource names that a client has received.
+			// 		streamState.SetKnownResourceNames(req.TypeUrl, lastResponse.resources)
+			// 	}
+			// }
+
+			typeURL := req.GetTypeUrl()
+			responder := make(chan cache.Response, 1)
+			if w, ok := sw.watches.responders[typeURL]; ok {
+				// We've found a pre-existing watch, lets check and update if needed.
+				// If these requirements aren't satisfied, leave an open watch.
+				if w.nonce == "" || w.nonce == nonce {
+					w.close()
+
+					sw.watches.addWatch(typeURL, &watch{
+						cancel:   s.cache.CreateWatch(req, stream.StreamState{}, responder),
+						response: responder,
+					})
+				}
+			} else {
+				// No pre-existing watch exists, let's create one.
+				// We need to precompute the watches first then open a watch in the cache.
+				sw.watches.addWatch(typeURL, &watch{
+					cancel:   s.cache.CreateWatch(req, stream.StreamState{}, responder),
+					response: responder,
+				})
+			}
+
+			// Recompute the dynamic select cases for this stream.
+			sw.watches.recompute(s.ctx, reqCh)
+		default:
+			fmt.Println("This is the ordered ADS code path default pipeline")
+
+			// Channel n -> these are the dynamic list of responders that correspond to the stream request typeURL
+			if !ok {
+				// Receiver channel was closed. TODO(jpeach): probably cancel the watch or something?
+				return status.Errorf(codes.Unavailable, "resource watch %d -> failed", index)
+			}
+
+			res := value.Interface().(cache.Response)
+			nonce, err := sw.send(res)
+			if err != nil {
+				return err
+			}
+
+			sw.watches.responders[res.GetRequest().TypeUrl].nonce = nonce
+		}
+	}
+}
+
+// typeUrl := req.TypeUrl
+// responseNonce, seen := values.nonces[typeUrl]
+// if !seen || responseNonce == nonce {
+// 	if cancel, seen := values.cancellations[typeUrl]; seen {
+// 		if cancel != nil {
+// 			cancel()
+// 		}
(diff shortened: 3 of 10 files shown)


alecholmez:
This PR branches off the work found here: https://github.com/envoyproxy/go-control-plane/pull/461 and brings it up to speed with the current state of go-control-plane. We've found quite a few resource ordering issues in upstream xDS conformance test suites and this will hopefully provide a remedy to some of those problems. 

I tried to introduce the change in a non intrusive way, if no `opts` are provided to the server everything will execute as before. If the new opt `sotw.WithOrderedADS()` is passed to a server, the resources will be in guaranteed ordering as in the prior art PR. This provides guaranteed eventual consistency which we did not have in the previous implementation but gives up some of the pressure relief from immediate response sending.

pkg/integration/ttl_integration_test.go
```diff
@@ -31,14 +31,11 @@ func (log logger) Warnf(format string, args ...interface{})  { log.t.Logf(format
 func (log logger) Errorf(format string, args ...interface{}) { log.t.Logf(format, args...) }
 
 func TestTTLResponse(t *testing.T) {
-
```
alecholmez, line 34: Ran this file through a formatter hence these changes.

alecholmez:
Currently investigating the integration test failure.... seems to me that the server gets stuck somewhere and disconnects

pkg/cache/types/types.go
```diff
@@ -35,8 +35,8 @@ type ResponseType int
 const (
 	Endpoint ResponseType = iota
 	Cluster
-	Route
 	ScopedRoute
+	Route
```
sunjayBhatia, line 39: this is nitpicky but could this technically be a breaking change? in case someone had comparisons on this type, seems unfortunate but maybe

sunjayBhatia, line 39: this is nitpicky but could this technically be a breaking change? in case someone had comparisons on this type, seems unfortunate but maybe

pkg/cache/types/types.go
```diff
@@ -35,8 +35,8 @@ type ResponseType int
 const (
 	Endpoint ResponseType = iota
 	Cluster
```
sunjayBhatia, line 37: should `Cluster` sort before `Endpoint`?

sunjayBhatia, line 37: should `Cluster` sort before `Endpoint`?

pkg/cache/v3/order_test.go
```diff
@@ -0,0 +1,49 @@
...
+		{
+			ID:      5,
+			TypeURL: resource.ScopedRouteType,
+		},
+	}
+
+	orderedKeys := unorderedKeys
+	sort.Sort(orderedKeys)
+
+	if !sort.IsSorted(orderedKeys) {
```
sunjayBhatia, line 38: i dont think this really works to test it this way, its just checking itself with itself, right now I get this order when i print `orderedKeys` and the test passes:

```
{4 type.googleapis.com/envoy.config.listener.v3.Listener}
{3 type.googleapis.com/envoy.config.route.v3.RouteConfiguration}
{5 type.googleapis.com/envoy.config.route.v3.ScopedRouteConfiguration}
{2 type.googleapis.com/envoy.config.cluster.v3.Cluster}
{1 type.googleapis.com/envoy.config.endpoint.v3.ClusterLoadAssignment}
```

pkg/cache/types/types.go
alecholmez: ^ this I have no idea, I had it in the original order and that was causing the bug that @maplebeats was seeing

pkg/cache/types/types.go
alecholmez: Ah yea that's a good point... hm

pkg/cache/v3/order_test.go
alecholmez: Yea so this test is pretty much pointless. It was really just to show that we can sort the keys in the order they're expected, LDS before RDS and SRDS, and CDS before EDS. I guess LDS before CDS doesn't matter though

pkg/cache/types/types.go
maplebeats: cluster should send before endpoints in cache server, otherwise cds will keep waiting for eds response.

alecholmez:
Haven’t had much bandwidth to work this recently will get back to it soon.

pkg/cache/types/types.go
alecholmez: I think the solution to this imo is just bump the released version. We can note this in the changelog but it's a critical part of the ordering problem.

pkg/cache/v3/order_test.go
alecholmez: Also want to note here that when I started sending scoped routes before routes envoy began to complain. I wasn't sure if that was a bug so I moved it back and all seemed to be okay. I think the most important ordering here is that LDS and CDS send first so envoy can build its config tree properly.

marcosrmendezthd:
any progress on this?

alecholmez:
Need to fix the integration tests and then this should be ready for a rereview

haoruolei:
Hi do we have updates on this PR?

alecholmez:
Tests pass now just working through lint issue

pkg/cache/v3/status.go
```diff
@@ -155,3 +159,26 @@ func (info *statusInfo) setDeltaResponseWatch(id int64, drw DeltaResponseWatch)
...
+		info.orderedWatches[index] = key{
+			ID:      id,
+			TypeURL: watch.Request.TypeUrl,
+		}
+		index++
+	}
+
+	// Sort our list which we can use in the SetSnapshot functions.
+	// This is only run when we enable ADS on the cache.
+	if order {
```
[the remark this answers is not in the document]
alecholmez: One of the original algorithm designs was to allow 2 modes in ADS:

- A slower ordered mode
- A faster unordered mode

The latter could be helpful in scenarios that the management server needs the least possible back pressure. I just followed the original algorithm but I think there is merit allowing users to pick the highest-bandwidth mode of operation.

pkg/cache/v3/status.go
```diff
@@ -155,3 +159,26 @@ func (info *statusInfo) setDeltaResponseWatch(id int64, drw DeltaResponseWatch)
 	defer info.mu.Unlock()
 	info.deltaWatches[id] = drw
 }
+
+// orderResponseWatches will track a list of watch keys and order them if
+// true is passed.
+func (info *statusInfo) orderResponseWatches(order bool) {
+	info.orderedWatches = make(keys, len(info.watches))
+
+	// This runs in O(n) which could become problematic when we have an extremely high watch count.
```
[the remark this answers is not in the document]
alecholmez: You're right the sort is really fast but building that indexed ordered list is worse case scenario `O(len(info.watches))`. Luckily inserting into a map is `O(1)`. Honestly I can probably delete this comment because this operation really isn't that expensive.

pkg/cache/v3/status.go
```diff
@@ -155,3 +159,26 @@ func (info *statusInfo) setDeltaResponseWatch(id int64, drw DeltaResponseWatch)
...
+	// TODO(alec): revisit this and optimize for speed.
+	var index int
```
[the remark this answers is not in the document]
alecholmez: I tried to do that as an optimization. Append is kind of slow under the hood and pre-allocating/using the index is slightly faster.

I can undo that though if that's impacting readability.

pkg/cache/v3/simple.go
```diff
@@ -232,20 +232,41 @@ func (cache *snapshotCache) SetSnapshot(ctx context.Context, node string, snapsh
...
 				delete(info.watches, id)
 			}
+			return nil
+		}
+
+		// If ADS is enabled we need to order response watches so we guarantee
+		// sending them in the correct order. Go's default implementation
+		// of maps are randomized order when ranged over.
+		if cache.ads {
+			info.orderResponseWatches(cache.ads)
```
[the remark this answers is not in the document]
alecholmez: Nice catch thank you

pkg/server/sotw/v3/ads.go
```diff
@@ -0,0 +1,135 @@
...
+			// Use the multiplexed channel for new watches.
+			responder := sw.watches.responders[resource.AnyType].response
+			if w, ok := sw.watches.responders[typeURL]; ok {
+				// We've found a pre-existing watch, lets check and update if needed.
+				// If these requirements aren't satisfied, leave an open watch.
+				if w.nonce == "" || w.nonce == nonce {
+					w.close()
+
+					// Only process if we have an existing watch otherwise go ahead and create.
+					if err := processAllExcept(typeURL); err != nil {
```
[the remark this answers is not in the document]
alecholmez: @valerian-roche good call I'll drop a comment there but the original algorithm as outlined here:

https://github.com/envoyproxy/go-control-plane/pull/461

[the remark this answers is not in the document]
alecholmez: It's been quite a long time since I've ported that logic over but if I remember correctly we don't actually drop the responses, they just get requeued and eventually are sent once the enforced order has been resolved. (CDS before EDS, LDS before RDS, etc...)

alecholmez:
@arkodg Yup we'll need to port this over

pkg/server/sotw/v3/ads.go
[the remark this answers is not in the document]
valerian-roche: From what I remember, here it is dequeuing all pre-existing responses in the channel. If of the type it seems to simply be discarded. I'm afraid of the impact if the version in the request is not changed, as the control-plane will consider there is no change and not send new resources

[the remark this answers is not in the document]
alecholmez: Hm maybe we can write a test to cover that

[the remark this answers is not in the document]
alecholmez: Quick update, the tests actually do cover partially sent snapshots!

pkg/server/sotw/v3/server.go
```diff
@@ -61,177 +72,78 @@ type server struct {
...
-		lastResponse := lastDiscoveryResponse{
-			nonce:     out.Nonce,
-			resources: make(map[string]struct{}),
-		}
-		for _, r := range resp.GetRequest().ResourceNames {
-			lastResponse.resources[r] = struct{}{}
-		}
-		lastDiscoveryResponses[resp.GetRequest().TypeUrl] = lastResponse
+	// increment nonce and convert it to base10
+	out.Nonce = strconv.FormatInt(atomic.AddInt64(&s.nonce, 1), 10)
```
valerian-roche, line 111: Minor: do we need this atomic increase here? The rest of this method is not synchronized anyway

valerian-roche (approved):

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/envoyproxy/go-control-plane
 
-go 1.17
+go 1.19
```
jpeach, line 3: Can we make this change in a separate PR, since it affects downstream projects?

pkg/cache/types/types.go
```diff
@@ -41,8 +41,8 @@ type ResponseType int
 const (
```
jpeach, line 41: Let's add a big fat comment to tell people that the ordering of these enum values matters, linking to the ADS reference if possible.

pkg/cache/v3/order_test.go
jpeach, line 38:
```suggestion
assert.True(t, sort.IsSorted(orderedKeys))
assert.NotEqual(t, unorderedKeys, orderedKeys)
```

This shows that the sorting does something, then you can add a canonically sorted expected value as well.

pkg/log/log_test.go
```diff
@@ -22,13 +22,11 @@ import (
...
-
 	xdsLogger := LoggerFuncs{
-		DebugFunc: logger.Printf,
-		InfoFunc:  logger.Printf,
-		WarnFunc:  logger.Printf,
-		ErrorFunc: logger.Printf,
+		DebugFunc: log.Printf,
+		InfoFunc:  log.Printf,
+		WarnFunc:  log.Printf,
+		ErrorFunc: log.Printf,
```
jpeach, line 29: Why do we need this change? Can we make a separate PR?

pkg/integration/ttl_integration_test.go
jpeach: Can we make a separate PR with this, and just land it?

pkg/server/config/config.go
```diff
@@ -0,0 +1,25 @@
+package config
+
+// Opts for individual xDS implementations that can be
+// utilized through the functional opts pattern
```
jpeach, line 4: general nit: check that doc comments are correctly capitalized and punctuated.

pkg/server/config/config.go
```diff
@@ -0,0 +1,25 @@
...
+type Opts struct {
+	// If true respond to ADS requests with a guaranteed resource ordering
+	Ordered bool
```
jpeach, line 7: TBH, I haven't looked at ordering for a long time. My recollection was that for ADS, the client was supposed to make requests in the order that was correct for it? If that's the case, do we really need to order anything in the server?

My other query is that, if ordering is a correctess issue, then shouldn't we be correct by default?

pkg/server/sotw/v3/ads.go
```diff
@@ -0,0 +1,137 @@
...
+				}
+			}
+
+			typeURL := req.GetTypeUrl()
+			// Use the multiplexed channel for new watches.
+			responder := sw.watches.responders[resource.AnyType].response
+			if w, ok := sw.watches.responders[typeURL]; ok {
+				// We've found a pre-existing watch, lets check and update if needed.
+				// If these requirements aren't satisfied, leave an open watch.
+				if w.nonce == "" || w.nonce == nonce {
```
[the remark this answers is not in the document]
alecholmez: Um that's a good question, I left this as the previous behavior but maybe worth an issue?

pkg/server/sotw/v3/server.go
alecholmez: I think we do need this since N number of "sends" can increment the nonce counter which is shared. I could be wrong though I ported that from the previous code

pkg/server/sotw/v3/xds.go
```diff
@@ -0,0 +1,163 @@
...
+				// When using ADS we need to order responses.
+				// This is guaranteed in the xDS protocol specification
+				// as ADS is required to be eventually consistent.
+				// More details can be found here if interested:
+				// https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#eventual-consistency-considerations
+				if s.opts.Ordered {
+					// send our first request on the stream again so it doesn't get
+					// lost in processing on the new control loop
+					go func() {
+						reqCh <- req
```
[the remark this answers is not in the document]
alecholmez: Yea I did think about that. I couldn't really come up with anything smarter than this since we need to shuttle the request down into the processors, would be open to suggestions! If this does cause a bug we can patch after the fact too

pkg/server/stream/v3/stream.go
```diff
@@ -71,56 +93,54 @@ func (s *StreamState) WatchesResources(resourceNames map[string]struct{}) bool {
...
-}
-
+// IsWildcard returns whether or not an xDS client requested in wildcard mode on the initial request.
 func (s *StreamState) IsWildcard() bool {
 	return s.wildcard
 }
 
+// IsOrdered returns whether or not the current stream should run as an ordered ADS stream.
+// This means less back-pressure relief but a guarantee of correct discovery response order.
+func (s *StreamState) IsOrdered(ordered bool) bool {
```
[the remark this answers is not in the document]
alecholmez: Ah yea I think I messed that up. I'll fix

go.mod
alecholmez: Yup I actually just opened this:
https://github.com/envoyproxy/go-control-plane/pull/714

pkg/log/log_test.go
alecholmez: I think the linter was complaining about this? Honestly can't remember haha

pkg/server/config/config.go
alecholmez: That's what I thought, the bug this closes out shows that envoy sometimes requests things out of order even in ADS. 

IMO for ADS we always want to be ordered so yes. If I recall correctly when a user configures the server to run in ADS mode it'll be ordered by default (but you can override that).

go.mod
alecholmez: Reverted back to 1.17 for this PR so we can get it in before the multi-module release.

pkg/server/sotw/v3/ads.go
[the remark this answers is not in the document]
valerian-roche: Yes I think an issue would be good to avoid missing this

pkg/server/sotw/v3/server.go
valerian-roche: My understanding is that all the "sends" run in the same goroutine on a stream, so I do not expect the need to synchronize, but might be wrong

pkg/cache/v3/status.go
```diff
@@ -65,7 +66,9 @@ type statusInfo struct {
 	node *core.Node
 
 	// watches are indexed channels for the response watches and the original requests.
-	watches map[int64]ResponseWatch
+	// watches map[int64]ResponseWatch
```
valerian-roche, line 69: Minor: comment to cleanup?

pkg/server/sotw/v3/xds.go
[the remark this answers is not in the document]
valerian-roche: For now I think adding a comment about the risk might be useful.
Longer term I'd like to do more work on improving the state machine and aligning between delta and sotw, and that work will hopefully cover this issue

sample/bootstrap-xds.yaml
```diff
@@ -53,16 +53,16 @@ static_resources:
...
-        rtds_config:
-          resource_api_version: V3
-          api_config_source:
-            transport_api_version: V3
-            api_type: GRPC
-            grpc_services:
-              envoy_grpc:
-                cluster_name: xds_cluster
-        name: runtime-0
+# layered_runtime:
```
valerian-roche, line 56: Is there a reason for this removal? Is there an issue with a new dependency version?

valerian-roche (approved):
Nothing blocking, just a comment for potential order issue would be nice

pkg/cache/v3/status.go
alecholmez: Oops I'll clean that up

pkg/server/sotw/v3/xds.go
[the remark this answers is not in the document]
alecholmez: Good call I can add a quick comment then we can get this merged

valerian-roche (approved):

Issue #423, opened 2021-04-20
Incremental snapshot update support for DeltaXDS


snowp:
The work in progress to add support for DeltaXDS relies on computing per resource versions from a snapshot every time it is updated, pushing updates if the computed version differs from the computed versions of the previous snapshot. This ensures that we only update the resources that change, but requires substantial computation per snapshot update.

We should support 1) specifying the per-resource versions explicitly and 2) partial snapshot updates. This would enable updating a small subset m of a snapshot of size n in O(m) instead of O(n), which would provide a huge improvement when dealing with a large number of resources. 

Whether this would be compatible with the current SimpleCache implementation will need some investigation, it might warrant a new cache implementation that provides a more incremental API. 

FYI @alecholmez (no action required)

PR #547, opened 2022-03-17
Linear cache: Avoid linear time iterations on all resources when applying delta updates. No longer build the versionMap when not using delta watches

valerian-roche:
The linear cache is currently building the versionMap every time it is receiving any update
This is quite expensive currently in two cases:
* When not using delta watches, it is never used but constantly maintained. It is very expensive when using SetResources 
* When using delta watches and using delta updates (UpdateResource/DeleteResource/UpdateResources), it is iterating on the entire resource set instead of iterating on modified objects

Also adds a constant time accessor to the nb of entries in the cache as GetResources is no longer constant time

Current profile in our implementation (based on updates usually touching less than 1% of the resources)
* Lot of time spent in GetResources (mostly called for statistics on the nb of entries)
* updateVersionMap takes a large share of the time. We do need it as using delta watches, but most users won't if using sotw
image

### DIFF
--- a/pkg/cache/v3/linear.go
+++ b/pkg/cache/v3/linear.go
@@ -371,8 +371,9 @@ func (cache *LinearCache) CreateDeltaWatch(request *DeltaRequest, state stream.S
 	defer cache.mu.Unlock()
 
 	if cache.versionMap == nil {
-		// If we had no delta watch before, we need to build the version map for the first time.
-		// The map will not be removed when the last delta watch is removed to avoid rebuilding it constantly when only a few delta watches are applicable.
+		// If we had no previously open delta watches, we need to build the version map for the first time.
+		// The version map will not be destroyed when the last delta watch is removed.
+		// This avoids constantly rebuilding when only a few delta watches are open.
 		modified := map[string]struct{}{}
 		for name := range cache.resources {
 			modified[name] = struct{}{}
--- a/pkg/cache/v3/linear.go
+++ b/pkg/cache/v3/linear.go
@@ -50,7 +50,7 @@ type LinearCache struct {
 	deltaWatches map[int64]DeltaResponseWatch
 	// Continuously incremented counter used to index delta watches.
 	deltaWatchCount int64
-	// versionMap holds the current hash map of all resources in the cache.
+	// versionMap holds the current hash map of all resources in the cache when delta watches are present.
 	// versionMap is only to be used with delta xDS.
 	versionMap map[string]string
 	// Continuously incremented version.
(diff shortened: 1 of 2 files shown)


pkg/cache/v3/linear.go
```diff
@@ -366,6 +370,18 @@ func (cache *LinearCache) CreateDeltaWatch(request *DeltaRequest, state stream.S
 	cache.mu.Lock()
 	defer cache.mu.Unlock()
 
+	if cache.versionMap == nil {
+		// If we had no delta watch before, we need to build the version map for the first time.
```
alecholmez, line 374:
```suggestion
		// If we had no previously open delta watches, we need to build the version map for the first time.
```

pkg/cache/v3/linear.go
```diff
@@ -366,6 +370,18 @@ func (cache *LinearCache) CreateDeltaWatch(request *DeltaRequest, state stream.S
...
+		// The map will not be removed when the last delta watch is removed to avoid rebuilding it constantly when only a few delta watches are applicable.
```
alecholmez, line 375:
```suggestion
		// The version map will not be destroyed when the last delta watch is removed. 
		// This avoids constantly rebuilding when only a few delta watches are open.
```

alecholmez (changes requested):
Overall this looks good to me. Just a few nits in the comments but other than that seems like a solid performance increase. One of the things I actually need to get around to is porting over a change to the simple cache where we only hash on modified resources instead of the entire set. Nice addition for the linear cache

pkg/cache/v3/linear.go
valerian-roche: Updated to match

pkg/cache/v3/linear.go
valerian-roche: Updated to match

valerian-roche:
> Overall this looks good to me. Just a few nits in the comments but other than that seems like a solid performance increase. One of the things I actually need to get around to is porting over a change to the simple cache where we only hash on modified resources instead of the entire set. Nice addition for the linear cache

Not sure how the snapshot cache can be optimized on that. In most cases the user will rebuild the entire snapshot for the node. On the linear cache side this is really bad today to build the cache map with SetResources (in our production systems we reached ~1s spent just on it) so it can probably become quite bad for snapshot also
Are you planning on supporting delta snapshot updates on simple cache? That'd be very useful for us (e.g. if we want to use runtime flags in the same simple cache)

alecholmez:
@valerian-roche We would like to add support for incremental snapshot updates with delta. We have this open: https://github.com/envoyproxy/go-control-plane/issues/423

I haven't given much thought to it yet but I imagine that designing an API for per resource updates will significantly increase performance in the simple cache. We have a similar problem there where we're constantly building a big hash map regardless of what changed.

alecholmez (approved):

valerian-roche:
Definitely make sense to me to allow delta updates of snapshots, especially to no longer require to rebuild the entire snapshot to just update one resource 
Thanks for looking at improving that!

Issue #558, opened 2022-04-19
Support new envoy wildcard flows in go-control-plane


valerian-roche:
Envoy [1.21.0](https://www.envoyproxy.io/docs/envoy/latest/version_history/v1.21.0) brought a new [wildcard definition](https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#how-the-client-specifies-what-resources-to-return) (itself related to odcds)
While it seems that envoy will not use this new model out of odcds for now, the go-control-plane should support it as there is no strong commitment in the xds protocol, and to allow users to use odcds

A few specific considerations:
- a stream can now define subscribing to wildcard by sending "*" instead of nothing
- a stream can now subscribe or unsubscribe from wildcard while still referencing resources, and it expects the control-plane to keep track of explicitly subscribed resources 
- when unsubscribing from a specific resource while using a wildcard subscription, envoy [expects](https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#xds-protocol-unsubscribe) the control-plane to return the state of such resource

PR #559, opened 2022-04-19
[#558] Support dynamic wildcard subscription in delta-xds

valerian-roche:
Envoy 1.21.0 introduced a new definition for wildcard watches in the xds protocol
I provided more details in #558, but it impacts both the semantics and the behavior

This PR is implementing this flow for delta-xds (I will follow-up with a sotw later on)


valerian-roche:
Unittests are not provided yet, will add soon

valerian-roche:
This build error seems to be related to an envoy upstream update in the repo 14 days ago
Envoy is failing to add listeners with
```[2022-04-22 17:03:59.328][9990][warning][config] [source/common/config/grpc_subscription_impl.cc:126] gRPC config for type.googleapis.com/envoy.config.listener.v3.Listener rejected: Error adding/updating listener(s) listener-3: Didn't find a registered implementation for 'envoy.filters.http.router' with type URL: ''
listener-0: Didn't find a registered implementation for 'envoy.filters.http.router' with type URL: ''
listener-1: Didn't find a registered implementation for 'envoy.filters.http.router' with type URL: ''
listener-2: Didn't find a registered implementation for 'envoy.filters.http.router' with type URL: ''
```

alecholmez:
@valerian-roche any updates on upstream still failing around this?

valerian-roche:
I think I need to rebase, will do

valerian-roche:
@alecholmez Build is fixed after merging main

valerian-roche:
A first draft for sotw is available at https://github.com/valerian-roche/go-control-plane/pull/4 but won't be raised here until this one has reached a conclusion as it's building on it

pkg/cache/v3/delta.go
```diff
@@ -53,29 +53,26 @@ func createDeltaResponse(ctx context.Context, req *DeltaRequest, state stream.St
 		}
 
 		// Compute resources for removal
+		// The resource version can be set to "" here to trigger a removal even if never returned before
```
alecholmez, line 56: Is this true? Does a removal need to happen if we've never sent the resource before? The protocol spec is a bit vague around this so want to make sure we make a good decision here.

pkg/cache/v3/delta.go
```diff
@@ -53,29 +53,26 @@ func createDeltaResponse(ctx context.Context, req *DeltaRequest, state stream.St
...
+		for name := range state.GetSubscribedResourceNames() {
+			prevVersion, found := state.GetResourceVersions()[name]
 			if r, ok := resources.resourceMap[name]; ok {
 				nextVersion := resources.versionMap[name]
 				if prevVersion != nextVersion {
 					filtered = append(filtered, r)
 				}
 				nextVersionMap[name] = nextVersion
-			} else {
-				// We track non-existent resources for non-wildcard streams until the client explicitly unsubscribes from them.
```
alecholmez, line 72: Don't we still need to track these non-existent resources in non wildcard mode subscriptions?

pkg/server/stream/v3/stream.go
```diff
@@ -23,20 +23,44 @@ type DeltaStream interface {
 
 // StreamState will keep track of resource state per type on a stream.
 type StreamState struct { // nolint:golint,revive
-	// Indicates whether the original DeltaRequest was a wildcard LDS/RDS request.
+	// Indicates whether the stream currently has a wildcard watch
```
alecholmez, line 26: Let's make this comment explicit to delta since SOTW also uses this StreamState internally.

pkg/server/stream/v3/stream.go
```diff
@@ -23,20 +23,44 @@ type DeltaStream interface {
...
 
 	// knownResourceNames contains resource names that a client has received previously
 	knownResourceNames map[string]map[string]struct{}
 
-	// indicates whether the object has beed modified since its creation
+	// indicates whether the object has been modified since its creation
 	first bool
 }
 
+func (s *StreamState) GetSubscribedResourceNames() map[string]struct{} {
```
alecholmez, line 44: Can we add some doc strings to these functions? Let's be explicit about delta xds here as well.

pkg/server/delta/v3/server.go
```diff
@@ -210,17 +213,35 @@ func (s *server) DeltaStreamHandler(str stream.DeltaStream, typeURL string) erro
...
+			continue
+		}
+		sv[resource] = struct{}{}
 	}
 }
 
-// Unsubscriptions remove resources from the stream state to
-// indicate to the cache that we don't care about the resource anymore
-func (s *server) unsubscribe(resources []string, sv map[string]string) {
+// Unsubscriptions remove resources from the stream subscribed resources.
```
alecholmez, line 228:
```suggestion
// Unsubscriptions remove resources from the streams subscribed resource list.
```

pkg/server/delta/v3/server.go
```diff
@@ -210,17 +213,35 @@ func (s *server) DeltaStreamHandler(str stream.DeltaStream, typeURL string) erro
...
+// If explicitly unsubscribing from wildcard, the stream is updated and is now watching only subscribed resources.
+func (s *server) unsubscribe(resources []string, streamState *stream.StreamState) {
+	sv := streamState.GetSubscribedResourceNames()
 	for _, resource := range resources {
+		if resource == "*" {
+			streamState.SetWildcard(false)
+			continue
+		}
+		if _, ok := sv[resource]; ok && streamState.IsWildcard() {
+			// xds protocol specifically states that if a resource if unsubscribed while a wildcard watch is present,
```
alecholmez, line 238: Can we clean this comment block up?

pkg/server/delta/v3/server.go
```diff
@@ -160,14 +160,17 @@ func (s *server) processDelta(str stream.DeltaStream, reqCh <-chan *discovery.De
 			if !ok {
 				// Initialize the state of the stream.
 				// Since there was no previous state, we know we're handling the first request of this type
-				// so we set the initial resource versions if we have any, and also signal if this stream is in wildcard mode.
+				// so we set the initial resource versions if we have any.
+				// We also set the stream as wildcard based on its legacy meaning (no resource subscribed).
```
alecholmez, line 164: This is a bit misleading IMO. The legacy meaning of wildcard was ALL resources subcribed but we never kept them in the resources subscription list.

alecholmez (changes requested):
Just some minor cleanup. Overall it looks good though!

pkg/cache/v3/delta.go
valerian-roche: If the resource was never sent the version will not be set here 
The only case where it will happen now is when we want to explicitly trigger this "deletion", as done at https://github.com/envoyproxy/go-control-plane/pull/559/files#diff-5119f0079a6d464b74323939220b02d9a7aebcafd906e2e5be0161637afe5022R243

pkg/server/delta/v3/server.go
valerian-roche: I will rephrase. By no resource subscribed I meant none sent in `req.GetResourceNamesSubscribe()`

pkg/server/stream/v3/stream.go
valerian-roche: I added docstring for the three related methods

pkg/server/delta/v3/server.go
valerian-roche: I tried to make it clearer, though it is still so counter-intuitive to me I have a hard time formalizing it

pkg/cache/v3/delta.go
alecholmez: Ah understood. Glad we can clean up this workaround

alecholmez (commented):
Awesome thanks for these changes @valerian-roche.

PR #571, opened 2022-07-06
sotw: [#558] Support dynamic wildcard subscription in sotw-xds 

valerian-roche:
Akin to #559, this PR is adding support for the new envoy wildcard semantics in the context of the SOTW server
It is also doing significant work merging logic between the delta and sotw parts, especially for shared objects like StreamState where we currently have different members representing the same functional concern based on the server used

The PR is quite large mainly due to:
* large nb of tests added or updated
* merge of stream state semantics between sotw and delta, and no longer relying on request resources for wildcard status in sotw
* merge of accessors on stream state required aligning `map[string]bool` and `map[string]struct{}`. An implementation choice here was to align on `map[string]struct{}` to not alter delta as much as possible, but the revert can be done

I can rework it in multiple smaller PRs to handle some parts independently (mostly aligning the maps) if preferred by reviewers


alecholmez:
Ah this is awesome thanks @valerian-roche. I'll start the review process on this now.

PR #572, opened 2022-07-08
server: add Node to stream closed callbacks

### LINKED ISSUE #569: Proposal: Add Node to OnStreamClosed and OnDeltaStreamClosed of Callbacks
When we implement a control plane, we may need to watch resources like k8s Ingress/Gateway using k8s informer. To save cpu and memory, we only want to watch specific the resources we care about. Then we can start watching when we get a request and stop watching when the stream closes. With current [Callbacks](https://github.com/envoyproxy/go-control-plane/blob/main/pkg/server/v3/server.go#L62) interface, we can start watching in (the first) `OnStreamRequest`(take sotw for an example) using `Node` in `DiscoveryRequest`, but we can't stop watching in `OnStreamClosed` because the params of doesn't have the `Node`.

So I propose adding Node to `OnStreamClosed` and `OnDeltaStreamClosed`, like:
```
# sotw
type Callbacks interface {
	OnStreamClosed(int64, *corev3.Node)
}

# delta
type Callbacks interface {
	OnDeltaStreamClosed(int64, *corev3.Node)
}
```

Any idea about the proposal? Or maybe there are other solutions to achieve this?

If this is a good idea, I can send a PR for this. But this will be a breaking change :(

tony612:
Add Node to OnStreamClosed and OnDeltaStreamClosed of Callbacks

Fixes #569

### DIFF
--- a/docs/server.md
+++ b/docs/server.md
@@ -22,6 +22,7 @@ import (
 	"log"
 	"sync"
 
+	core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	discovery "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
 )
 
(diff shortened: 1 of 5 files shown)


tony612:
I didn't change CHANGELOG for the moment. Should I do that by myself? If so, I guess we should use a new Release(v0.10.4)?

tony612:
@alecholmez Could you help review this PR?

alecholmez:
No worry on bumping the changelog I'll have to do that when I cut a new release.

alecholmez (approved):

tony612:
Thanks!

PR #573, opened 2022-07-08
cache: fix empty context in response of linear cache

tony612:
Signed-off-by: Bing Han <h.bing612@gmail.com>

A small fix. Without this, `Ctx` in response of linear cache will be nil, which may be used in callbacks and cause problems.

Besides this, maybe we can support passing Context in some methods of `LinearCache` like what we do in `SnapshotCache`? This can be done in another PR(or another issue to discuss).


alecholmez (approved):

alecholmez:
@tony612 I think that would be helpful. We made that change in the `SnapshotCache` so people could cancel out of potentially long running functions. I'm sure there are some methods in linear that may benefit from this

PR #579, opened 2022-08-02
snapshot cache: fix heartbeat panic

danielhochman:
Encountering a panic where the snapshot has not been set yet but the cache is heartbeating and a watch comes in. Add a check on the snapshot existence for the node to avoid the panic.

Added a test to verify.

```
2022-08-02T15:18:27.520Z       DEBUG   v3/simple.go:391        open watch 1 for type.googleapis.com/envoy.service.runtime.v3.Runtime[rtds] from nodeID "test-cluster", version ""
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0xb3a163]

goroutine 7 [running]:
github.com/envoyproxy/go-control-plane/pkg/cache/v3.(*snapshotCache).sendHeartbeats(0xc0003a2180, {0xf26d10, 0xc000040030}, {0xc000496754, 0xc})
       /go/src/github.com/envoyproxy/go-control-plane/pkg/cache/v3/simple.go:193 +0x163
github.com/envoyproxy/go-control-plane/pkg/cache/v3.NewSnapshotCacheWithHeartbeating.func1()
       /go/src/github.com/envoyproxy/go-control-plane/pkg/cache/v3/simple.go:176 +0x154
created by github.com/envoyproxy/go-control-plane/pkg/cache/v3.NewSnapshotCacheWithHeartbeating
       /go/src/github.com/envoyproxy/go-control-plane/pkg/cache/v3/simple.go:167 +0x1a9
```

### DIFF
--- a/pkg/cache/v3/simple_test.go
+++ b/pkg/cache/v3/simple_test.go
@@ -604,3 +604,34 @@ func TestSnapshotSingleResourceFetch(t *testing.T) {
 		protocmp.Transform()),
 	)
 }
+
+func TestAvertPanicForWatchOnNonExistentSnapshot(t *testing.T) {
+	ctx := context.Background()
+	c := cache.NewSnapshotCacheWithHeartbeating(ctx, false, cache.IDHash{}, nil, time.Millisecond)
+
+	// Create watch.
+	req := &cache.Request{
+		Node:          &core.Node{Id: "test"},
+		ResourceNames: []string{"rtds"},
+		TypeUrl:       rsrc.RuntimeType,
+	}
+	ss := stream.NewStreamState(false, map[string]string{"cluster": "abcdef"})
+	responder := make(chan cache.Response)
+	c.CreateWatch(req, ss, responder)
+
+	go func() {
+		// Wait for at least one heartbeat to occur, then set snapshot.
+		time.Sleep(time.Millisecond * 5)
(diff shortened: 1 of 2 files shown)


sunjayBhatia (commented):
LGTM

probably never hit this case in other tests since we use a longer heartbeat interval and set a snapshot before it triggers!

pkg/cache/v3/simple_test.go
```diff
@@ -604,3 +604,34 @@ func TestSnapshotSingleResourceFetch(t *testing.T) {
...
+		// Wait for at least one heartbeat to occur, then set snapshot.
+		time.Sleep(time.Millisecond * 5)
+		srs := &singleResourceSnapshot{
+			version:  "version-one",
+			typeurl:  rsrc.RuntimeType,
+			name:     "one-second",
+			resource: durationpb.New(time.Second),
+		}
+		if err := c.SetSnapshot(ctx, "test", srs); err != nil {
+			panic(err)
```
dontlaugh, line 632: Should we prefer `t.Fatalf("..")` here?

sunjayBhatia: ah good catch, yeah agree with that change, should never error but good to not have raw panics

alecholmez (approved):

PR #584, opened 2022-08-25
Rework Cache interface

valerian-roche:
It has been raised that the addition of StreamState within the Cache interface is breaking the separation of concern between the server side managing the xds protocol itself and the cache which is only in charge of returning responses based on watches
PR https://github.com/envoyproxy/go-control-plane/pull/577 has been opened proposing to fully remove any client state from the CreateWatch interface

This PR is an alternate proposal based on a few concerns/aspects:
- one of the argument on removing it is that the sotw protocol is stateless. It actually is not, and is explicitly visible in this PR: the lastResponse latched value as well as the version hack when the resource list is changed very well show that it is stateful
- in cases outside of lds and cds, it is clearly stated that the control-plane is not expected to send back all resources when a single one change. This is highly critical in the case of ads with eds, as there is a likelihood of both having a lot of resources and a high churn. The implementation sending everything each time is technically tolerated, but is pushing on the data-plane work that could be done at low cost on the control-plane
- the current support of both delta and sotw, for both linear and simple cache means that every change has to be done 4 times, with four very different implementations, each one supporting only a subset of use cases (e.g. linear does not support cds/lds as it doesn't handle wildcard in this case, and sotw does not work properly as resource list is improperly maintained). I strongly believed that both implementation should converge. The only remaining differences should be in the messages themselves, abstracted in the server part. Having a common interface for sotw and delta is a first step for this. It would also allow in the future to make snapshot far more efficient by supporting a linear version per snapshot

In this context, this PR is changing:
- creates a ClientState within the cache package, driven by the data needed from the cache perspective to be able to answer a client request
- simplifies streamState
  - makes it compatible with the new ClientState interface
  - removes the IsFirst notion, as the protocol states it should come from the nonce in the request
  - removes the duplicate/parallel knownResourceNames and use the delta one instead. It will also properly carry the version in a future PR


alecholmez:
Hey @valerian-roche can you resolve those merge conflicts so I can start taking a look at this?

valerian-roche:
> Hey @valerian-roche can you resolve those merge conflicts so I can start taking a look at this?

Hey @alecholmez 
Sorry for the delay, rebased on upstream. As the change is only in a log line I rebased the commits rather than merged on top

valerian-roche:
This PR will need to be rebased on `main` once #657 is merged

valerian-roche:
I rebased this PR on the latest changes (especially the rollback of #615 which was blocking it)

jpeach (commented):
Just a couple of drive-by notes.

I'd also suggest adding more commentary to the commit message to describe the overall change, why the change is needed and the effects of the change on the system (particularly for the rework commit). The reader kinda has to try to stitch that together there from the PR description, and a bit of hand-holing would be helpful :)

pkg/cache/v3/cache.go
```diff
@@ -37,6 +36,27 @@ type Request = discovery.DiscoveryRequest
 // DeltaRequest is an alias for the delta discovery request type.
 type DeltaRequest = discovery.DeltaDiscoveryRequest
 
+// ClientState provides additional data on the client knowledge for the type matching the request
+// This allows proper implementation of stateful aspects of the protocol (e.g. returning only some updated resources)
+// Though the methods may return mutable parts of the state for performance reasons,
+// the cache is expected to consider this state as immutable and thread safe between a watch creation and its cancellation
+type ClientState interface {
+	// GetKnownResources returns the list of resources the clients has ACKed and their associated version.
```
alecholmez, line 44:
```suggestion
	// GetKnownResources returns a list of resources that clients have ACK'd and their associated version.
```

pkg/cache/v3/cache.go
```diff
@@ -37,6 +36,27 @@ type Request = discovery.DiscoveryRequest
...
+	// The versions are:
+	//  - delta protocol: version of the specific resource set in the response
+	//  - sotw protocol: version of the global response when the resource was last ACKed
+	GetKnownResources() map[string]string
```
alecholmez, line 48: I'm beginning to think that using `map[string]string` everywhere is asking for trouble. How do you feel about defining a data structure that holds some of this info? We could guard it and make it thread safe too. If we're controlling this interface it might be nice to provide some guard rails for consuming users

pkg/cache/v3/cache.go
```diff
@@ -37,6 +36,27 @@ type Request = discovery.DiscoveryRequest
 // DeltaRequest is an alias for the delta discovery request type.
 type DeltaRequest = discovery.DeltaDiscoveryRequest
 
+// ClientState provides additional data on the client knowledge for the type matching the request
+// This allows proper implementation of stateful aspects of the protocol (e.g. returning only some updated resources)
+// Though the methods may return mutable parts of the state for performance reasons,
+// the cache is expected to consider this state as immutable and thread safe between a watch creation and its cancellation
+type ClientState interface {
```
alecholmez, line 43: What if we just called this `State`? To me it this has always been a data structure that bridges the clients and servers and their state management. My original thought behind StreamState was that it was ambiguous enough to be the middle ground between the client/server but contextual enough to know it's on an individual gRPC stream. 

If we want to call it ClientState I'm cool with that, just a thought though.

pkg/cache/v3/cache.go
```diff
@@ -37,6 +36,27 @@ type Request = discovery.DiscoveryRequest
...
+	// GetKnownResources returns the list of resources the clients has ACKed and their associated version.
+	// The versions are:
+	//  - delta protocol: version of the specific resource set in the response
+	//  - sotw protocol: version of the global response when the resource was last ACKed
+	GetKnownResources() map[string]string
+
+	// GetSubscribedResources returns the list of resources currently subscribed to by the client for the type.
+	// For delta it keeps track of subscription updates across requests
+	// For sotw it is a normalized view of the last request resources
+	GetSubscribedResources() map[string]struct{}
```
alecholmez, line 53: Ah I think I remember why this is needed as a map. Wondering if we that's actually necessary in the cache... I'll take a look

alecholmez (commented):
Just some minor thoughts. IMO we should really linger on what we want for this interface here. I like the idea but this is going to be a big part of the state machine between the client/server.

pkg/cache/v3/cache.go
```diff
@@ -50,7 +70,7 @@ type ConfigWatcher interface {
 	//
 	// Cancel is an optional function to release resources in the producer. If
 	// provided, the consumer may call this function multiple times.
-	CreateWatch(*Request, stream.StreamState, chan Response) (cancel func())
+	CreateWatch(*Request, ClientState, chan Response) (cancel func())
```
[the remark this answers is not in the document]
valerian-roche: Will do

pkg/cache/v3/cache.go
```diff
@@ -37,6 +36,27 @@ type Request = discovery.DiscoveryRequest
...
+	//  - sotw protocol: version of the global response when the resource was last ACKed
+	GetKnownResources() map[string]string
+
+	// GetSubscribedResources returns the list of resources currently subscribed to by the client for the type.
+	// For delta it keeps track of subscription updates across requests
+	// For sotw it is a normalized view of the last request resources
+	GetSubscribedResources() map[string]struct{}
+
+	// IsWildcard returns whether the client has a wildcard watch.
+	// This considers subtilities related to the current migration of wildcard definition within the protocol.
```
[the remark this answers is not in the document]
valerian-roche: Will add links for those

valerian-roche:
Rebased the PR on the ads sotw changes. Will address further comments and update the PR

pkg/cache/v3/cache.go
valerian-roche: I renamed it to `SubscriptionState`, but I'm absolutely open to rename it

pkg/cache/v3/cache.go
valerian-roche: The split between `IsWildcard` and `GetSubscribedResources` should now allow to remove the old assumption of "no resources" == "wildcard" which is no longer true. Also a request with `IsWildcard` set to true and explicit resources in `GetSubscribedResources` carries the meaning of "client requested wildcard and explicit resources", and a cache implementation can act on this

valerian-roche:
I fully rebased the PR and replied to comments. I'd like this or #718 to move forward to be able to push major fixes on sotw and sotw-ads which are critical for xds-grpc

Issue #583, opened 2022-08-23
Removing and adding a cluster will cause EDS not sending to Envoy


tony612:
I found my go-control-plane doesn't send EDS sometimes when I remove a Cluster and then add it back via CDS. And this will cause 503 with `no healthy upstream`.

My operations are like:
1. Add needed resources like LDS, CDS, EDS. LDS and CDS are using Snapshot cache and EDS uses Linear cache. Now the API call via Envoy returns `200 OK`.
2. Cluster A is removed via CDS. Now the API returns `404`.
3. Cluster A is added back again. Now it got `503 no healthy upstream`.

I researched a little and guess it's because:
1. When Cluster A is removed, Envoy will also remove the ClusterLoadAssignment.
2. When Cluster A is added again, Envoy will send EDS request using the previous version.
3. Then Linear cache will think the cache is not stale and will open watch but don't send a response.

The Envoy logs(some logs are ignored) are:
```
# access log. 202
[2022-08-22T11:23:31.488Z] "GET /ctx1/echo/abc HTTP/1.1" 200 - 0 60 5 3 "-" "curl/7.77.0" "6223f5b7-86d5-4e50-bd20-a25b4d45224c" "my-ip:32068" "10.0.0.128:18081"

# remove cluster
[2022-08-22 11:23:39.195][14][debug][config] [external/envoy/source/common/config/grpc_mux_impl.cc:165] Received gRPC message for type.googleapis.com/envoy.config.cluster.v3.Cluster at version 650059771229369675
[2022-08-22 11:23:39.196][14][debug][config] [external/envoy/source/common/config/grpc_mux_impl.cc:143] Pausing discovery requests for type.googleapis.com/envoy.config.endpoint.v3.ClusterLoadAssignment (previous count 0)
[2022-08-22 11:23:39.196][14][info][upstream] [external/envoy/source/common/upstream/cds_api_helper.cc:35] cds: add 0 cluster(s), remove 2 cluster(s)
[2022-08-22 11:23:39.196][14][debug][upstream] [external/envoy/source/common/upstream/cluster_manager_impl.cc:769] removing cluster my-cluster
[2022-08-22 11:23:39.196][14][debug][upstream] [external/envoy/source/common/upstream/cds_api_helper.cc:67] cds: remove cluster 'my-cluster'
[2022-08-22 11:23:39.196][21][debug][connection] [external/envoy/source/common/network/connection_impl.cc:139] [C7] closing data_to_write=0 type=1
[...]

PR #586, opened 2022-08-25
[#583][Delta] Ensure resources are properly sent again if envoy unsubscribes then subscribes again to a resource

valerian-roche:
This PR is part of the fix for https://github.com/envoyproxy/go-control-plane/issues/583 related to resources being unsubscribed from then subscribed to again
This one applies to both simple and linear cache when using the delta protocol.
Actionable cases would likely only impact delta-ADS as EDS outside of ADS does not share streams and will therefore usually not have multiple resources this way

The logger issue is related to sotw-ADS which is part of a followup PR
This PR is built on top of https://github.com/envoyproxy/go-control-plane/pull/584 as the sotw part will be heavily relying on it


haoruolei:
is this work still active?

valerian-roche:
This work builds on #584, which itself now depends on rolling back #615. Sadly graphite and branch stacking cannot be done across forks
Alec has had few cycles recently to review, but when he becomes available I hope to have those move ahead

alecholmez:
@valerian-roche I just took a quick pass at the PR. I like the changes. We'll have to be careful with some of the breaking API changes and call that out when we release. Or we keep them around and tag them as `// DEPRECATED` in a comment and pull them out in the next release to at least give folks sometime to move away. What do you think about that?

Can you rebase? I'm cool with shipping this through if you're ready to pull it out of draft.

valerian-roche:
@alecholmez On our side we pushed for multiple fixes in order to support grpc-xds natively on this [fork](https://github.com/DataDog/go-control-plane/commits/dd/main/). We have been running for a while with this (also got some fix on grpc to have it properly working). 
It does include the API breaking changes in this PR. I was hoping to get #714 merged first to be able to break API here, as imo if we remove the protobuf dep we can be a more pushy on those changes. Wdyt?

Issue #595, opened 2022-10-03
Include Rate-Limit xDS config server related control plane functionality


renuka-fernando:
I am working on https://github.com/envoyproxy/ratelimit/issues/201 which is a feature for updating Rate Limit Service configurations dynamically through an xDS server.

As per https://github.com/envoyproxy/ratelimit/issues/201 I thought of adding the server-related implementation within the go-control-plane repository. Then we could re-use most of the readily available functions here. Hence, I created this git issue to discuss how we can achieve this.

Please share your thoughts on this.

RateLimit config proto: https://github.com/envoyproxy/ratelimit/pull/371

alecholmez:
Hi @renuka-fernando this would be awesome!

I'd actually like to have the rate limit service become a first class citizen of this repo since it's needed for control-plane functionality :)

renuka-fernando:
hi @alecholmez,

Is it possible to add the "no stalebot" label?

kkcmadhu:
hi @renuka-fernando can you kindly add this to java-control-plane too, Please.. It will be super helpful.

renuka-fernando:
> hi @renuka-fernando can you kindly add this to java-control-plane too, Please.. It will be super helpful.

sure, I will. Once @alecholmez confirms the changes on go control plane, I will have a look at the java control plane as well.

kkcmadhu:
kindly, have this included in java control plane too, i am eagerly awating to try it out.

renuka-fernando:
Hi @alecholmez,

Missed one more thing, we have to have a way to generate pb files. As we discussed here https://github.com/envoyproxy/go-control-plane/pull/598#discussion_r1008130688

So I will reopen this again to track it. Source proto files: https://github.com/envoyproxy/ratelimit/pull/371

alecholmez:
@renuka-fernando okay great! Thanks for following up with that.

renuka-fernando:
@alecholmez Can you re-open this? looks like I can't re-open this.

PR #598, opened 2022-10-13
Add Rate-Limit xDS config server related control plane functionality

renuka-fernando:
### Related issue
Fix envoyproxy/go-control-plane#595

### Proto files
PR: https://github.com/envoyproxy/ratelimit/pull/368

### Description
To configure envoy rate limit service via xDS


renuka-fernando:
@alecholmez could you please review this?

renuka-fernando:
Hi @alecholmez,

What do you think is the best? Using ADS or having a separate RLS config service?
I removed the RLS config service and commit https://github.com/envoyproxy/go-control-plane/pull/598/commits/d0a39a043e3bfaf89934ad6bf8eb5c2e35f2860a to this PR

renuka-fernando:
Is there an xDS client for ADS that can be reused in the rate limit? So I do not want to implement an xDS client in the rate limit repo, just import from go-control-plane and use it.

alecholmez:
Sorry for the slow response times! I'll begin to look into this.

I would actually like us to have RLS over ADS. That's the most common use case and if people begin to request that we supported a separated RLS service we can address that down the line.

ratelimit/config/ratelimit/v3/rls_conf.pb.go
```diff
@@ -0,0 +1,551 @@
+// Code generated by protoc-gen-go. DO NOT EDIT.
```
alecholmez, line 1: Some questions on this generated pb file...

Where does this come from? How can we automate that it keeps up with changes from its source? If this is from the envoy data-plane mirror or the udpa repo, why is it receiving it's own folder instead of living under the `envoy/` generated code folder.

Can you give me some background here?

alecholmez (changes requested):

pkg/resource/v3/resource.go
```diff
@@ -25,6 +25,9 @@ const (
 	ExtensionConfigType = APITypePrefix + "envoy.config.core.v3.TypedExtensionConfig"
 	RuntimeType         = APITypePrefix + "envoy.service.runtime.v3.Runtime"
 
+	// Rate Limit service
+	RateLimitConfigType = APITypePrefix + "ratelimit.config.ratelimit.v3.RateLimitConfig"
```
alecholmez, line 29: Do we not need to add anything in the `server/` folder to support sending configuration over the grpc streams? This is only in the cache which adds support there but is the intention to send RLS configs over an ADS stream?

Forgive me its been a while since I've looked into the ratelimiting subsystem so if Im missing something obvious let me know. But if the former is the use case target we will need to add the functionality inside the gRPC server alongside the cache.

alecholmez (changes requested):

pkg/resource/v3/resource.go
alecholmez: I see that you have this outlined in one of your previous issues:
```proto
// [#protodoc-title: RateLimitConfigDiscoveryService]
service RateLimitConfigDiscoveryService {
    rpc StreamDescriptors(stream envoy.service.discovery.v3.DiscoveryRequest)
        returns (stream envoy.service.discovery.v3.DiscoveryResponse) {
    }
    
    rpc FetchConfigs(envoy.service.discovery.v3.DiscoveryRequest) returns (envoy.service.discovery.v3.DiscoveryResponse) {
    }
}
```
This is what im alluding to

ratelimit/config/ratelimit/v3/rls_conf.pb.go
renuka-fernando: I planned to keep the source proto file in the rate limit service repository https://github.com/envoyproxy/ratelimit/pull/371 since it contains the rate limit config message.

The pb file is generated on my local machine (definitely we have to automate it 🙂). The location to keep this pb file is just what I decided, we can move it to a proper location. Could you please help me on automating this pb file generation?

It is better if we can keep the source proto file in the rate limit repository, but I think it is not an issue to keep that in another repository.

pkg/resource/v3/resource.go
renuka-fernando: > But if the former is the use case target we will need to add the functionality inside the gRPC server alongside the cache.

In the begging, we decided to have a separate RateLimitConfigDiscoveryService. And later based on @mattklein123's suggestion: https://github.com/envoyproxy/ratelimit/pull/368#issuecomment-1281098702 we are going to use ADS stream. If we are going with this approach could you please point me to the functionalities you mentioned above?

alecholmez: So yes we will want to be use ADS, but in the go-control-plane repo we still have to implement the service in the `server/` code. Once it's implemented it can be multiplexed over the ADS stream handler found here: https://github.com/envoyproxy/go-control-plane/blob/main/pkg/server/v3/server.go#L300

Note that the other xDS services have similar style implementations but can all be served over ADS.

ratelimit/config/ratelimit/v3/rls_conf.pb.go
alecholmez: Okay no worries let's definitely keep the proto file close to the implementation. 

I think it's worth moving it. I'm about to travel for the day so I will think about the automation and where we want to keep the generated code.

pkg/resource/v3/resource.go
renuka-fernando: As I understand your comment, I think I have to add `RateLimitConfigDiscoveryService` back. Please correct me if I am wrong.

```proto
// [#protodoc-title: RateLimitConfigDiscoveryService]
service RateLimitConfigDiscoveryService {
    rpc StreamDescriptors(stream envoy.service.discovery.v3.DiscoveryRequest)
        returns (stream envoy.service.discovery.v3.DiscoveryResponse) {
    }
    
    rpc FetchConfigs(envoy.service.discovery.v3.DiscoveryRequest) returns (envoy.service.discovery.v3.DiscoveryResponse) {
    }
}
```

Could you please have a look at the commit https://github.com/envoyproxy/go-control-plane/pull/598/commits/bcfa6c3495909221ba33bed473201f06add8e4b1 in this PR? I removed `RateLimitConfigDiscoveryService`, since I was thought to use ADS only. So I think I have to rollback this commit.

kkcmadhu:
please include this for java control plane too.

pkg/resource/v3/resource.go
alecholmez: @renuka-fernando yes you'll want to roll back that commit. We need those so ADS can actually hook into them. ADS simply calls out to the xDS endpoints over a single stream. The control plane still needs to register those gRPC endpoints regardless of the protocol type.

renuka-fernando: Added back with https://github.com/envoyproxy/go-control-plane/pull/598/commits/b4ea9af5085a55a1cf817fecc881dabc8b683daa

renuka-fernando:
Hi @alecholmez,

Could you please run the failed build again and please review the PR? The issue with the build is fixed now: https://github.com/envoyproxy/go-control-plane/issues/625

pkg/server/v3/server.go
```diff
@@ -220,6 +222,10 @@ func (s *server) StreamExtensionConfigs(stream extensionconfigservice.ExtensionC
 	return s.StreamHandler(stream, resource.ExtensionConfigType)
 }
 
+func (s *server) StreamRlsConfigs(stream rlsconfigservice.RateLimitConfigDiscoveryService_StreamRlsConfigsServer) error {
```
alecholmez, line 225: Can we add a method for the Delta RLS stuff as well? Or does RLS not support delta yet?

alecholmez (changes requested):

pkg/server/v3/server.go
renuka-fernando: Rls does not support delta yet.

alecholmez: Cool can you make an issue to track that?

alecholmez (approved):
Merging this since it's ready to go

renuka-fernando:
@alecholmez, thank you very much for the support in getting this done.

pkg/server/v3/server.go
renuka-fernando: created an issue: https://github.com/envoyproxy/ratelimit/issues/393

Issue #599, opened 2022-10-19
Snapshot Cache shouldnt log resource content


arkodg:
The cache logs which resources were added or removed[ here](https://github.com/envoyproxy/go-control-plane/blob/799a7af9e5b9b3c492642319bf6a71cdfffc9cac/pkg/cache/v3/simple.go#L530), exposing sensitive information that is part of the xds config.

PR #601, opened 2022-10-21
rm printing resources in cache logs

arkodg:
Fixes: https://github.com/envoyproxy/go-control-plane/issues/599

Signed-off-by: Arko Dasgupta <arko@tetrate.io>

### DIFF
--- a/pkg/cache/v3/linear.go
+++ b/pkg/cache/v3/linear.go
@@ -186,8 +186,8 @@ func (cache *LinearCache) respondDelta(request *DeltaRequest, value chan DeltaRe
 	// Only send a response if there were changes
 	if len(resp.Resources) > 0 || len(resp.RemovedResources) > 0 {
 		if cache.log != nil {
-			cache.log.Debugf("[linear cache] node: %s, sending delta response with resources: %v removed resources %v wildcard: %t",
-				request.GetNode().GetId(), resp.Resources, resp.RemovedResources, state.IsWildcard())
+			cache.log.Debugf("[linear cache] node: %s, sending delta response for typeURL %s with resources: %v removed resources %v with wildcard: %t",
+				request.GetNode().GetId(), request.TypeUrl, GetResourceNames(resp.Resources), resp.RemovedResources, state.IsWildcard())
 		}
 		value <- resp
 		return resp
--- a/pkg/cache/v3/linear.go
+++ b/pkg/cache/v3/linear.go
@@ -186,7 +186,7 @@ func (cache *LinearCache) respondDelta(request *DeltaRequest, value chan DeltaRe
 	// Only send a response if there were changes
 	if len(resp.Resources) > 0 || len(resp.RemovedResources) > 0 {
 		if cache.log != nil {
-			cache.log.Debugf("[linear cache] node: %s, sending delta response for typeURL %s with resources: %v removed resources %v with wildcard: %t",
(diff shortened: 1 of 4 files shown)


pkg/cache/v3/simple.go
```diff
@@ -527,8 +527,8 @@ func (cache *snapshotCache) respondDelta(ctx context.Context, snapshot ResourceS
 	// otherwise, envoy won't complete initialization
 	if len(resp.Resources) > 0 || len(resp.RemovedResources) > 0 || (state.IsWildcard() && state.IsFirst()) {
 		if cache.log != nil {
-			cache.log.Debugf("node: %s, sending delta response with resources: %v removed resources %v wildcard: %t",
-				request.GetNode().GetId(), resp.Resources, resp.RemovedResources, state.IsWildcard())
+			cache.log.Debugf("node: %s, sending delta response with wildcard: %t",
```
sunjayBhatia, line 530: I think it would still be useful to get a sense of what resources are being changed in this debug log, would be nice to have the typeurl and names of resources or at the very least how many were requested/removed

arkodg: @sunjayBhatia is there an existing helper method in the repo that converts `resources` to its name and typedUrl, or would I need to create one ?

alecholmez: +1 for at least keeping the names. We can definitely remove the entire payload but let's keep the names for debugging purposes.

sunjayBhatia: the typeurl you should get with `request.TypeUrl`, `resp.RemovedResources` should give you the names of the computed resources to remove, and for the resources that are to be included in the response, as far as i know there isn't a direct helper, but there is a helper to get a resource name without fully marshalling it, see: https://github.com/envoyproxy/go-control-plane/blob/b9e8d765366d34116028760e6192d40d69628636/pkg/cache/v3/cache.go

we may want to instead of storing as a slice store these as a map? so its easier to get the names out?

a message like the below would be great:

```
node: %s, sending delta response for typeURL %s with resources: %v removed resources %v wildcard: %t
```

pkg/cache/v3/linear.go
```diff
@@ -186,8 +186,8 @@ func (cache *LinearCache) respondDelta(request *DeltaRequest, value chan DeltaRe
 	// Only send a response if there were changes
 	if len(resp.Resources) > 0 || len(resp.RemovedResources) > 0 {
 		if cache.log != nil {
-			cache.log.Debugf("[linear cache] node: %s, sending delta response with resources: %v removed resources %v wildcard: %t",
-				request.GetNode().GetId(), resp.Resources, resp.RemovedResources, state.IsWildcard())
+			cache.log.Debugf("[linear cache] node: %s, sending delta response for typeURL %s with resources: %v removed resources %v with wildcard: %t",
```
sunjayBhatia, line 189: tiny nit, missed a colon in my earlier suggested change

```suggestion
			cache.log.Debugf("[linear cache] node: %s, sending delta response for typeURL %s with resources: %v removed resources: %v with wildcard: %t",
```

sunjayBhatia (dismissed):
looks good, made some tiny nits if you wanted to address them

cc @alecholmez if you had any other notes

arkodg:
@sunjayBhatia thanks for review, addressed the comments, can you ptal

sunjayBhatia (approved):

alecholmez:
> looks good, made some tiny nits if you wanted to address them
> 
> cc @alecholmez if you had any other notes

Nope looks good to me!

PR #603, opened 2022-10-24
Fix golangci-lint errors on main

sunjayBhatia:
Includes:
- bump Makefile golangci-lint version to 1.50.1
- spelling fixes
- add line between license/package name
- other gofmt issues
- stop using ioutil package (deprecated in go 1.16)
- ignore a couple gosec issues in test code


alecholmez (approved):
Thanks for this!

PR #614, opened 2022-12-11
ci(deps): setup dependabot

mmorel-35:
Setup dependabot for github-actions and go dependencies. 
It also intregates the updates of github-actions and go submodules of the project as @jacob-delgado already created #607 for the main module dependencies. 


alecholmez:
This seems like a good idea to me +1 for dependabot

xdsmatcher/go.mod
```diff
@@ -3,14 +3,9 @@ module github.com/envoyproxy/go-control-plane/xdsmatcher
 go 1.15
```
alecholmez, line 3: Might be a good idea to bump this to whatever is at the root go.sum too. We can do that later though

alecholmez (approved):

xdsmatcher/go.mod
mmorel-35: I think it worth another PR Indeed. As 1.19 has been published, it might even be an opportunity to upgrade go version to 1.17 for all thé modules

alecholmez (approved):

PR #615, opened 2022-12-12
Fix for the SDS update failure

AmitKatyal-Sophos:
### Summary
This PR includes the changes for the SDS update failure

### Problem
We started using Delta ADS for LDS, CDS, RDS and SDS resources. We have three instances of envoy proxy client connected to the management server. During testing, we found that some of the envoy proxy's were not receiving the SDS updates.

### Root cause analysis
We investigated the issue by adding logs to the go-control-plane and figured out that control plane is managing the list of the subscribed resources for each client and before sending the updates it compares the hash of the resource set in the snapshot with the hash of the same resource in the subscription list. If the specified resource is not specified in the client's subscription list, then updates are not sent to the respective client. And we found that somehow the entry from the subscription list was getting removed and as a result the updates were not getting pushed to the envoy proxy.

Why resources were getting removed from the subscription list ?
processDelta handler responsible for handling the client requests and responses updates the "watch.state.resourceVersions" at two places,

[1] As part of the subscription request from the envoy "s.subscribe(req.GetResourceNamesSubscribe(), watch.state.GetResourceVersions())"

[2] After sending the response to the envoy proxy "watch.state.SetResourceVersions(resp.GetNextVersionMap())"

Here, the problem is in the update of the resourceversion map after sending the response. After sending the response, we are directly replacing the resourceversion map with the map specified in the response. This is problem for the non wildcard resource type.

Please find below the detailed use-case to understand why replacing the resourceversion map inside the response handling is problematic.

Envoy requests for the resource (resource-1)

In the request handler, stream state resourceversion map is updated to [resource-1: empty hash] & response-1 is created with newresouceversion map [resource-1: hash]

Before the response is sent to the envoy, envoy sends the new request for the resource (resource-2). In the request handler, stream state resourceversion map is updated to [resource-1: empty hash, resource-2 empty hash] and response-2 is created with newresouceversion map [resource-1: hash, resource-2: hash].

Response-1 is sent to the envoy proxy and stream state version map is updated with the newresouceversion map [resource-1: hash]. After sending the response-1, the stream state resourceversion map will be updated to [resource-1: hash]

Envoy sends another request for the resource-3. In the request handler, stream state resourceversion map is updated to [resource-1: empty hash, resource-3 empty hash] and response-3 is created with newresouceversion map [resource-1: hash, resource-3: hash].

Response-2 is sent to the envoy proxy and stream state version map is updated with the newresouceversion map [resource-1: hash, resource-2: hash]. After sending the response-2, the stream state resourceversion map will be [resource-1: hash, resource-2: hash]

Response-3 is sent to the envoy proxy and stream state version map is updated with the newresouceversion map [resource-1: hash, resource-3: hash]. After sending the response-3, the stream state resourceversion map will be [resource-1: hash, resource-3: hash]

Final resourceversion map will be [resource-1: hash, resource-3: hash]. This is incorrect as we lost the subscription to the resource-2. Hence, any config updates to the resource-2 will not be sent to the envoy proxy.

### DIFF
--- a/pkg/server/delta/v3/server.go
+++ b/pkg/server/delta/v3/server.go
@@ -119,8 +119,26 @@ func (s *server) processDelta(str stream.DeltaStream, reqCh <-chan *discovery.De
 
 			watch := watches.deltaWatches[typ]
 			watch.nonce = nonce
+			// As per XDS protocol, for the non wildcard resources, management server should only respond to the resources
+			// requested by the client. Since we were replacing (instead of updating) the complete state resource version
+			// map after responding to the client, it was overriding/removing the resources subscribed by the client intermittently.
+			// As a result, the update of the such resources was never sent to the client.
+			// In order to address the issue, started updating the resources hash in the existing map instead of replacing
+			// the completed map.
+			// In case of wildcard resources, client never subscribes for the resources, replacing the state resource version based
+			// on the response by the management server is not an issue. Hence, the fix is only applicable for the non wildcard resources.
+			if !watch.state.IsWildcard() {
+				for k, hash := range resp.GetNextVersionMap() {
+					if currHash, found := watch.state.GetResourceVersions()[k]; found {
+						if currHash != hash {
+							watch.state.GetResourceVersions()[k] = hash
+						}
+					}
+				}
+			} else {
+				watch.state.SetResourceVersions(resp.GetNextVersionMap())
+			}
 
-			watch.state.SetResourceVersions(resp.GetNextVersionMap())
 			watches.deltaWatches[typ] = watch
 		case req, more := <-reqCh:
 			// input stream ended or errored out


AmitKatyal-Sophos:
Can someone please review the changes ? Thanks.

pkg/server/delta/v3/server.go
```diff
@@ -120,7 +120,18 @@ func (s *server) processDelta(str stream.DeltaStream, reqCh <-chan *discovery.De
 			watch := watches.deltaWatches[typ]
 			watch.nonce = nonce
 
-			watch.state.SetResourceVersions(resp.GetNextVersionMap())
+			if !watch.state.IsWildcard() {
```
alecholmez, line 123: Can we drop a comment in here about why non-wildcard has to be treated differently? Or a link to some docs would also be nice. It'll help me updating the changelog down the line.

alecholmez (changes requested):

pkg/server/delta/v3/server.go
AmitKatyal-Sophos: sure, will do that.

AmitKatyal-Sophos: done

alecholmez (approved):

alecholmez (approved):

PR #634, opened 2023-01-27
Remove unused context from arguments to internal/example/server.

tortuoise:
Add Server type to rework exaple.

Signed-off-by: R C Sridhar <rcs@m0v.in>

### DIFF
--- a/internal/example/server.go
+++ b/internal/example/server.go
@@ -31,7 +31,9 @@ import (
 	routeservice "github.com/envoyproxy/go-control-plane/envoy/service/route/v3"
 	runtimeservice "github.com/envoyproxy/go-control-plane/envoy/service/runtime/v3"
 	secretservice "github.com/envoyproxy/go-control-plane/envoy/service/secret/v3"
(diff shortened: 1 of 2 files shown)


internal/example/server.go
```diff
@@ -41,6 +43,58 @@ const (
...
+type Server struct {
+	xdsserver server.Server
+}
+
+func NewServer(ctx context.Context, cache cache.Cache, cb *test.Callbacks) *Server {
+	srv := server.NewServer(ctx, cache, cb)
+	return &Server{srv}
+}
+
+func (s *Server) registerServer(grpcServer *grpc.Server) {
```
alecholmez, line 55: Can we not just use the helper functions in the test package instead of duplicating that code?

https://github.com/envoyproxy/go-control-plane/blob/main/pkg/test/v3/register.go#L38

Same with `Run()` theres ` RunManagementServer` helper function.

alecholmez (changes requested):

PR #662, opened 2023-03-14
Set safe directory in CI script to fix build issue related to VCS stamping in binaries

valerian-roche:
A change in go 1.18 now include some versioning metadata in binaries. 
This became broken in CI, potentially related to an update of git in the base image of the build image

We can flag our own repo as safe to avoid this issue as we know it is the mounter volume

### DIFF
--- a/build/do_ci.sh
+++ b/build/do_ci.sh
@@ -3,6 +3,9 @@
 set -e
 set -x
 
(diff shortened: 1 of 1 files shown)


valerian-roche:
Merged through #657

PR #663, opened 2023-03-20
Fix test to pass through vet and properly behave as designed

valerian-roche:
A test in `simple_test.go` is currently failing the `go vet` check as it is referencing an external variable by reference while in a parallel run
Further test confirmed the value of `i` was not the expected one in the test
The fix does not attempt to simplify the test and simply make it work as designed, while fixing go vet on the repo


valerian-roche:
@alecholmez I will review #544 
Ideally I'd like this fix to go separately if we need to revert or readjust #544 as CI is currently broken on the repo

PR #692, opened 2023-05-14
add size hints for resource maps and slice

dynajoe:
Noticed some areas that could be improved with initial size hints.


jpeach (approved):

alecholmez (approved):
Looks good thanks for the change!

PR #700, opened 2023-05-23
ci: build dyplomat as part of CI

jpeach:
Even though dyplomat has no tests, we should still add it to CI to get a signal when dependabot wants to update the Go dependencies. We just want to know whether it still builds.

### DIFF
--- a/build/do_ci.sh
+++ b/build/do_ci.sh
@@ -22,6 +22,7 @@ make examples
 make test
 make integration
 
-cd ./xdsmatcher
-make test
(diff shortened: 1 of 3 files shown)


jpeach:
> could also add a make target to build the dyplomat container image in addition/instead in CI

Yeh it would probably be worth doing that. We are already in a Docker container in this CI job though, so we probably want a different approach to verify it.



> could also add a make target to build the dyplomat container image in addition/instead in CI
> 
> seems reasonable as a lightweight way to validate things compile

PR #701, opened 2023-05-28
ci: Add sync workflow

phlax:
Currently Envoy CI uses an SSH key with all powers to sync this repo by pushing to it

This is not optimally secure, and furthermore this pattern regularly causes Envoy CI to flake racing to checkout/push downstream repos.

I have created an app with just wf trigger permissions and added it to this repo

This PR adds a workflow that can be triggered by the app to sync Envoy by pulling from it


.github/workflows/envoy-sync.yaml
```diff
@@ -0,0 +1,40 @@
...
+jobs:
+  sync:
+    runs-on: ubuntu-20.04
+    permissions:
+      contents: write
+    if: |
+      ${{
+          !contains(github.actor, '[bot]')
+          || github.actor == 'sync-envoy[bot]'
+          || github.actor == 'sync-envoy-testing[bot]'
```
phlax, line 22: rm

alecholmez:
@phlax let me know when this gets out of wip, thanks for this change this is great

phlax:
>  let me know when this gets out of wip

will do, shouldnt take much, mostly i want to test it, but i think its also not quite right yet

jpeach:
> I have created an app with just wf trigger permissions and added it to this repo
> 
> This PR adds a workflow that can be triggered by the app to sync Envoy by pulling from it

When this is ready, could you please also add an explanation (to the README maybe?) of where all the moving parts are, and how someone could update them? If the app has any source code, then it might be good for it to also live in the `envoyproxy` GitHub org.

README.md
```diff
@@ -61,8 +66,8 @@ The Envoy xDS APIs follow a well defined [versioning scheme](https://www.envoypr
 
 ### Deprecated
 
-`V2` control-plane code has been removed and will no longer be supported. For previous conversations on support for various xDS versions, see here: 
-- [here](https://docs.google.com/document/d/1ZkHpz6DwEUmAlG0kb2Mgu4iaeQC2Bbb0egMbECoNNKY/edit?ts=5e602993#heading=h.15nsmgmjaaml)
```
phlax, line 65: my editor did a ws cleanup here

phlax:
@alecholmez i have shifted the algo from the python script to here, i will update to use the envoy `api.go` target once it has been added (https://github.com/envoyproxy/envoy/pull/27675)

ive tested this pretty well now, so reasonably confident this should hit the ground working, but will follow up if there any issues

currently it takes ~3 mins, at the cost of some complexity it could be a lot faster with caching

> When this is ready, could you please also add an explanation (to the README maybe?)

note about sync added

> ... of where all the moving parts are, and how someone could update them?

they are here, pr to this repo

> If the app has any source code

it doesnt

> then it might be good for it to also live in the envoyproxy GitHub org.

it does

phlax:
for ref, you can see a triggered wf run here https://github.com/phlax/go-control-plane/actions/runs/5111042721/jobs/9187570100

and a commit here https://github.com/phlax/go-control-plane/commit/ed3867062062a6a2b7f3f2ac0ab30480315b05c4

i have fixed the commit message

alecholmez (approved):
This looks great thanks for the addition

.github/workflows/envoy-sync.yaml
```diff
@@ -0,0 +1,43 @@
...
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  sync:
+    runs-on: ubuntu-20.04
```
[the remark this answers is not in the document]
phlax: 20.04 is what is used/tested in envoy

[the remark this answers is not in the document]
alecholmez: Cool sounds good

phlax:
can we land? i have the other repos that envoy syncs to updated, so we should be good to update envoy

ci/sync_envoy.sh
```diff
@@ -0,0 +1,95 @@
...
+build_protos () {
+    # TODO(phlax): use envoy do_ci target once https://github.com/envoyproxy/envoy/pull/27675 lands
+    local go_protos go_proto go_file rule_dir proto input_dir output_dir
+    echo "Building go protos ..."
+    cd "${ENVOY_SRC_DIR}" || exit 1
+
+    # shellcheck disable=SC1091
+    . ci/setup_cache.sh
+
+    read -a go_protos <<< "$(bazel query "kind('go_proto_library', ${GO_TARGETS[*]})" | tr '\n' ' ')"
```
jpeach, line 31: nit:
```
$ shellcheck ci/sync_envoy.sh

In ci/sync_envoy.sh line 31:
    read -a go_protos <<< "$(bazel query "kind('go_proto_library', ${GO_TARGETS[*]})" | tr '\n' ' ')"
    ^--^ SC2162 (info): read without -r will mangle backslashes.

For more information:
  https://www.shellcheck.net/wiki/SC2162 -- read without -r will mangle backs...
```

ci/sync_envoy.sh
```diff
@@ -0,0 +1,95 @@
+#!/bin/bash -e
+
+set -o pipefail
```
jpeach, line 3: nit: if you add `set -o errexit` here, it's the same as `bash -e`, except (IMHO) a bit clearer to the reader.

ci/sync_envoy.sh
```diff
@@ -0,0 +1,95 @@
...
+COMMITTER_EMAIL="envoy-sync[bot]@users.noreply.github.com"
+
+
+if [[ -z "$ENVOY_SRC_DIR" ]]; then
+    echo "ENVOY_SRC_DIR not set, it should point to a cloned Envoy repo" >&2
+    exit 1
+elif [[ ! -e "$ENVOY_SRC_DIR" ]]; then
+    echo "ENVOY_SRC_DIR ($ENVOY_SRC_DIR) not found, did you clone it?" >&2
+    exit 1
+fi
```
jpeach, line 19:
```suggestion
if [[ -z "$ENVOY_SRC_DIR" ]]; then
     echo "ENVOY_SRC_DIR not set, it should point to a cloned Envoy repo" >&2
     exit 1
 fi
 
 if [[ ! -d "$ENVOY_SRC_DIR" ]]; then
     echo "ENVOY_SRC_DIR ($ENVOY_SRC_DIR) not found, did you clone it?" >&2
     exit 1
 fi
```

ci/sync_envoy.sh
jpeach: Also consider `set -o nounset`, which is also a useful check, though not always what you want.

README.md
```diff
@@ -8,6 +8,11 @@ This repository contains a Go-based implementation of an API server that
 implements the discovery service APIs defined in
 [data-plane-api](https://github.com/envoyproxy/data-plane-api).
 
+## Proto files
+
+The go proto files are synced from the upstream Envoy repository (https://github.com/envoyproxy/envoy) on every upstream commit.
```
jpeach, line 13: Nit:

```suggestion
The Go proto files are synced from the upstream Envoy repository (https://github.com/envoyproxy/envoy) on every upstream commit.
```

ci/sync_envoy.sh
```diff
@@ -0,0 +1,95 @@
...
+        return
+    fi
+    last_envoy_sha="$(get_last_envoy_sha)"
+    changes="$(git -C "${ENVOY_SRC_DIR}" rev-list "${last_envoy_sha}"..HEAD)"
+    echo "Changes detected: "
+    echo "$changes"
+    latest_commit="$(git -C "${ENVOY_SRC_DIR}" rev-list "${last_envoy_sha}"..HEAD | head -n1)"
+    echo "$latest_commit" > envoy/COMMIT
+    git config --global user.email "$COMMITTER_EMAIL"
+    git config --global user.name "$COMMITTER_NAME"
```
jpeach, line 86: Nit: is the `--global` flag needed?

jpeach (approved):
Nice! Added a few review comments, but just for tiny things. Not a big deal.

ci/sync_envoy.sh
phlax: nouunset would not work here if it is checking for an unset var

alecholmez:
@phlax do you have merge permissions? If not I can roll it in

ci/sync_envoy.sh
phlax: > nit: if you add set -o errexit here, it's the same as bash -e, except (IMHO) a bit clearer to the reader.

thinking on this im not sure - my guess is that you if you surveyed devs more would recognize `-e` as its just far more common

re nounset - i guess it can probably work if i re/set it from the env

ci/sync_envoy.sh
phlax: switched `-e` -> `-d`

ci/sync_envoy.sh
phlax: would be good to add shellcheck to the ci here - i was running it locally before submitting but i think this slipped through

also, would be good to clear up the whitespace issues 

```console
$ pip install envoy.code.check
$ envoy.code.check . -c glint
```

will show you issues

ci/sync_envoy.sh
phlax: no

phlax:
> do you have merge permissions?

i do, but would rather let you guys do it

ive updated the pr with feedback

we can test it manually straight away when it lands, but i will also move to get the envoy side set up asap so we can test it properly - im happy to follow up if necessary

alecholmez:
I'm good to merge this... @jpeach I'll defer to you clicking the green button since you had some comments

jpeach (approved):

jpeach:
> I'm good to merge this... @jpeach I'll defer to you clicking the green button since you had some comments

Yeh, I just had a few nitpicks. This looks good overall 👍

phlax:
after removing `nounset` (#706 ) manual run has worked (with no change)

```console
Syncing go protos ...
Copying ../envoy/build_go/envoy -> envoy
Copying ../envoy/build_go/contrib -> contrib
Committing changes ...
Nothing changed, not committing
```

https://github.com/envoyproxy/go-control-plane/actions/runs/5141777442/jobs/9254666886#step:5:160

i will land the update in Envoy shortly https://github.com/envoyproxy/envoy/pull/27294 and check from there

the real test will be when there is some change to commit, so will keep watching

phlax:
looking good

image

PR #706, opened 2023-06-01
ci/sync: Remove nonunset from script

phlax:
this does not work as it source an Envoy script that uses vars that are not set here


phlax:
pushing this through so we can get this set up in envoy

PR #710, opened 2023-06-03
sync/ci: Assorted fixes

phlax:
- add `NO_BUILD_SETUP` before calling Envoy ci
- use app auth to push changes


phlax:
@alecholmez i reckon this should work now - i have it tested/working in other synced repos

in this case there were a couple of additional fixes required - one here, and the other in Envoy

jpeach (approved):

PR #711, opened 2023-06-04
sync/ci: Fix commit

phlax:


phlax:
@alecholmez @jpeach not sure why this is not working - it gets a 141 exit code - which is a pipefail from the git command

this doesnt make much sense given that the command is the same as the command above but just with piped output

its quite possible this is some git bug - there are certainly issues like this discussed on the interwebs

i have removed the pipe as it wasnt really necessary, and updated the command - if it fails now we should at least get a bit more information

i have also corrected the bot's commit username/email

jpeach (approved):

Issue #271, opened 2020-03-13
Support for Go 1.11 with GO111MODULES=on


gnossen:
It seems that https://github.com/envoyproxy/go-control-plane/pull/179 introduced a hard dependency on Go 1.12. While 1.11 is no longer supported by the core language team, it is still critical for environments like appengine, where it is still supported. I ran into this problem while trying to compile [`grpcurl` with Go 1.11](https://travis-ci.org/github/fullstorydev/grpcurl/jobs/661748053?utm_medium=notification&utm_source=github_status) ([PR](https://github.com/fullstorydev/grpcurl/pull/137)):

```
go test -race ./...
go build golang.org/x/sys/unix: module requires Go 1.12
go build github.com/cncf/udpa/go/udpa/annotations: module requires Go 1.13
go build github.com/envoyproxy/go-control-plane/envoy/annotations: module requires Go 1.12
go build github.com/envoyproxy/go-control-plane/envoy/type: module requires Go 1.12
FAIL	github.com/fullstorydev/grpcurl [build failed]
FAIL	github.com/fullstorydev/grpcurl/cmd/grpcurl [build failed]
Makefile:82: recipe for target 'test' failed
```

There seems to be a general problem with the Go tooling auto-inserting hard dependencies like this when none was intended by the author. Is that the case here? Or is there some feature that the library truly does depend on in Go 1.12+?

CC @kyessenov 
CC @davecheney

PR #714, opened 2023-06-05
Support multi-module releases in go-control-plane

### LINKED ISSUE #391: Develop release policy
gRPC-Go has several 1.x releases that depend upon this repo extensively for xds support.  If a breaking change were to happen here, it would break gRPC-Go users.

If there are concerns about some parts of the repo being unstable, then we would be happy with a sub-module inside `envoy` that can be released as a stable v1 for the generated proto messages.

See also https://github.com/cncf/udpa/issues/42. I believe this repo would also need a v1 release from the UDPA repo, since it is a dependency of several protos here.

cc @htuch, @markdroth

alecholmez:
This PR introduces support for multi-module releases in go-control-plane following the plan of action outlined in issue #391.

A few things have happened:
- Bumped modules to Go 1.20 (to inherit support for workspaces)
- Introduced a go workspace that orchestrates all the subcomponents of this repo
- Initialized new modules for `./envoy` and `./contrib` containing the generated go stubs
- Renamed `build/` to `scripts/` so we have a bucket for more general repo automation

When this PR is merged, a tag will be cut at `v0.11.2` but a new tag will be introduced with `envoy/v1.26.2`. closes #391


contrib/go.mod
```diff
@@ -0,0 +1,23 @@
+module github.com/envoyproxy/go-control-plane/contrib
+
+go 1.20
+
+replace github.com/envoyproxy/go-control-plane/envoy => ../envoy
```
valerian-roche, line 5: My understanding from `go.work` is that it should no longer be needed to add those replace commands
Is that actually not the case?

valerian-roche (approved):
I'm not familiar with go workspaces, but overall lgtm
Nice to see independent versioning through this though

contrib/go.mod
alecholmez: I was able to drop the replace from `./envoy` but that one still complained so I left it in.

jpeach: Maybe this is an indication that we don't have the right structure? Since the "envoy" and "contrib" protobufs are bound to the same Envoy release, does it make sense to allow the possibility for a program to import different versions of them?

How hard would it be to move the contrib protobufs to "envoy/contrib"?

jpeach: And I suppose that a similar question arises for the "./ratelimit" directory.

.gitignore
```diff
@@ -1,4 +1,6 @@
 /vendor
+/contrib/vendor
+/envoy/vendor
```
jpeach, line 3: Not sure that we need this, or "/vendor", since there's no local vendoring in they repo AFAICT.

contrib/go.mod
```diff
@@ -0,0 +1,23 @@
+module github.com/envoyproxy/go-control-plane/contrib
+
+go 1.20
```
jpeach, line 3: JFYI, we should be prepared for some pushback on this. It's possible that some consumer of this module will be using earlier Go versions (see #271 for example).

jpeach:
> * Initialized new modules for `./envoy` and `./contrib` containing the generated go stubs

We probably want to do the same for `./ratelimit` too?

contrib/go.mod
jpeach: OK, contrib seems weird, any maybe we want to fix it first.

The [dlb.proto](https://github.com/envoyproxy/envoy/blob/main/api/contrib/envoy/extensions/network/connection_balance/dlb/v3alpha/dlb.proto#L10) Go package option says it is in the "github.com/envoyproxy/go-control-plane/envoy" module. However, the actual source code is in "github.com/envoyproxy/go-control-plane/contrib/envoy/extensions".

Seems like the protobuf sync has been putting them in the wrong module? cc @phlax

phlax: is that an issue with this proto ?

if i do:

```console
$ find build_go/ -name "dlb*"
build_go/contrib/envoy/extensions/network/connection_balance/dlb
build_go/contrib/envoy/extensions/network/connection_balance/dlb/v3alpha/dlb.pb.go
build_go/contrib/envoy/extensions/network/connection_balance/dlb/v3alpha/dlb.pb.validate.go
```

it appears that the path is correct going on the src path

phlax: grepping further, none of the contrib apis appear to have contrib in their go_package path  - im not sure if i misunderstand the issue

jpeach: For example:

```
$ git grep go_package | grep vcl
api/contrib/envoy/extensions/vcl/v3alpha/vcl_socket_interface.proto:option go_package = "github.com/envoyproxy/go-control-plane/envoy/extensions/vcl/v3alpha";
```

But the actual import path that works is:
```
	_ "github.com/envoyproxy/go-control-plane/contrib/envoy/extensions/vcl/v3alpha"
```

So this seems like the two repositories don't agree on where contrib ought to live?

phlax: i suspect that when these were moved to contrib their `go_package` was never updated

jpeach: Is bazel the thing that chooses to generate them into the contrib path?

phlax: yeah i think so - and it uses the src path i believe

phlax: afaict none of the contrib extensions have the correct `go_package` on this basis

phlax:
```console
$ git grep go-control-plane api/contrib/ | grep go-control-plane/contrib
```

alecholmez: I'm in agreement here and would like contrib to fold under `/envoy`. I didn't do much investigation but seems like its in-scope for this PR. If we are going to break that import path I would say now is the time to do it since this will be the first multi-module release.

.gitignore
alecholmez: Ah yea I can delete that thanks for the catch. Was just testing stuff

contrib/go.mod
alecholmez: I can try and roll back to 1.18 but that is a bare minimum requirement for workspaces

contrib/go.mod
phlax: not sure iiuc - one thing im thinking is that if the `go_package` that envoy has listed doesnt match what it produces then we probably want to fix something there one way or other

the diff above would need to be applied in envoy, so also a bit concerned who else might be consuming it but then i think if we are fixing the package/path mismatch that is kinda ok (and perhaps suggests not so many)

alecholmez: @phlax you think its a worthy issue to open in envoy? We can see what people think

phlax: certainly if there is a mismatch i think we should

not clear what the best solution is but would totally defer to you - definitely good to get other opinions if its a breaking change tho

sunjayBhatia: > Correct me if I'm wrong the contrib folder is for things that don't see much active dev right?

contrib extensions might get active dev, depending on the instance, whether that surfaces in API changes or impl changes, I'm not sure, seems like right now API changes are less frequent

relevant: https://github.com/envoyproxy/envoy/blob/main/EXTENSION_POLICY.md#contrib-extensions

sunjayBhatia: one other (untested, so it may not work tbh) idea would be to rather than have separate `envoy` or `contrib` modules at the root or to try to reparent the `contrib` module under `envoy` would be to instead add an `api` (or some other name) module that contains the `envoy` and `contrib` packages so they can be versioned together

we could make this a purely additive change to give people time to switch over to using this new module versioning scheme/import path (i.e. keep mirroring to the existing locations in this repo for a couple releases), internally in the root module we might be able to switch over right away

i haven't thought about the specifics of modifying the import path/the `go_package` directive, but this might be a way to make that sort of breaking change easier to stomach

it adds some duplication but we can document that we will remove it soon

sunjayBhatia: Has anyone tried to use this branch in an existing project? I'm having trouble with Contour which would have to import the main module and envoy modules

sunjayBhatia: I'm getting ambiguous import issues since the new module maintains the same import path at the moment

jpeach: I did try the imports with a trivial program (that just did underscore imports from each module), and it worked for me. Testing on a real client project is much better though!

jpeach: > @jpeach Since we're creating 2 tags, one for the control server and another for everything else, would they not be versioned the same even if they're in separate folders?

When you release, I expect you would tag both of them, but projects that consume them would need to remember to update both modules together. I think that the intention of Go modules dependencies would be for the "contrib" module to require the matching version of the "envoy" module (so the release process would need to update `go.mod` in "contrib".

> contrib folder is for things that don't see much active dev right

Always having the API for contrib extensions doesn't bother me much since the IIUC, they are not enabled in the standard Envoy build.

>  instead add an api (or some other name) module that contains the envoy and contrib packages so they can be versioned together

Yes, that's what I'm proposing (by suggesting that we move the "contrib" directory inside "envoy").

sunjayBhatia: > Yes, that's what I'm proposing (by suggesting that we move the "contrib" directory inside "envoy").

I think my comment above is a slightly different, it intends to mirror the structure of the envoy `api` folder structure so `contrib` and `envoy` packages live at the same level under a new module (move/duplicate the current `envoy` and `contrib` packages into `github.com/envoyproxy/go-control-plane/api/[envoy|contrib]`)

sunjayBhatia: this branch has that set up: https://github.com/sunjayBhatia/go-control-plane/tree/multi-module-release-api-module

contrib/go.mod
dfawley: +1 please don't use 1.20.  In grpc-go, which uses this, we support 1.18+ and also haven't made any changes AFAIK that would prevent even 1.17 or maybe 1.16 from working (e.g. no use of `any` or generics).  IMO this (and the use of new language features) should stay at the oldest version you can reasonably handle supporting.

go.work
```diff
@@ -0,0 +1,10 @@
+go 1.20
```
dfawley, line 1: Sorry for the dumb question...what do go workspaces actually do for you?  Someone wanted to add a `go.work` file to our repo, but I didn't fully understand how it would help.  We have several modules and don't use a workspace, and things generally work fine.

sunjayBhatia: I think they are more useful in local development etc. when you have modules side-by-side rather than a module hierarchy like we have here, you *should* mainly be able to remove the `replace` directives but it doesn't seem (from my playing around with it as well) that we get that here because of the package/module structure

contrib/go.mod
jpeach: > > Yes, that's what I'm proposing (by suggesting that we move the "contrib" directory inside "envoy").
> 
> I think my comment above is a slightly different, it intends to mirror the structure of the envoy api folder structure so contrib and envoy packages live at the same level under a new module

I see what you mean. That would have the same effect, but nested one directory. That sounds fine to me, and will be a pretty easy change to the envoy sync script as well. I agree with you that leaving the old packages in place makes the migration path smoother.

sunjayBhatia: heres a branch of Contour using the above branch: https://github.com/sunjayBhatia/contour/tree/use-go-control-plane-multi-module

sunjayBhatia: (the above go-control-plane branch also copies the `ratelimit` module to a package in the new `api` module)

go.work
sunjayBhatia: e.g. to get the other modules in to actually use the toher modules in this repo all the way, had to do this: https://github.com/sunjayBhatia/go-control-plane/commit/4be6f62a029bd5e32eb259df831d2d3b9fb9e7cb

so I'm not sure how much the workspace is getting us

contrib/go.mod
sunjayBhatia: i think either approach will work just fine, though the upgrade/transition UX might be cleaner (no possibility for users complaining about ambiguous package paths if they are requiring mismatched versions) if the new modules have a new package path

jpeach: > (the above go-control-plane branch also copies the `ratelimit` module to a package in the new `api` module)

I'll check later, but me recollection is that the ratelimit proto it's done come from the envoy repository. So they probably need to be versioned differently and it might not make sense for them to be in the same module as the envoy api.

alecholmez:
Hey all, been away for quite a while. Starting to catch back up on this project. Did we ever come to a consensus on the way forward here? I have some cycles I can spend to push this across so we can get onto a more regular release cycle.

contrib/go.mod
craigjbass: @alecholmez it sounds like this is the most contentious point within this change set.

go.work
alecholmez: Is the decision to ultimately remove the workspace? I'm fine with that

valerian-roche:
I merged head, but my understanding is that we need to create the tags for the sub-modules first in order to allow proper importing. Trying to import in a project with the current setup, I get 
```
 % go mod tidy
go: downloading github.com/envoyproxy/go-control-plane v0.12.1-0.20240804223412-e9538224ae87
go: github.com/envoyproxy/go-control-plane@v0.12.0 requires
	github.com/envoyproxy/go-control-plane/envoy@v0.0.0-00010101000000-000000000000: invalid version: unknown revision 000000000000
```

In this case this PR would need:
 - first create the changes within the submodules, and tag them
 - this would require to follow some specific order, to add the different steps (envoy, then contrib and ratelimit, then the rest)
 - update tooling to allow consistency. I'm still unclear how we can maintain consistency in tagging with a multi-module repository. We can auto-tag envoy and contrib when syncing from envoy, but we'd need to still manually update all the other modules

go.work
```diff
@@ -0,0 +1,9 @@
+go 1.21
+
+use (
+	./
+	./contrib
+	./envoy
```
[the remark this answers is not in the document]
valerian-roche: My bad, forgot to re-add it. Had difficulties to have go mod tidy work on it

go.work
```diff
@@ -0,0 +1,10 @@
+go 1.21
```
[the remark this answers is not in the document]
dfawley: And the related issue that ultimately led to this addition:

https://github.com/golang/go/issues/53502

valerian-roche (approved):
LGTM, can we set this PR as ready for review?
I believe we should be able to push on this now. 
Thanks for all the work on this!

valerian-roche:
> There is one last part to be added which is the prepare-release See https://github.com/open-telemetry/opentelemetry-collector/blob/main/Makefile#L355-L393
> 
> Or https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/Makefile#L413-L426 For examples , if you'd like to take it from here as that's the crucial part that releasers will need to manipulate

I pushed a few changes, and raised a separate PR to actually create the first releases. 
I'd then go and update the envoy sync jobs to tag when possible

PR #718, opened 2023-06-10
New cache interface proposal

valerian-roche:
This draft PR is proposing a new interface for the Cache package
The main goals are:
 - abstract away the request. This makes the interface more generic while also avoiding having conflicting resource lists
 - pass a subscription state to get the current stateful state of the stream for the given type. This is needed also for sotw for some behaviors
 - make private all internals of the Cache to limit the API scope
 - make both interfaces more compatible to limit the amount of duplication between sotw and delta. From a Cache perspective they should behave the same in the end. It's not fully there yet but closing on it

This PR is not yet finalized (e.g. it doesn't compile currently), but it shows a first proposal. It brings a lot of change and the final implementation would likely be broken in pieces


pkg/cache/v3/cache.go
```diff
@@ -83,299 +116,44 @@ type Cache interface {
...
 	// Get the version in the Response.
 	GetVersion() (string, error)
 
+	// Get the list of resources included in the response.
+	GetReturnedResources() []string
+
 	// Get the context provided during response creation.
 	GetContext() context.Context
+
+	// Get the Constructed DiscoveryResponse
```
valerian-roche, line 130: Imo those should not be in Cache, as those depend on the server and the cache does not really need to know what requests/responses are passed
For now I kept them there to limit the impact

pkg/cache/v3/delta.go
```diff
@@ -16,73 +16,292 @@ package cache
...
+	"sync/atomic"
 
+	discovery "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
 	"github.com/envoyproxy/go-control-plane/pkg/cache/types"
-	"github.com/envoyproxy/go-control-plane/pkg/server/stream/v3"
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/anypb"
 )
 
+type resourceWithNameAndVersion interface {
```
valerian-roche, line 29: The idea is to avoid constraints on the resources that are actually not needed. Any added resource version in the cache is already set with a name
For the version it avoids having to rehash the resource. In the future we could also avoid the double serialization we currently do, which is quite expensive in CPU

pkg/cache/v3/delta.go
```diff
@@ -16,73 +16,292 @@ package cache
...
+	for _, res := range resources {
+		names = append(names, res.getName())
+	}
+	return names
+}
+
+var _ DeltaResponse = &rawDeltaResponse{}
+
+// RawDeltaResponse is a pre-serialized xDS response that utilizes the delta discovery request/response objects.
+type rawDeltaResponse struct {
```
valerian-roche, line 46: Made private. This allows us to evolve Cache implementations with less impact on users

pkg/server/delta/v3/server.go
```diff
@@ -252,10 +267,13 @@ func (s *server) unsubscribe(resources []string, streamState *stream.StreamState
 			// * a resource is explicitly unsubscribed by name
 			// Then the control-plane must return in the response whether the resource is removed (if no longer present for this node)
 			// or still existing. In the latter case the entire resource must be returned, same as if it had been created or updated
-			// To achieve that, we mark the resource as having been returned with an empty version. While creating the response, the cache will either:
+			// To achieve that, we mark the resource as having been returned with a fake version. While creating the response, the cache will either:
 			// * detect the version change, and return the resource (as an update)
 			// * detect the resource deletion, and set it as removed in the response
-			streamState.GetResourceVersions()[resource] = ""
+			streamState.GetKnownResources()[resource] = "unsubscribed"
```
valerian-roche, line 273: This is for sure hacky (though not much more than before), while allowing the use of "" for the standard case of "non-existent" resource

PR #719, opened 2023-06-12
cache: add request TypeUrl and ResourceNames to warn log

110y:
- Add `request.TypeUrl` and `request.ResourceNames` to the WARN log regarding the resource mismatch for ease of debugging.


jpeach (approved):
Looks like a nice improvement.

Issue #705, opened 2023-05-30
Order resources for guaranteed eventual consistency in Delta + ADS mode


arkodg:
Make changes similar to https://github.com/envoyproxy/go-control-plane/pull/544 when the server is running with ADS and Delta (DELTA_GRPC) enabled to ensure resources are ordered to ensure no traffic drops are seen while updating related resources at the same time.
Relates to https://github.com/envoyproxy/go-control-plane/issues/526

PR #752, opened 2023-08-14
Enable ordered responses for ADS delta watches

zhaohuabing:
Fix https://github.com/envoyproxy/go-control-plane/issues/705

### Why we need this?
go-control-plane doesn't guarantee the orders of responses for delta ADS. If there are two requests, the response to the second request may be sent before the first request. This causes problems for us since our application generates xDS resources for clusters and listeners in sequence. Sometimes, the the listener resources are sent to Envoy before the cluster resources. As a result, Envoy complains that it can't find referenced clusters in the Listener configuration and fail to apply the configuration.

### How this PR fix this issue?
This PR follow an approach similar to https://github.com/envoyproxy/go-control-plane/pull/544 to enable ordered responses for ADS delta watches.
* Sort the watches to ensure that the responses are handled in the correct order(Endpoint, Cluster, Route, Listener ...)
* Send all the responses to the shared `deltaMuxedResponses` channel to guarantee that the orders won't be changed
* Since responses and requests are handled in a single go routine, to avoid deadlocks, the response channel is purged before processing a request, and the `deltaMuxedResponses` channel is created with a buffer size of 2x the number of types.


arkodg:
ptal @valerian-roche @alecholmez

pkg/server/delta/v3/server.go
```diff
@@ -74,10 +85,24 @@ func (s *server) processDelta(str stream.DeltaStream, reqCh <-chan *discovery.De
 	// a collection of stack allocated watches per request type
 	watches := newWatches()
 
+	// use a single go routine to send responses to the muxedResponses channel to retain resource orders
+	go func() {
+		for {
+			select {
```
valerian-roche, line 91: Minor: unneeded select if there is a single case

valerian-roche (commented):
This is only an initial review of the PR. I need to spend more time on the impact of this new shared channel.
My initial analysis is that there is a potential deadlock case as a watch can have queued a response prior to being cancelled. In some edge cases when the client has sent a lot of requests in a short sequence, I believe this channel could end up overflowing the muxed one and then itself. The previous model guarantees that the muxed channel being filled does not block other processes (as they each dequeue within their own goroutine), but here it is no longer the case

pkg/server/delta/v3/server.go
```diff
@@ -66,6 +74,9 @@ func NewServer(ctx context.Context, config cache.ConfigWatcher, callbacks Callba
 }
 
 func (s *server) processDelta(str stream.DeltaStream, reqCh <-chan *discovery.DeltaDiscoveryRequest, defaultTypeURL string) error {
+	// create a sharedChan for the watches to send ordered responses to
+	sharedChan := make(chan cache.DeltaResponse, types.UnknownType)
```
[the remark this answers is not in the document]
zhaohuabing: You're right, we don't need to create the shared channel if the server is not in ordered mode.

pkg/server/delta/v3/server.go
```diff
@@ -74,10 +85,24 @@ func (s *server) processDelta(str stream.DeltaStream, reqCh <-chan *discovery.De
 	// a collection of stack allocated watches per request type
 	watches := newWatches()
 
+	// use a single go routine to send responses to the muxedResponses channel to retain resource orders
+	go func() {
+		for {
+			select {
+			case resp, more := <-sharedChan:
```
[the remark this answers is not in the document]
zhaohuabing: Yes, we should use the existing deltaMuxedResponses channel rather than creating a new one.

pkg/server/delta/v3/server.go
```diff
@@ -184,16 +213,26 @@ func (s *server) processDelta(str stream.DeltaStream, reqCh <-chan *discovery.De
 			s.subscribe(req.GetResourceNamesSubscribe(), &watch.state)
 			s.unsubscribe(req.GetResourceNamesUnsubscribe(), &watch.state)
 
-			watch.responses = make(chan cache.DeltaResponse, 1)
+			if ordered {
+				// Use the shared channel for ordered responses
+				watch.responses = sharedChan
```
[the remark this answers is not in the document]
zhaohuabing: I was thinking we need to handle ordered and non-ordered responses at the same time, which is wrong. The ADS server only runs in one mode: ordered or non-ordered.  So we can just reuse the existing watches.deltaMuxedResponses as the shared channel for ordered responses.

pkg/server/delta/v3/server.go
```diff
@@ -184,16 +195,25 @@ func (s *server) processDelta(str stream.DeltaStream, reqCh <-chan *discovery.De
 			s.subscribe(req.GetResourceNamesSubscribe(), &watch.state)
 			s.unsubscribe(req.GetResourceNamesUnsubscribe(), &watch.state)
 
-			watch.responses = make(chan cache.DeltaResponse, 1)
+			if ordered {
+				// Use the shared channel to keep the order of responses.
+				watch.UseSharedResponseChan(watches.deltaMuxedResponses)
```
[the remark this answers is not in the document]
zhaohuabing: @valerian-roche If the responses and requests are handled in two different go routines, will that solve the deadlock issue?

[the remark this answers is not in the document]
zhaohuabing: @valerian-roche If the responses and requests are handled in two different go routines, will that solve the deadlock issue?

zhaohuabing:
Some tests failed because of this new change. If this approach is acceptable, I'll go on and fix them.

pkg/server/delta/v3/server.go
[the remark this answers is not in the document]
zhaohuabing: @valerian-roche I tried to follow your suggestion to guarantee that the responses ar processed prior to new requests.

pkg/server/delta/v3/server.go
```diff
@@ -142,6 +142,14 @@ func (s *server) processDelta(str stream.DeltaStream, reqCh <-chan *discovery.De
 				return status.Errorf(codes.Unavailable, "empty request")
 			}
 
+			// make sure responses are processed prior to new requests to avoid deadlock
+			if len(watches.deltaMuxedResponses) > 0 {
+				go func() {
```
valerian-roche, line 147: This can potentially spin for a while (or forever) based on the internal of this select (which is undefined behavior). I think the implementation done in sotw (basically move reply in a separate method, and call here if needed) is more robust

pkg/server/delta/v3/watches.go
```diff
@@ -19,7 +19,7 @@ func newWatches() watches {
 	// deltaMuxedResponses needs a buffer to release go-routines populating it
 	return watches{
 		deltaWatches:        make(map[string]watch, int(types.UnknownType)),
-		deltaMuxedResponses: make(chan cache.DeltaResponse, int(types.UnknownType)),
+		deltaMuxedResponses: make(chan cache.DeltaResponse, int(types.UnknownType)*2),
```
valerian-roche, line 22: Can you add a comment on why we provide 2 times the buffering here?

pkg/server/delta/v3/server.go
```diff
@@ -184,16 +192,9 @@ func (s *server) processDelta(str stream.DeltaStream, reqCh <-chan *discovery.De
 			s.subscribe(req.GetResourceNamesSubscribe(), &watch.state)
 			s.unsubscribe(req.GetResourceNamesUnsubscribe(), &watch.state)
 
-			watch.responses = make(chan cache.DeltaResponse, 1)
+			watch.responses = watches.deltaMuxedResponses
```
valerian-roche, line 195: I don't think `responses` is actually needed anymore in this case, as it was mostly used to keep track of closing to clean-up the other goroutine

pkg/server/delta/v3/server.go
zhaohuabing: @valerian-roche do you mean purge the response channel before handling the request?

zhaohuabing: Thought it again. This time I purge the response channel before processing a request, so we don't rely on the behavior of select which is undefined.

pkg/server/delta/v3/watches.go
zhaohuabing: add comment:

        
        // because deltaMuxedResponses can be populated by an update from the cache
	// and a request from the client, we need to create the channel with a buffer
	// size of 2x the number of types to avoid deadlocks.

pkg/server/delta/v3/server.go
zhaohuabing: Yes, we should just remove it.

valerian-roche (approved):
LGTM, thanks a lot for your patience on this!

Were you able to test this on a running environments? I might be able to test it on some of our test systems next week if you cannot easily run it

zhaohuabing:
> LGTM, thanks a lot for your patience on this!
> 
> Were you able to test this on a running environments? I might be able to test it on some of our test systems next week if you cannot easily run it
> 
> Can you expand in the description describing a bit the implementation and the reasoning behind if people need to take a look later on?

@valerian-roche Thanks. I can test it since I have already encountered this issue in my local development environment. But it may take a few days since I'll be attending KubeCon CN next week and I'm preparing on my talk. 

Description updated with reasons and hows.

zhaohuabing:
@valerian-roche Tested in my local development environment.

arkodg:
thanks @zhaohuabing for fixing this, can we get this merged ? Envoy Gateway (based on delta xDS) needs this

valerian-roche (approved):

PR #762, opened 2023-08-24
Remove special cases from GetResourceName.

jpeach:
Remove the unnecessary type checks from GetResourceName. Modern resource types should always have a `GetName` accessor, so the only real special case is for clusters.


jpeach:
@sunjayBhatia Do you mind taking a look?

valerian-roche (approved):

sunjayBhatia (approved):
lgtm, test coverage shows no issues removing these cases 👍🏽

PR #769, opened 2023-08-30
Add logic for get nodeIDs from snapshot cache

zvlb:
Hello. 

In my case, I use many nodeIDs when forming the cache. During processing, it's necessary to know which nodeIDs are present in the cache. 

Please consider reviewing this PR.

### DIFF
--- a/pkg/cache/v3/simple.go
+++ b/pkg/cache/v3/simple.go
@@ -83,6 +83,9 @@ type SnapshotCache interface {
 	// ClearSnapshot removes all status and snapshot information associated with a node.
 	ClearSnapshot(node string)
 
+	// GetNodeIDs gets the all node ID in cache.
+	GetNodeIDs() []string
+
 	// GetStatusInfo retrieves status information for a node ID.
 	GetStatusInfo(string) StatusInfo
 
@@ -604,6 +607,20 @@ func (cache *snapshotCache) Fetch(ctx context.Context, request *Request) (Respon
 	}
 
 	return nil, fmt.Errorf("missing snapshot for %q", nodeID)
+}
+
+// GetNodeIDs gets the all node ID in cache.
+func (cache *snapshotCache) GetNodeIDs() []string {
+	cache.mu.RLock()
+	defer cache.mu.RUnlock()
+
+	nodeIDs := []string{}
+
+	for nodeID := range cache.snapshots {
+		nodeIDs = append(nodeIDs, nodeID)
+	}
+
+	return nodeIDs
 }
 
 // GetStatusInfo retrieves the status info for the node.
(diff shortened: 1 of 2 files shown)


zvlb:
Someone?

jpeach:
> Hello.
> 
> In my case, I use many nodeIDs when forming the cache. During processing, it's necessary to know which nodeIDs are present in the cache.


Can you just keep track of the node IDs from when you formed the cache?

zvlb:
> Can you just keep track of the node IDs from when you formed the cache?

I'm currently doing just that. However, I'm facing issues with cases where entities are evicted from the cache, and consequently, there are no instructions left for a specific nodeID.

To handle this "properly," I need to fully replicate the cache in my code and update the go-control-plane cache through this intermediary layer. It seems to me that this is quite redundant.

jpeach:
Hopefully next week I'll have enough time to read through the code and understand your problem.

zvlb:
@jpeach Hi. Did you find time to review this PR?

pkg/cache/v3/simple.go
```diff
@@ -606,6 +609,20 @@ func (cache *snapshotCache) Fetch(ctx context.Context, request *Request) (Respon
 	return nil, fmt.Errorf("missing snapshot for %q", nodeID)
 }
 
+// GetNodeIDs gets the all node ID in cache.
+func (cache *snapshotCache) GetNodeIDs() []string {
+	cache.mu.RLock()
+	defer cache.mu.RUnlock()
+
+	nodeIDs := []string{}
```
jpeach, line 617:
```suggestion
	nodeIDs := make([]string, 0, Len(cache.snapshots))
```

Micro-optimize the allocation :)

pkg/cache/v3/simple.go
```diff
@@ -606,6 +609,20 @@ func (cache *snapshotCache) Fetch(ctx context.Context, request *Request) (Respon
 	return nil, fmt.Errorf("missing snapshot for %q", nodeID)
 }
 
+// GetNodeIDs gets the all node ID in cache.
+func (cache *snapshotCache) GetNodeIDs() []string {
```
jpeach, line 613: Since we have `GetStatusKeys`, how about following the same pattern and calling this `GetSnapshotKeys`?

pkg/cache/v3/simple.go
jpeach: In fact, you could copy the `GetStatusKeys` implementation directly here :)

jpeach (commented):
Hi @zvlb, sorry for the delay.

I think that this change makes sense in terms of consistency. Since the snapshot cache stores maps of snapshots and status separately, it would be consistent to be able to get the keys for both of those.

I was originally concerned that adding to the `SnapshotCache` interface would break any external projects that implemented it. However, we don't have any APIs that accept this interface as a parameter, so I think that adding to it is find in practice.

However, the only time where the snapshot and status keys would differ is when a snapshot is set, but there is nothing that ever does a CreateWatch on it. I think that implies that you'd be in a situation where no envoy is requesting config for that nodeID?

I also don't think I understand your eviction scenario. IIUC, snapshotCache doesn't evict anything, so do you have some other code that is involved here?

PR #763, opened 2023-08-24
Document that well-known names are no longer needed.

jpeach:
Envoy hasn't needed well-known filter names for a long while. Document that fact so that this package stops confusing people.

This fixes #756.

PR #771, opened 2023-09-04
Fix lint warnings on deprecated wellknown names.

jpeach:
Since we deprecated wellknown names in #763, fix the lint warnings that resulted by removing our internal use of the package.


jpeach:
We could also undo the deprecation. I don't have strong opinions about this :)

jpeach:
@alecholmez @sunjayBhatia @valerian-roche Any thoughts on this?

internal/example/resource.go
```diff
@@ -118,7 +117,7 @@ func makeHTTPListener(listenerName string, route string) *listener.Listener {
 			},
 		},
 		HttpFilters: []*hcm.HttpFilter{{
-			Name:       wellknown.Router,
+			Name:       "http-router",
```
valerian-roche, line 120: I'm overall fine with it, but is it on purpose not matching the previous content?

valerian-roche (approved):

PR #774, opened 2023-09-10
ci(deps): group k8s dependencies in /examples/dyplomat

mmorel-35:
This PR allows to group the update of k8s.io in a single PR.


jpeach (approved):
Nice! Thanks @mmorel-35

PR #778, opened 2023-09-13
docs: fix spell error

ankur-anand:
Function Name in Documentation is incorrect. OnStreamDelatResponse should be OnStreamDeltaResponse


jpeach (approved):
Thanks @ankur-anand !

PR #794, opened 2023-10-11
Provide callback to decide if nack should be retried

### LINKED ISSUE #46: Avoid repeated push of rejected configuration
If envoy rejects a config, the snapshot cache will attempt to push the same config immediately after envoy requests it again. If envoy does not limit its requests, the server will drive envoy into a loop of request-receive-reject, potentially causing unnecessary CPU load.

akshaysngupta:
Currently, when go-control-plane receives a NACK from Envoy, it proceeds by resending the same request again. This causes Envoy to send the NACK causing a never-ending loop.

Some have suggested to that rate limiting Envoy can help but this is not required when it is apparent that error will not go away.

To give implementors ability to select when they want to retry sending the response, This PR adds `OnStreamResponseNacked` and `OnStreamDeltaResponseNacked` callbacks which are called when Envoy responds with a `NACK` for a previous response sent by go-control-plane.

Fixes #46


jpeach:
I'm not sure about this in principle. If a response is rejected by the client, then isn't the right approach to actually remove the invalid configuration from the cache?

Issue #800, opened 2023-10-16
the resource order is not in line with the description in the xds doc


zhaohuabing:
It looks like the current order is not in line with the description in the xds doc.

The order in the current implementation: Listener -> Route -> Cluster -> Endpoint
According to xds doc, the order should be like: Cluster -> Endpoint -> Listener -> Route 

References:

https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#eventual-consistency-considerations

```
In general, to avoid traffic drop, sequencing of updates should follow a make before break model, wherein:

CDS updates (if any) must always be pushed first.

EDS updates (if any) must arrive after CDS updates for the respective clusters.

LDS updates must arrive after corresponding CDS/EDS updates.

RDS updates related to the newly added listeners must arrive after CDS/EDS/LDS updates.

VHDS updates (if any) related to the newly added RouteConfigurations must arrive after RDS updates.

Stale CDS clusters and related EDS endpoints (ones no longer being referenced) can then be removed.
```

https://github.com/envoyproxy/go-control-plane/blob/52093cd01fd1a94d243081f6dec566db6987c145/pkg/cache/v3/order_test.go#L66-L71

AmitKatyal-Sophos:
@zhaohuabing Thanks for creating the bug

PR #801, opened 2023-10-17
fix resource order

zhaohuabing:
fix https://github.com/envoyproxy/go-control-plane/issues/800

According to xds doc, the order should be like: Cluster -> Endpoint -> Listener -> Route

References:

https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#eventual-consistency-considerations

@valerian-roche @AmitKatyal-Sophos

### DIFF
--- a/pkg/cache/types/types.go
+++ b/pkg/cache/types/types.go
@@ -43,12 +43,12 @@ type ResponseType int
 // ADS expects things to be returned in a specific order.
 // See the following issue for details: https://github.com/envoyproxy/go-control-plane/issues/526
 const (
-	Endpoint ResponseType = iota
-	Cluster
-	ScopedRoute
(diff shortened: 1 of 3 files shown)


AmitKatyal-Sophos:
can you please correct the ordering comment as well - https://github.com/envoyproxy/go-control-plane/pull/801/files#diff-f6d778fc9cc5f603b70f55a4a04fefa7db6183a04be349d13ab1f8e68ee1ae71R65

jpeach:
ref #544 

cc @alecholmez @sunjayBhatia

valerian-roche (approved):

sunjayBhatia:
yeah I was a bit curious about the order from that original PR tbh

PR #815, opened 2023-11-01
ci: fix and enable more linters

mmorel-35:


pkg/cache/v3/cache_test.go
```diff
@@ -78,27 +79,27 @@ func TestHeartbeatResponseGetDiscoveryResponse(t *testing.T) {
 	}
 
 	discoveryResponse, err := resp.GetDiscoveryResponse()
-	assert.Nil(t, err)
-	assert.Equal(t, discoveryResponse.VersionInfo, resp.Version)
-	assert.Equal(t, len(discoveryResponse.Resources), 1)
-	assert.False(t, isTTLResource(discoveryResponse.Resources[0]))
+	require.NoError(t, err)
+	assert.Equal(t, discoveryResponse.GetVersionInfo(), resp.Version)
+	assert.Len(t, discoveryResponse.GetResources(), 1)
```
jpeach, line 84:
```suggestion
	require.Len(t, discoveryResponse.GetResources(), 1)
```

We should require here because the next assertion depends on it.

pkg/cache/v3/resource_test.go
```diff
@@ -57,15 +58,15 @@ var (
...
-	assert.NoError(t, testExtensionConfig.Validate())
+	require.NoError(t, testEndpoint.Validate())
+	require.NoError(t, testCluster.Validate())
+	require.NoError(t, testRoute.Validate())
+	require.NoError(t, testScopedRoute.Validate())
+	require.NoError(t, testVirtualHost.Validate())
+	require.NoError(t, testListener.Validate())
+	require.NoError(t, testScopedListener.Validate())
+	require.NoError(t, testRuntime.Validate())
+	require.NoError(t, testExtensionConfig.Validate())
```
jpeach, line 69: Nit - `assert` is correct here. If the linter insists on this change, we can live with it though :)

pkg/cache/v3/simple_test.go
```diff
@@ -591,7 +595,8 @@ func TestSnapshotSingleResourceFetch(t *testing.T) {
 
 	resp, err := c.Fetch(context.Background(), &discovery.DiscoveryRequest{
 		TypeUrl:       durationTypeURL,
-		ResourceNames: []string{"one-second"}},
+		ResourceNames: []string{"one-second"},
+	},
 	)
```
jpeach, line 600:
```suggestion
	})
```

pkg/cache/v3/snapshot_test.go
```diff
@@ -54,7 +54,8 @@ func TestClusterWithMissingEndpointIsInconsistent(t *testing.T) {
 
 func TestListenerWithMissingRoutesIsInconsistent(t *testing.T) {
 	if snap, _ := cache.NewSnapshot(fixture.version, map[rsrc.Type][]types.Resource{
-		rsrc.ListenerType: {testListener}},
+		rsrc.ListenerType: {testListener},
+	},
 	); snap.Consistent() == nil {
```
jpeach, line 59:
```suggestion
	}); snap.Consistent() == nil {
```

pkg/integration/ttl_integration_test.go
```diff
@@ -118,25 +119,25 @@ func TestTTLResponse(t *testing.T) {
 func isFullResponseWithTTL(t *testing.T, response *envoy_service_discovery_v3.DiscoveryResponse) {
 	t.Helper()
 
-	assert.Len(t, response.Resources, 1)
-	r := response.Resources[0]
+	assert.Len(t, response.GetResources(), 1)
```
jpeach, line 122:
```suggestion
	require.Len(t, response.GetResources(), 1)
```

pkg/integration/ttl_integration_test.go
```diff
@@ -118,25 +119,25 @@ func TestTTLResponse(t *testing.T) {
...
+	assert.NotNil(t, resource.GetTtl())
+	assert.NotNil(t, resource.GetResource())
 }
 
 func isHeartbeatResponseWithTTL(t *testing.T, response *envoy_service_discovery_v3.DiscoveryResponse) {
 	t.Helper()
 
-	assert.Len(t, response.Resources, 1)
-	r := response.Resources[0]
+	assert.Len(t, response.GetResources(), 1)
```
jpeach, line 135:
```suggestion
	require.Len(t, response.GetResources(), 1)
```

pkg/server/v3/server_test.go
```diff
@@ -391,91 +392,91 @@ func TestFetch(t *testing.T) {
 	s := server.NewServer(context.Background(), config, cb)
 	out, err := s.FetchEndpoints(context.Background(), &discovery.DiscoveryRequest{Node: node})
 	assert.NotNil(t, out)
-	assert.NoError(t, err)
+	require.NoError(t, err)
```
jpeach, line 395: Here and below, we can use `assert` instead of `require`. Again, if the linter insists, this is fine.

jpeach (changes requested):
Thanks @mmorel-35, this looks great. I just have a few trivial suggestions.

Could you please also update the golangci-lint version in the Makefile?

pkg/cache/v3/resource_test.go
mmorel-35: Evey assert.NoError to require.NoError change was required by the linter.`

jpeach (approved):

PR #817, opened 2023-11-09
ci: enable more linters

mmorel-35:


jpeach (approved):
I think that this makes sense. Thanks @mmorel-35!

Issue #524, opened 2021-12-13
Switch to new Go protobuf v2 module


jpeach:
Upstream switched to a new Go protobuf module, [protobuf v2](https://go.dev/blog/protobuf-apiv2). go-control-plane should fully switch over to this so that client projects can switch too. I think the main change that will be needed is to switch the protoc toolchain over to generate wrapper code with the new module.

alecholmez:
+1 for this. We looked into this previously and I believe the action items were:
- Update rules_go
- Update Envoy's internal bazel pipeline
- Update the build script: https://github.com/envoyproxy/envoy/blob/main/tools/api/generate_go_protobuf.py

jpeach:
>     * Update rules_go: [proto: add go_proto_apiv2 and go_grpc_apiv2 compilers bazelbuild/rules_go#2522](https://github.com/bazelbuild/rules_go/issues/2522)

Oof, that does not look promising ...

alecholmez:
> Oof, that does not look promising ...

Yea... this is a big rabbit hole I went down a while back, might be worth revisiting though

alecholmez:
I have a PR open for this now: https://github.com/envoyproxy/envoy/pull/21790

alecholmez:
Okay looks like this pr: https://github.com/envoyproxy/envoy/pull/21790 got merged so we can now proceed here.

jpeach:
> Okay looks like this pr: [envoyproxy/envoy#21790](https://github.com/envoyproxy/envoy/pull/21790) got merged so we can now proceed here. @jpeach would you be interested in picking this back up?

I can probably take a look in a week or so.

yckbilly1929:
https://github.com/envoyproxy/go-control-plane/pull/607

Would like to know if we will resolve this issue together with this PR?

PR #819, opened 2023-11-19
Use the v2 protobuf module internally.

jpeach:
There's only one place where we use the protobuf APIs, so switch to the v2 protobuf module.

This updates #524.


sunjayBhatia (approved):
LGTM, all the other usages of github.com/golang/protobuf seem to be from the generated APIs

valerian-roche (approved):
LGTM, but as it changes a public interface we might want to reflect it in the version when releasing

jpeach:
> LGTM, but as it changes a public interface we might want to reflect it in the version when releasing

Agreed. I'm going to look at moving the generated protos to V2 as well, so maybe we can cut a release after that?

PR #849, opened 2024-01-05
use replace in examples and xdsmatcher

mmorel-35:


phlax:
can you give some explanation of change (apologies if its obvious)

mmorel-35:
go.mod requires pined versions.
When there are multiples modules on a repository that are interdependant, the use of replace allows to use the up to date version of the local modules in the go.mod file.

phlax:
thnks for explanation - i have seen it used in Envoy and wasnt 100% clear about what it did

phlax:
> Why was it not merged ?

fell through the cracks i think

> Shall I reopen this ?

yep

PR #856, opened 2024-01-10
[sotw][issue-540] Return full state when applicable for watches in linear cache

### LINKED ISSUE #540: In LinearCache, respond can't tell the caller is from update or delete.

valerian-roche:
This PR is currently based on #854, and include its diff too, but in a separate commit

Fixes #540 

This PR is fixing issue #540: the xds protocol defines that for the resource types `Listener` and `Cluster`, when using sotw watches, the cache must return as part of the response **all** existing resources matching the requests if not wildcard. 
Currently the control-plane will only return the modified resources on cache updates, and the protocol states that the client should interpret such a response as a deletion

In simple cache this distinction is not applicable as we currently always return all resources. While we may want to improve this behavior in the future (e.g. to avoid sending all clusters if only one was updated in delta, or to send a single route configuration in sotw on update), this is out of scope of this PR as this requires keeping more state

### DIFF
--- a/pkg/cache/v3/linear.go
+++ b/pkg/cache/v3/linear.go
@@ -27,7 +27,7 @@ import (
 	"github.com/envoyproxy/go-control-plane/pkg/server/stream/v3"
 )
 
-type watches = map[chan Response]struct{}
+type watches = map[ResponseWatch]struct{}
 
 // LinearCache supports collections of opaque resources. This cache has a
 // single collection indexed by resource names and manages resource versions
@@ -113,7 +113,7 @@ func NewLinearCache(typeURL string, opts ...LinearCacheOption) *LinearCache {
 	return out
 }
 
-func (cache *LinearCache) respond(value chan Response, staleResources []string) {
+func (cache *LinearCache) respond(watch ResponseWatch, staleResources []string) {
 	var resources []types.ResourceWithTTL
 	// TODO: optimize the resources slice creations across different clients
 	if len(staleResources) == 0 {
@@ -130,8 +130,8 @@ func (cache *LinearCache) respond(value chan Response, staleResources []string) 
 			}
 		}
 	}
-	value <- &RawResponse{
-		Request:   &Request{TypeUrl: cache.typeURL},
+	watch.Response <- &RawResponse{
+		Request:   watch.Request,
 		Resources: resources,
 		Version:   cache.getVersion(),
 		Ctx:       context.Background(),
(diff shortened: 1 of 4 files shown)


pkg/cache/v3/linear.go
```diff
@@ -327,8 +341,12 @@ func (cache *LinearCache) CreateWatch(request *Request, _ stream.StreamState, va
 	case err != nil:
 		stale = true
 		staleResources = request.GetResourceNames()
+		cache.log.Debugf("Watch is stale as version failed to parse %s", err.Error())
```
atollena, line 344: nit: add a colon to separate error from the string: `Watch is stale as version failed to parse: %s`

Also I see some log lines are prefixed with `[linear cache]`, you may want to add that here and at L348, too.

pkg/cache/v3/resource.go
```diff
@@ -97,6 +97,23 @@ func GetResourceName(res types.Resource) string {
 	}
 }
 
+// ResourceRequiresFullStateInSotw indicates whether when building the reply in Sotw,
+// the response must include all existing resources or can return only the modified ones
+func ResourceRequiresFullStateInSotw(typeURL resource.Type) bool {
+	// From https://www.envoyproxy.io/docs/envoy/v1.28.0/api-docs/xds_protocol#grouping-resources-into-responses,
+	// when using sotw the control-plane MUST return all requested resources (or simply all if wildcard)
+	// for some types. This is relied on by xds-grpc which is explicitly requesting clusters but expect
+	// to receive all existing resources
```
atollena, line 106: In a way, yes, this is a specificity of xDS gRPC, because xDS gRPC uses resource names with LDS and CDS, whereas IIUC, envoy does not. For cases where Envoy does (ODCDS?) then I don't know how resource deletions are handled. I would add that for gRPC clients, it not including a resource in LDS or CDS indicates that the resource has been deleted. See https://github.com/grpc/proposal/blob/master/A53-xds-ignore-resource-deletion.md#background which has more details.

atollena (approved):

atollena (approved):

Issue #875, opened 2024-02-07
Deadlock in Delta XDS if number of resources is more than 10


lobkovilya:
In [Kuma](https://github.com/kumahq/kuma) we're implementing the `ResourceSnapshot` interface and our snapshot has more than 20 distinct resource types. 

After upgrading to `v0.12.0` we see deadlocks on the server and it seems like it was caused by https://github.com/envoyproxy/go-control-plane/pull/752. AFAICT the potential deadlock was fixed by [increasing the channel capacity](https://github.com/envoyproxy/go-control-plane/pull/752/files#diff-2239e4e23a62885d2c0b6d6d6cc0b453adf1886d26f445d601b2f1dc57131360R26), but it's not fixed for us since our snapshot has more resource types.

I'm looking for help or advice on what'd be the best way to fix it? It doesn't seem right that `server` depends on the number of resource types in the snapshot to work properly.

slonka:
There is a PR open so I wouldn't call it stale.

lobkovilya:
Some additional context.

### Why the deadlock appeared after https://github.com/envoyproxy/go-control-plane/pull/752?

The PR replaces `a response channel per watch` model with `a response channel per all watches`. This was important to support ADS ordering, but at the same time, xDS was affected by the change.

### Why there is no deadlock in SotW?

SotW has 2 different implementations for [xds](https://github.com/envoyproxy/go-control-plane/blob/main/pkg/server/sotw/v3/xds.go) and [ads](https://github.com/envoyproxy/go-control-plane/blob/main/pkg/server/sotw/v3/ads.go). ADS has `a response channel per all watches` while xDS still has `a response channel per watch`.

### How the deadlock can be solved in Delta?

1. Introduce proper xDS implementation, so that the Delta server has both `xds.go` and `ads.go` similar to the SotW server.
2. Let the capacity of a response channel be configurable (see my PR https://github.com/envoyproxy/go-control-plane/pull/876)
3. Build some kind of an "unbounded channel" abstraction when the writer never blocks and the internal buffer is reallocated and shrunk dynamically (I know it's a tiny bit crazy idea 😅).

Probably there are some other options I haven't thought of. @valerian-roche @jpeach @alecholmez please take a look, I'd appreciate any input here, and I'd be happy to contribute the solution.

valerian-roche:
Hey, sorry for the delay, I have not had much capacity to work on this recently. 
[...]

PR #876, opened 2024-02-08
Option for the delta server to override the channel capacity and avoid deadlocks

lobkovilya:
Please check the context at https://github.com/envoyproxy/go-control-plane/issues/875.

This is quite a naive and straightforward way to solve it, but it'd work perfectly fine for us. Please let me know your thoughts, I'm happy to discuss a better solution.

### DIFF
--- a/pkg/server/config/config.go
+++ b/pkg/server/config/config.go
@@ -1,15 +1,19 @@
 package config
 
+import "github.com/envoyproxy/go-control-plane/pkg/cache/types"
+
 // Opts for individual xDS implementations that can be
 // utilized through the functional opts pattern.
 type Opts struct {
 	// If true respond to ADS requests with a guaranteed resource ordering
-	Ordered bool
+	Ordered               bool
+	DistinctResourceTypes int
 }
 
 func NewOpts() Opts {
 	return Opts{
-		Ordered: false,
+		Ordered:               false,
+		DistinctResourceTypes: int(types.UnknownType),
 	}
 }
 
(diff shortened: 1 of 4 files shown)


valerian-roche:
Hey 👋, I do believe it would solve this issue, but this raises concerns in my opinion with leaking too many internals of the server implementation. It would be very brittle to require users to set this, especially if it does not fail at startup but only in some edge cases. We might be able to do it in a more efficient way based on the current number of distinct opened watches in the given stream, or by using the old model when watches do not watch the resources with an ADS order requirement.

lobkovilya:
Hey @valerian-roche! Yeah, I agree, that the solution is far from ideal. 

> We might be able to do it in a more efficient way based on the current number of distinct opened watches in the given stream, or by using the old model when watches do not watch the resources with an ADS order requirement.

Do you know if something like that is already in progress or planned?

valerian-roche:
Hey, there is no ongoing work on the issue, but I may be able to take a stab at it soon.

lobkovilya:
Hey @valerian-roche, I'd like to take on this issue as we're blocked right now with `go-control-plane` upgrades. Could you please elaborate on the idea you mentioned here:
> We might be able to do it in a more efficient way based on the current number of distinct opened watches in the given stream

lukidzi:
Hi @valerian-roche,

Is it possible to include this simple fix, considering there are no other ongoing changes? This modification doesn't alter the current user behavior, as the default value remains the same, but it allows users to provide a custom size. Since there has been no progress on this issue and the change does not break anything, could we consider incorporating it into the codebase?

valerian-roche:
Hey, I'd like to not merge this. This is far more likely to create deadlocks for users if they are not correct with the value, and would be forever part of the public API. I'd much rather we address the core problem of this sized channel

PR #908, opened 2024-03-22
Apply 'go mod tidy' after API sync

mmorel-35:
@phlax ,
I believe there is no go mod tidy applied after envoy synchronisation

### DIFF
--- a/ci/sync_envoy.sh
+++ b/ci/sync_envoy.sh
@@ -25,6 +25,7 @@ build_protos () {
     cd "${ENVOY_SRC_DIR}" || exit 1
     ./ci/do_ci.sh api.go
     cd - || exit 1
+    go mod tidy
 }
 
 get_last_envoy_sha () {


ci/sync_envoy.sh
```diff
@@ -72,5 +72,6 @@ commit_changes () {
 
 
 build_protos
+go mod tidy
```
phlax, line 75: i think lets either put it to its own function or add it after the build

phlax (commented):
dont worry about testing it we can land and if there is an issue we can revert

the only question i have is what go version this is using - one way/another i think we want to make sure its pinned

phlax:
trying to rem how i tested previously - maybe i copied the workflow to trigger on pr altho cant rem if that is allowed for prs to do

phlax:
not sure - im kinda surprised this repo uses make tbh

iiuc you would mostly want to run tidy after changing deps so probs a standalone target - but i have limited go experience, so happy to defer to others with more knowledge of go workflows

sunjayBhatia (approved):
I think what you have here is sufficient, a new make target would be a bit verbose for just running `go mod tidy` and `make clean` isn't necessary, that seems more for a local dev environment where you may have old binaries etc lying around

PR #911, opened 2024-03-26
Move go mod tidy to run in sync_protos

sunjayBhatia:
Was being run in build_protos and also not git adding any changed go.mod/go.sum files

Also re-runs go mod tidy locally to fix up main


phlax (approved):
lgtm, thanks @sunjayBhatia

PR #912, opened 2024-04-01
use replace in `examples/dyplomat` and `xdsmatcher` modules

mmorel-35:


valerian-roche (approved):
Thanks for your work on this

PR #914, opened 2024-04-05
Bump vtprotobuf import

howardjohn:
Goes along with https://github.com/envoyproxy/envoy/pull/33365. This will be required to pull in that Envoy change, but it can merge before it.

### DIFF
--- a/examples/dyplomat/go.mod
+++ b/examples/dyplomat/go.mod
@@ -40,7 +40,7 @@ require (
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
 	github.com/modern-go/reflect2 v1.0.2 // indirect
 	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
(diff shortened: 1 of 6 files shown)


valerian-roche (dismissed):
LGTM 
Out of curiosity, is there a reason to not provide unmarshal in the protobuf generation? It would remain conditioned under the build tag, and it would be quite useful for people implementing services like sds or ext_authz servers to be able to use vtprotobuf also on the unmarshaling path for performance reasons

howardjohn:
> LGTM Out of curiosity, is there a reason to not provide unmarshal in the protobuf generation? It would remain conditioned under the build tag, and it would be quite useful for people implementing services like sds or ext_authz servers to be able to use vtprotobuf also on the unmarshaling path for performance reasons

I wouldn't mind adding it, mostly excluded since it wasn't terribly important to us and there were already concerns about binary size

howardjohn:
@valerian-roche the Envoy PR landed. I think that means the sync job will be broken until this lands

Issue #913, opened 2024-04-04
Goroutine leak in DeltaStreamHandler


zhiyanfoo:
go-control-plane version: 2259f2656a1f9c0622c315e58ad1e93338373ba5

In our control-plane implementation, we are seeing a significant memory leak in specific cases.
Taking a closer look the memory leak can be attributed with a goroutine leak. 
<img width="1299" alt="Screenshot 2024-04-04 at 2 20 25 PM" src="https://github.com/envoyproxy/go-control-plane/assets/8400483/e038aad6-59e0-4843-837a-f366cf2b279d">

Looking at profiles we see that we have far too many goroutines from the DeltaStreamHandler method
(significantly more than active streams).

This is currently occurring when using the delta protocol with ads from an envoy client,
targeting a snapshot cache.

The [following goroutine](https://github.com/envoyproxy/go-control-plane/blob/9c62c968b253792310a10dda82789b43cb53e812/pkg/server/delta/v3/server.go#L232) has the potential to leak when `s.processDelta` returns with an error.

Here is an example of how the goroutine leaks.
1. Two or more requests are made on the same stream (as in the case when using ADS)
2. The first request is [sent to `reqCh`](https://github.com/envoyproxy/go-control-plane/blob/9c62c968b253792310a10dda82789b43cb53e812/pkg/server/delta/v3/server.go#L245) and read from the channel
[here](https://github.com/envoyproxy/go-control-plane/blob/9c62c968b253792310a10dda82789b43cb53e812/pkg/server/delta/v3/server.go#L164).
3. The goroutine for-loop gets [past](https://github.com/envoyproxy/go-control-plane/blob/9c62c968b253792310a10dda82789b43cb53e812/pkg/server/delta/v3/server.go#L235) and is blocked on the request being sent on the (unbuffered) channel.
4. While processing the first request, an error is returned (e.g. from a callback) and the processDelta method is exited, stopping any processing of the request channel. For example you could return an error [here](https://github.com/envoyproxy/go-control-plane/blob/9c62c968b253792310a10dda82789b43cb53e812/pkg/server/delta/v3/server.go#L180).
5. The goroutine is still blocked on the channel sending. [here](https://github.com/envoyproxy/go-control-plane/blob/9c62c968b253792310a10dda82789b43cb53e812/pkg/server/delta/v3/server.go#L245). Because processDelta has returned, there will not be a read from `reqCh` and the goroutine ends up leaking.

A possible fix would be the following
```
for {
[...]

PR #916, opened 2024-04-08
Goroutine leak in DeltaStreamHandler: fix for issue 913:

zhiyanfoo:
This PR adds a unit test for the go routine leak identified in issue [913](https://github.com/envoyproxy/go-control-plane/issues/913). The bug fix is simply to re-use the same logic that already exists for the state-of-the-world StreamHandler.

You can re-create the goroutine leak by running the commit with just the test, and seeing that it fails before the fix.


valerian-roche (commented):
Thank you for the PR. The implementation looks good to me, but a small detail on the test implementation.

valerian-roche (approved):
Looks good for tests now, thanks for this fix

PR #949, opened 2024-05-29
xds server: Update reference/resource checks for DefaultFilterChain

sfc-gh-randerson:
This updates the consistency check to also consider the "DefaultFilterChain" in addition to the list of filter chains.

### DIFF
--- a/pkg/cache/v3/resource.go
+++ b/pkg/cache/v3/resource.go
@@ -202,22 +202,11 @@ func getListenerReferences(src *listener.Listener, out map[resource.Type]map[str
 
 	// Extract route configuration names from HTTP connection manager.
 	for _, chain := range src.GetFilterChains() {
-		for _, filter := range chain.GetFilters() {
-			config := resource.GetHTTPConnectionManager(filter)
-			if config == nil {
-				continue
-			}
+		getListenerReferencesFromChain(chain, routes)
+	}
 
-			// If we are using RDS, add the referenced the route name.
-			if name := config.GetRds().GetRouteConfigName(); name != "" {
-				routes[name] = true
-			}
-
-			// If the scoped route mapping is embedded, add the referenced route resource names.
-			for _, s := range config.GetScopedRoutes().GetScopedRouteConfigurationsList().GetScopedRouteConfigurations() {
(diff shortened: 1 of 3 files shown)


pkg/cache/v3/resource.go
```diff
@@ -202,22 +202,11 @@ func getListenerReferences(src *listener.Listener, out map[resource.Type]map[str
...
-			}
+		getListenerReferencesFromChain(chain, routes)
+	}
 
-			// If the scoped route mapping is embedded, add the referenced route resource names.
-			for _, s := range config.GetScopedRoutes().GetScopedRouteConfigurationsList().GetScopedRouteConfigurations() {
-				routes[s.GetRouteConfigurationName()] = true
-			}
-		}
+	if src.DefaultFilterChain != nil {
```
valerian-roche, line 208: Minor: the current code does not make an assumption on src not being `nil`, whereas this change would panic. Can you use the protobuf safe getter for consistency?

valerian-roche (commented):
Minor comment only
Tbh this code is not really maintained and very likely out of sync with envoy which has evolved on supporting partial snapshots. 
Can you confirm if you rely on this functionality? My long term goal is to fully remove this logic as it is imo out of place and is also creating issues to support other data-planes (for instance grpc libraries)

pkg/cache/v3/resource.go
sfc-gh-randerson: Ah, good point. Fixed.

sfc-gh-randerson:
> Minor comment only Tbh this code is not really maintained and very likely out of sync with envoy which has evolved on supporting partial snapshots. Can you confirm if you rely on this functionality? My long term goal is to fully remove this logic as it is imo out of place and is also creating issues to support other data-planes (for instance grpc libraries)

By "this code", do you mean the resource/reference check part of cache.Snapshot? Or something more broad?

We are setting ads=true when we call NewSnapshotCache() which seems to imply that our snapshots have to be consistent, which is what prompted this change. We can probably change away from that, but there's not really an indication right now that it's discouraged.

valerian-roche (approved):
lgtm

PR #964, opened 2024-07-03
chore: update go version to 1.21

mmorel-35:
I don't know what it the policy with golang version usage.
1.22 is out with several patches and 1.17 is not supported according to golang versioning policy


sunjayBhatia (dismissed):
I'm in favor

cc @envoyproxy/go-control-plane

valerian-roche (approved):
Thanks for this change! This should enable work on splitting the generated API in its own module to resume

phlax (approved):
landing on the basis of prior approvals

lgtm, thanks @mmorel-35

PR #993, opened 2024-08-29
fix: golangci-lint

mmorel-35:
- exclude G115


valerian-roche (approved):
Thanks for the fix, missed listing was broken

PR #1034, opened 2024-10-25
Fix envoy initial fetch time out when CDS updated

haorenfsa:
Fix https://github.com/envoyproxy/go-control-plane/issues/1035

related: https://github.com/envoyproxy/gateway/issues/4445

### DIFF
--- a/pkg/cache/v3/simple.go
+++ b/pkg/cache/v3/simple.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/envoyproxy/go-control-plane/pkg/cache/types"
 	"github.com/envoyproxy/go-control-plane/pkg/log"
+	"github.com/envoyproxy/go-control-plane/pkg/resource/v3"
 	"github.com/envoyproxy/go-control-plane/pkg/server/stream/v3"
 )
 
@@ -586,10 +587,18 @@ func (cache *snapshotCache) respondDelta(ctx context.Context, snapshot ResourceS
 		systemVersion: snapshot.GetVersion(request.GetTypeUrl()),
 	})
 
-	// Only send a response if there were changes
-	// We want to respond immediately for the first wildcard request in a stream, even if the response is empty
+	// Only send a response if:
+	// 1. there were changes
+	// 2. the first wildcard request in a stream
+	// 3. it's for EDS and there's a warming cluster
+	// in case 2 & 3 is resp is sent even if the response is empty
 	// otherwise, envoy won't complete initialization
-	if len(resp.Resources) > 0 || len(resp.RemovedResources) > 0 || (state.IsWildcard() && state.IsFirst()) {
+	switch {
+	case len(resp.Resources) > 0 || len(resp.RemovedResources) > 0:
+		fallthrough
+	case state.IsWildcard() && state.IsFirst():
+		fallthrough
+	case state.HasWarmingCluster() && request.TypeUrl == resource.EndpointType:
 		if cache.log != nil {
 			cache.log.Debugf("node: %s, sending delta response for typeURL %s with resources: %v removed resources: %v with wildcard: %t",
 				request.GetNode().GetId(), request.GetTypeUrl(), GetResourceNames(resp.Resources), resp.RemovedResources, state.IsWildcard())
--- a/pkg/server/delta/v3/server.go
+++ b/pkg/server/delta/v3/server.go
@@ -118,6 +118,17 @@ func (s *server) processDelta(str stream.DeltaStream, reqCh <-chan *discovery.De
 		watch := watches.deltaWatches[typ]
 		watch.nonce = nonce
 
+		if typ == resource.EndpointType {
+			watch.state.SetClusterWarming(false)
+		}
+		if typ == resource.ClusterType {
+			edsWatch, found := watches.deltaWatches[resource.EndpointType]
+			if found {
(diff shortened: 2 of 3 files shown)


haorenfsa:
WIP, I'm sure the version map is messed up some where, not sure where yet. I'll update when I finally fixed this.

haorenfsa:
@arkodg  Please help review, thank you in advance

lukidzi:
encountered the same issue. Here is my PR, but unfortunately, it hasn't received a review https://github.com/envoyproxy/go-control-plane/pull/981

pkg/cache/v3/simple.go
```diff
@@ -586,10 +587,16 @@ func (cache *snapshotCache) respondDelta(ctx context.Context, snapshot ResourceS
...
-	// Only send a response if there were changes
-	// We want to respond immediately for the first wildcard request in a stream, even if the response is empty
+	// Only send a response if:
+	// 1. there were changes
+	// 2. the first wildcard request in a stream
+	// 3. it's for EDS and there's a warming cluster
+	// in case 2 & 3 is resp is sent even if the response is empty
 	// otherwise, envoy won't complete initialization
-	if len(resp.Resources) > 0 || len(resp.RemovedResources) > 0 || (state.IsWildcard() && state.IsFirst()) {
+	if len(resp.Resources) > 0 ||
```
alecholmez, line 596: Does this make more sense as a `switch{}` now where we can fall through cases? Having this many `||` seems a bit unreadable

pkg/server/stream/v3/stream.go
```diff
@@ -137,3 +142,11 @@ func (s *StreamState) SetKnownResourceNamesAsList(url string, names []string) {
 	}
 	s.knownResourceNames[url] = m
 }
+
+func (s *StreamState) HasWarmingCluster() bool {
```
alecholmez, line 146: Should this be more generic given that envoy also supports listener warming? Maybe `HasResourceWarming()`?

pkg/server/delta/v3/server.go
```diff
@@ -118,6 +118,17 @@ func (s *server) processDelta(str stream.DeltaStream, reqCh <-chan *discovery.De
 		watch := watches.deltaWatches[typ]
 		watch.nonce = nonce
 
+		if typ == resource.EndpointType {
+			watch.state.SetClusterWarming(false)
+		}
+		if typ == resource.ClusterType {
```
alecholmez, line 124: What about the cases where we get a warming listener?

alecholmez (changes requested):

pkg/server/delta/v3/server.go
haorenfsa: Yes! When I wrote this patch, I only had dynamic Cluster, so only ClusterType was considered. If we're to merge this patch, I'll add the rest resources need warming.

haorenfsa: @alecholmez Thank you for taking time reviewing this.
I just read the envoy xDS protocol document again https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#resource-warming,  and found that only `Cluster` warming needs an additional `EDS` response. No RDS response is required for `Listener` warming. So I think we only need consider about the `ClusterType` in this patch, what do you think?

image

pkg/server/stream/v3/stream.go
haorenfsa: Seems only Cluster warming needs to be considered, see below comments.

pkg/cache/v3/simple.go
haorenfsa: updated

pkg/server/delta/v3/server.go
haorenfsa: Hi @lukidzi, I checked your PR. As I stated above, seems only `CDS` & `EDS` needs to be taking care of ?  Anything I missed, please help point it out.

lukidzi: Hi, if both LDS and RDS are in use, you may encounter an issue when a Listener enters the warming phase and depends on RDS data to proceed. My PR only addressed the issue for CDS/EDS, so it doesn’t currently resolve the LDS/RDS dependency during the warming phase.

pkg/server/stream/v3/stream.go
```diff
@@ -54,6 +58,7 @@ func NewStreamState(wildcard bool, initialResourceVersions map[string]string) St
 		first:                   true,
 		knownResourceNames:      map[string]map[string]struct{}{},
 		ordered:                 false, // Ordered comes from the first request since that's when we discover if they want ADS
+		hasWarmingCluster:       false,
```
lukidzi, line 61: As @valerian-roche suggested in my PR [envoyproxy/go-control-plane#981](https://github.com/envoyproxy/go-control-plane/pull/981), it would be beneficial to push only the endpoints of updated clusters. When your Envoy instance is watching over 1000 clusters, forcing a full update for all of them can be excessive if only one has changed. I could reopen my PR and finish it up if it makes sense to do it. Envoy has a new feature to load cached EDS but it works ONLY when `initial_fetch_timeout` is not disabled.

haorenfsa: It seems to me, only 1 extra empty EDS response is needed no matter how many Cluster is updated.

So my patch will only set an empty EDS response, if nothing changes.

PR #1038, opened 2024-11-05
chore(deps): upgrade golang version to 1.23.2

lukidzi:
Since golang 1.21 is out of life, I've upgraded Golang to 1.23.2

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ feedback, we might decide to revisit this aspect at a later point in time.
 
 ## Requirements
 
-1. Go 1.21+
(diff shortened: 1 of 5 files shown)


valerian-roche:
I don't think we can update to latest as a lot of components import this repository to source the envoy protobuf schemas in go (including grpc-go that's currently in go 1.22.x)

valerian-roche (approved):

lukidzi:
If everything is ok, please merge, I don't have permission. Thanks!

PR #1075, opened 2024-12-24
#1074: Add imports of previous root package in new subpackages to avoid dependency upgrade issues.

### LINKED ISSUE #1074: v0.13.2 broke dependency upgrades for cloud.google.com/go/storage
The change in https://github.com/envoyproxy/go-control-plane/pull/714 (released in v0.13.2) seems to have broken dependency upgrades for the google cloud Go packages (see https://github.com/googleapis/google-cloud-go/issues/11283#issuecomment-2560313410).

workaround:

```
go mod edit --exclude=github.com/envoyproxy/go-control-plane@v0.13.2
```

valerian-roche:
The module split seems to be breaking when the `get -u` command is used. 
Using the [link](https://go.dev/wiki/Modules#is-it-possible-to-add-a-module-to-a-multi-module-repository) referred in this [post](https://github.com/googleapis/google-cloud-go/issues/11283#issuecomment-2560350667), this change attempts to address the issue. 

While I was able to test this change within one of our repository depending on this repository (for go-control-plane and envoy API), I do not know of a way to guarantee that this will resolve all cases.

Fixes #1074


valerian-roche:
Also added a fix to `./ci/sync_envoy.sh` to avoid the workflow removing the submodule files

contrib/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/envoyproxy/go-control-plane/contrib
 
-go 1.21
+go 1.22.8
```
codyoss, line 3: nit: it would be great if the go.mod version can remain the same for this fix.

valerian-roche: This version is automatically updated due to the added dependency on the root-level go.mod which has been using 1.22.8 for some time.
If a blocker I can see to rollback this change but it is not, per se, part of the module split

envoy/go.mod
```diff
@@ -1,9 +1,13 @@
 module github.com/envoyproxy/go-control-plane/envoy
 
-go 1.21
+go 1.22.8
+
+// Used to resolve import issues related to go-control-plane package split (https://github.com/envoyproxy/go-control-plane/issues/1074)
+replace github.com/envoyproxy/go-control-plane@v0.13.2 => ../
```
Arkelenia, line 6: Is there a reason for pinning the version here but not in the go.mod files for `examples/dyplomat` and `ratelimit` packages?

contrib/go.mod
valerian-roche: In a separate PR we removed the dependency on the minor to now only depend on go 1.22, aligning with grpc-go required version.

envoy/go.mod
valerian-roche: Those packages inherit it through the envoy submodule, and therefore don't require it directly

envoy/go.mod
```diff
@@ -2,8 +2,12 @@ module github.com/envoyproxy/go-control-plane/envoy
 
 go 1.22
 
+// Used to resolve import issues related to go-control-plane package split (https://github.com/envoyproxy/go-control-plane/issues/1074)
+replace github.com/envoyproxy/go-control-plane@v0.13.3 => ../
```
codyoss, line 6: nit: unless you need this for your project replace statements don't do anything for downstream users. So if you don't need this it can be removed.

codyoss (commented):
left a nit otherwise LGTM. I'm a maintainer from the google-cloud-go project affected by this issue. Thanks for the fix.

envoy/go.mod
valerian-roche: Thank you for the review. I took this from https://go.dev/wiki/Modules#is-it-possible-to-add-a-module-to-a-multi-module-repository, where they explicitly state this replace as a way to avoid the "ambiguous" import.
I do not have enough understanding of go.mod resolution logic to know if this actually helps or not

codyoss: I believe the section you are referring to is:

```
cd path-to/github.com/my-repo/mig
go mod edit -replace github.com/my-repo@v1.3.0=../
go test ./...
go mod edit -dropreplace github.com/my-repo@v1.3.0
```

If so this is meant to be a temporary test, as you can see be the drop command after. This is because it assumes the module is not yet published that you are referencing. In this case though the version where this module has been carved out already exists. 

This causes no harm either way though, just thought I would share :)

valerian-roche: Thanks for the update. I think I get it better now, and they cut tag out of the intermediate commit with the replace. In order to keep it simpler to track I'll do this in two PRs. I'll create a new one to drop the replace

PR #1078, opened 2025-01-02
deps: Remove Go minor version from go.mods

### LINKED ISSUE #1077: Specifying Go minor version in go.mod forces users to upgrade their Go minor versions as well
In https://github.com/envoyproxy/go-control-plane/pull/1038. the Go major version was bumped to 1.22 AND the Go minor version was also specified. Quoting [this comment](https://github.com/grpc/grpc-go/pull/7624#discussion_r1792654660) from @ash2k:
> The issue with specifying the patch version that is not 0 is that now all modules that import this module will have to use this or a newer version. There may be reasons people don't or cannot use the more recent patch version. FWIW I think a library shouldn't have an opinion on the patch version used.
>
> An example of this - we use https://github.com/golang-fips/go/ to provide FIPS-compatible builds. Not all Go versions may be available there e.g. right now there is [no 1.23.2](https://github.com/golang-fips/go/tags). [Go 1.22.7](https://go.dev/doc/devel/release#go1.22.0) was released 2024-09-05 but FIPS version was [tagged](https://github.com/golang-fips/go/releases/tag/go1.22.7-1-openssl-fips) 2024-09-27. If grpc-go released a CVE fix in the window of those 22 days, we wouldn't have been able to upgrade.
>
> TL;DR this doesn't benefit grpc-go in any way but might hurt your users.

Due to user feedback, gRPC Go decided to drop the Go minor version in https://github.com/grpc/grpc-go/pull/7831. However using the latest version of go-control-plane is re-introducing the minor version constraint: https://github.com/grpc/grpc-go/pull/7974

Looking at https://github.com/envoyproxy/go-control-plane/pull/1038, it doesn't appear necessary to specify the minor version. I’d like to request that the minor version constraint be dropped.

mmorel-35:
Closes #1077

### DIFF
--- a/contrib/go.mod
+++ b/contrib/go.mod
@@ -1,6 +1,6 @@
 module github.com/envoyproxy/go-control-plane/contrib
 
-go 1.21
+go 1.22
 
 replace github.com/envoyproxy/go-control-plane/envoy => ../envoy
 
--- a/envoy/go.mod
+++ b/envoy/go.mod
@@ -1,6 +1,6 @@
 module github.com/envoyproxy/go-control-plane/envoy
 
-go 1.21
+go 1.22
 
 require (
 	github.com/cncf/xds/go v0.0.0-20240723142845-024c85f92f20
--- a/examples/dyplomat/go.mod
+++ b/examples/dyplomat/go.mod
@@ -1,6 +1,6 @@
 module github.com/envoyproxy/go-control-plane/examples/dyplomat
 
-go 1.22.8
+go 1.22
 
 replace (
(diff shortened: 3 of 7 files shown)


valerian-roche:
> @valerian-roche , With this and a new version for each module, grpc-go could be using the new modules. WDYT ?

I think we'd still need #1075 on top, but this would be the first piece. The main question is: should we go back to 21? This way the new version would not increase the go version compared to the previous release, and once the versioning issue has been solved we can take a look at updating again?

valerian-roche (approved):

PR #1231, opened 2025-06-11
Add `OTelAccessLog` to list of Access log sink names

davidjumani:
Adds the `OTelAccessLog` to the list of Access log sink names. This defines a constant for users to import who wish to use this sink rather than defining it in their implementation


davidjumani:
Closing this as I just realized that this is deprecated

PR #1298, opened 2025-10-01
[sotw][linear] Fix missing watch cleanup in linear cache for sotw watches subscribing to multiple resources

valerian-roche:
The logic to cleanup watches in the linear cache for non-wildcard requests is currently leaking watches when a watch is replied. This PR ensures watches are properly cleaned in this case.


valerian-roche:
Previous review of this commit available on a fork in https://github.com/DataDog/go-control-plane/pull/4

PR #1299, opened 2025-10-01
[sotw][issue-540] Return full state when applicable for watches in linear cache

valerian-roche:
The xds specification defines (in a very envoy-specific way), that even if state-of-the-world specifies only updated resources have to be returned when sending a response, in the specific context of CDS and LDS all resources must always be returned. 
This PR ensures this change is done, allowing the use of linear cache for LDS and CDS

Fixes: #540

### DIFF
--- a/pkg/cache/v3/linear.go
+++ b/pkg/cache/v3/linear.go
@@ -106,6 +106,7 @@ func NewLinearCache(typeURL string, opts ...LinearCacheOption) *LinearCache {
 		versionMap:    nil,
 		version:       0,
 		versionVector: make(map[string]uint64),
+		log:           log.NewDefaultLogger(),
 	}
 	for _, opt := range opts {
 		opt(out)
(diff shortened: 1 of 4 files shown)


valerian-roche:
This PR was originally reviewed as part of https://github.com/DataDog/go-control-plane/pull/5

pkg/cache/v3/linear.go
```diff
@@ -339,20 +357,26 @@ func (cache *LinearCache) CreateWatch(request *Request, _ stream.StreamState, va
...
+			cache.log.Debugf("Watch is stale with stale resources %v", staleResources)
+		}
 	}
 	if stale {
 		cache.respond(watch, staleResources)
 		return nil
 	}
 	// Create open watches since versions are up to date.
 	if len(request.GetResourceNames()) == 0 {
+		cache.log.Infof("[linear cache] open watch for %s all resources, system version %q", cache.typeURL, cache.getVersion())
```
Arkelenia: nit: `for all %s resources`

valerian-roche: Thanks, updated

PR #1300, opened 2025-10-01
[server][cache] Ensure subscription is properly updated for sotw and delta

valerian-roche:
Currently sotw and delta watches use different interfaces to track the stream state An assumption used to be that sotw watches are stateless and can be processed from the request itself. This was already altered in envoyproxy#508, but it only covers the usecases partially (e.g. it does not cover subscribing/unsubscribing to wildcard). The current object passed to track the subscription provides multiple issues:
 - it is not an interface, restricting evolution or testing for users building a separate cache
 - it tracks some objects per types, other globally, based on sotw or delta, making it confusing and hard to use in a consistent way

This commit replaces the legacy streamState by the notion of Subscription per type. StreamState is not relevant to the cache, as it only requires to know the subscription status (e.g. the last nonce is not needed), while a stream can multiplex multiple types. Further work will come to:
 - fill the subscription with the same attributes in sotw and delta, allowing to take proper decisions on what to reply
 - Fix multiple issues within the cache in sotw
 - Fix handling of explicit vs. legacy wildcard in sotw and delta

Now that the `Subscription` has been standardized to be uniform for `sotw` and `delta`, as well as always for the given request type, this PR ensures all the fields are properly populated in all cases A follow-up PR will rely on those informations to address current issues in the cache implementations related to resource tracking

Note: this PR has been running for a while in our infrastructure using both envoy and gRPC as clients. While this obviously does not guarantee fully functional code, this has significant production bake time


valerian-roche:
This PR includes the changes of fork PRs https://github.com/DataDog/go-control-plane/pull/8 and https://github.com/DataDog/go-control-plane/pull/9

pkg/server/stream/v3/subscription.go
```diff
@@ -0,0 +1,176 @@
...
+
+	if initialResourceVersions == nil {
+		state.returnedResources = make(map[string]string)
+	}
+
+	return state
+}
+
+func NewSotwSubscription(subscribed []string) Subscription {
+	sub := newSubscription(len(subscribed) == 0, nil)
```
Arkelenia, line 43: Isn't it possible for the initial set of resources to contain an explicit wildcard?

pkg/server/stream/v3/subscription.go
```diff
@@ -0,0 +1,176 @@
...
+		}
+	}
+
+	// Explicit subscription to wildcard as we are not in legacy wildcard behavior
+	s.wildcard = explicitWildcardSet
+	s.subscribedResourceNames = subscribedResources
+}
+
+func NewDeltaSubscription(subscribed, unsubscribed []string, initialResourceVersions map[string]string) Subscription {
+	sub := newSubscription(len(subscribed) == 0, initialResourceVersions)
```
Arkelenia, line 83: Similarly, isn't it possible to set an explicit wildcard in the first delta request?

pkg/server/stream/v3/subscription.go
```diff
@@ -0,0 +1,176 @@
...
+	return sub
+}
+
+// UpdateResourceSubscriptions updates the subscribed resources (including the wildcard state)
+// based on newly subscribed or unsubscribed resources
+// Used in delta subscriptions
+func (s *Subscription) UpdateResourceSubscriptions(subscribed, unsubscribed []string) {
+	// Handles legacy wildcard behavior first to exit if we are still in this behavior
+	if s.allowLegacyWildcard {
+		// The protocol (as of v1.29.0) only references subscribed as triggering
```
Arkelenia, line 94: nit: the version can be updated to v1.36.0

pkg/server/stream/v3/subscription.go
valerian-roche: It is, and it would be handled in `SetResourceSubscription` below. 
This code is only as legacy makes the behavior of "nothing" later not wildcard, hence the split behavior here

pkg/server/stream/v3/subscription.go
valerian-roche: The same way it will be handled below. `UpdateResourceSubscriptions` with empty usually does not set wildcard, hence the split handling

pkg/server/stream/v3/subscription.go
valerian-roche: I'll update in the next PR if possible, to avoid rebuilding/rebasing

Arkelenia (approved):
A few questions but overall, looks good to me 👍

PR #1301, opened 2025-10-01
[Sotw][Linear cache] Ensure watches are properly considering subscription changes to not miss new resources

valerian-roche:
This PR leverages the new `Subscription` interface passed into the `CreateWatch` method to properly track the resources returned as well as the evolution of the watch subscription. This allows a proper handling of resource re-subscription as well as semantics needed for the new wildcard mode

Fixes: #431

### DIFF
--- a/pkg/cache/v3/cache.go
+++ b/pkg/cache/v3/cache.go
@@ -118,6 +118,11 @@ type Response interface {
 	// Get the version in the Response.
 	GetVersion() (string, error)
 
+	// GetReturnedResources returns the map of resources and their versions returned in the subscription.
+	// It may include more resources than directly set in the response to consider the full state of the client.
+	// The caller is expected to provide this unchanged to the next call to CreateWatch as part of the subscription.
+	GetReturnedResources() map[string]string
+
 	// Get the context provided during response creation.
 	GetContext() context.Context
 }
@@ -154,6 +159,12 @@ type RawResponse struct {
 
 	// Resources to be included in the response.
 	Resources []types.ResourceWithTTL
+
+	// ReturnedResources tracks the resources returned for the subscription and the version when it was last returned,
+	// including previously returned ones when using non-full state resources.
+	// It allows the cache to know what the client knows. The server will transparently forward this
+	// across requests, and the cache is responsible for its interpretation.
+	ReturnedResources map[string]string
 
 	// Whether this is a heartbeat response. For xDS versions that support TTL, this
 	// will be converted into a response that doesn't contain the actual resource protobuf.
@@ -207,6 +218,12 @@ type PassthroughResponse struct {
 	DiscoveryResponse *discovery.DiscoveryResponse
 
 	ctx context.Context
+
+	// ReturnedResources tracks the resources returned for the subscription and the version when it was last returned,
+	// including previously returned ones when using non-full state resources.
+	// It allows the cache to know what the client knows. The server will transparently forward this
+	// across requests, and the cache is responsible for its interpretation.
+	ReturnedResources map[string]string
 }
 
 // DeltaPassthroughResponse is a pre constructed xDS response that need not go through marshaling transformations.
@@ -262,6 +279,10 @@ func (r *RawResponse) GetDiscoveryResponse() (*discovery.DiscoveryResponse, erro
 	}
 
 	return marshaledResponse.(*discovery.DiscoveryResponse), nil
+}
+
+func (r *RawResponse) GetReturnedResources() map[string]string {
+	return r.ReturnedResources
 }
 
 // GetDeltaDiscoveryResponse performs the marshaling the first time its called and uses the cached response subsequently.
@@ -365,6 +386,10 @@ func (r *RawResponse) maybeCreateTTLResource(resource types.ResourceWithTTL) (ty
 // GetDiscoveryResponse returns the final passthrough Discovery Response.
 func (r *PassthroughResponse) GetDiscoveryResponse() (*discovery.DiscoveryResponse, error) {
 	return r.DiscoveryResponse, nil
+}
+
+func (r *PassthroughResponse) GetReturnedResources() map[string]string {
+	return r.ReturnedResources
 }
 
 // GetDeltaDiscoveryResponse returns the final passthrough Delta Discovery Response.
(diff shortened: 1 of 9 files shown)


pkg/cache/v3/cache.go
```diff
@@ -367,6 +388,10 @@ func (r *PassthroughResponse) GetDiscoveryResponse() (*discovery.DiscoveryRespon
 	return r.DiscoveryResponse, nil
 }
 
+func (r *PassthroughResponse) GetReturnedResources() map[string]string {
```
Arkelenia, line 393: Out of curiosity, are passthrough responses even used anymore?

pkg/cache/v3/linear.go
```diff
@@ -100,74 +99,184 @@ func NewLinearCache(typeURL string, opts ...LinearCacheOption) *LinearCache {
...
+	// but this ensures we detect unsubscription then resubscription.
+	for _, resourceName := range removedResources {
+		delete(returnedVersions, resourceName)
+	}
+
+	if !ignoreReturnedResources && !ResourceRequiresFullStateInSotw(cache.typeURL) && len(resources) == 0 {
+		// If the request is not the initial one, and the type for not require full updates,
+		// do not return if noting is to be set
+		// For full-state resources an empty response does have a semantic meaning
+		return nil
```
Arkelenia, line 241: Doesn't this potentially break unsubscription then resubscription?

pkg/cache/v3/linear.go
```diff
@@ -100,74 +99,184 @@ func NewLinearCache(typeURL string, opts ...LinearCacheOption) *LinearCache {
...
+		for resourceName := range knownVersions {
+			// If the subscription no longer watches a resource,
+			// we mark it as unknown on the client side to ensure it will be resent to the client if subscribing again later on.
+			if _, ok := watch.subscription.SubscribedResources()[resourceName]; !ok {
+				removedResources = append(removedResources, resourceName)
+			}
+		}
+	}
+
+	if len(changedResources) == 0 && len(removedResources) == 0 && !ignoreReturnedResources {
```
Arkelenia, line 180: I'm not sure that we need to filter on `!ignoreReturnedResources`

pkg/cache/v3/linear.go
```diff
@@ -314,90 +425,86 @@ func (cache *LinearCache) GetResources() map[string]types.Resource {
...
-	var err error
-	if strings.HasPrefix(request.GetVersionInfo(), cache.versionPrefix) {
-		lastVersion, err = strconv.ParseUint(request.GetVersionInfo()[len(cache.versionPrefix):], 0, 64)
-	} else {
-		err = errors.New("mis-matched version prefix")
+
+	// If the request does not include a version the client considers it has no current state.
+	// In this case we will always reply to allow proper initialization of dependencies in the client.
+	ignoreCurrentSubscriptionResources := request.GetVersionInfo() == ""
+	if !strings.HasPrefix(request.GetVersionInfo(), cache.versionPrefix) {
```
Arkelenia, line 436: I'm wondering if this logic cannot be abstracted away in the `ResponseWatch`. I think the `ignoreCurrentSubscriptionResources` is functionally irrelevant. What seems to matter is the "real" set of returned resources.

pkg/cache/v3/simple.go
```diff
@@ -402,54 +402,58 @@ func (cache *snapshotCache) CreateWatch(request *Request, sub Subscription, valu
...
-			for _, name := range diff {
-				if _, exists := resources[name]; exists {
-					if err := cache.respond(context.Background(), request, value, resources, version, false); err != nil {
-						cache.log.Errorf("failed to send a response for %s%v to nodeID %q: %s", request.GetTypeUrl(),
-							request.GetResourceNames(), nodeID, err)
-						return nil, fmt.Errorf("failed to send the response: %w", err)
-					}
-					return func() {}, nil
+		} else {
+			// Check if a resource present in the snapshot is currently not returned,
```
Arkelenia, line 438: I think this is a good precautionary check but I'm curious whether we expect to ever find a diff here. Adding a resource to the snapshot should increment the version, shouldn't it?

pkg/cache/v3/cache.go
valerian-roche: Not in our repo, but those are part of the public API so I'd rather not remove those

pkg/cache/v3/linear.go
valerian-roche: Hmm on resub the `resources` should not be empty. Not breaking here would still send a fully empty request, and not provide the resource

pkg/cache/v3/linear.go
valerian-roche: What do you mean? If we are in the initial response and it's lds/cds we need to reply empty if the client has lost context (which triggers `ignoreReturnedResources`)

pkg/cache/v3/linear.go
valerian-roche: I'm trying to keep the `ResponseWatch` specific to the request, without stream context, to properly split from `Subscription`, which does have context and is read-only within the context of the cache
I'm unsure what you mean by `ignoreCurrentSubscriptionResources` being irrelevant. While for full state "empty" always has the same meaning it is not true for non-full state, which is a stateful protocol

pkg/cache/v3/simple.go
valerian-roche: Yes it can happen in the case when a subscription switched from non-wildcard to wildcard. The snapshot version would not have changed but the subscribed resources "increased" silently (i.e. not in the subscribed resources). While I do not know of existing use-cases, it is clearly documented as a valid case in the xds protocol

pkg/cache/v3/linear.go
Arkelenia: Right, but it also doesn't update `returnedResources` which could prevent us from resubscribing later, no?

Arkelenia: After discussing offline, this isn't an issue since this is updated in the subscription when the subscribed resources change so that `returnedResources` is always a subset of `subscribed Resources`

pkg/cache/v3/linear.go
Arkelenia: This one is necessary because Envoy expects an explicit empty response on a first CDS or LDS request for example.

PR #1319, opened 2025-10-17
[Simple] Fix request/response loop if a non-wildcard request points to a resource not currently in the cache

valerian-roche:
In #1301 a change to properly handle unsubscribing/resubscribing of resources mistakenly removed the check to ensure we don't reply if an explicitly subscribed resource is present in the request but not in the snapshot.
As a consequence we currently reply with the same resources and version, which triggers a loop between the control-plane and the client. 

This PR fixes the issue. The issue still raises though:
- there is no protection against such loop. We should likely check that if we reply with the same version and the same resources, we should not reply, but I need to confirm there are no cases where it is actually expected
- I noticed that if we trigger the ads superset check, we do not reply but also do not create a watch (at this [line](https://github.com/envoyproxy/go-control-plane/blob/main/pkg/cache/v3/simple.go#L490) we silently return). This should likely be addressed

### DIFF
--- a/pkg/cache/v3/simple.go
+++ b/pkg/cache/v3/simple.go
@@ -17,6 +17,8 @@ package cache
 import (
 	"context"
 	"fmt"
+	"maps"
+	"slices"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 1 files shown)


Arkelenia (approved):
LGTM 👍

PR #1320, opened 2025-10-19
[Linear cache] rework watches and storage in the cache to make code more common between delta and sotw

valerian-roche:
Following previous commits on linear cache fixes in sotw, the implementation of linear cache has become more common between sotw and delta watches. This commit makes it clearer and reorganizes the setup:
 - reorganizes resource cache and watch tracking per resource, instead of per mode. This will enable changes like sotw using resource versions
 - avoids multiple marshaling of the resource when the version is already computed through marshaling (not in this commit, as potentially impactful on memory)
 - sotw and delta watches use a common model for their handling, with common code to compute subscription impact vs. cache. Only the generation of the response is different

### DIFF
--- a/pkg/cache/v3/linear.go
+++ b/pkg/cache/v3/linear.go
@@ -18,14 +18,74 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"maps"
 	"strconv"
 	"strings"
 	"sync"
-	"sync/atomic"
 
 	"github.com/envoyproxy/go-control-plane/pkg/cache/types"
 	"github.com/envoyproxy/go-control-plane/pkg/log"
 )
+
+// cachedResource is used to track resources added by the user in the cache.
+// It contains the resource itself and its associated version (currently in two different modes).
+type cachedResource struct {
+	types.Resource
+
+	// cacheVersion is the version of the cache at the time of last update, used in sotw.
+	cacheVersion string
+	// stableVersion is the version of the resource itself (a hash of its content after deterministic marshaling).
+	// It is lazy initialized and should be accessed through getStableVersion.
+	stableVersion string
+}
+
+func newCachedResource(res types.Resource, cacheVersion string) *cachedResource {
+	return &cachedResource{
+		Resource:     res,
+		cacheVersion: cacheVersion,
+	}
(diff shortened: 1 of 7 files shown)


valerian-roche:
This PR backports commits from https://github.com/DataDog/go-control-plane/pull/11

pkg/cache/v3/linear.go
```diff
@@ -97,58 +155,72 @@ func WithLogger(log log.Logger) LinearCacheOption {
...
+		log:             log.NewDefaultLogger(),
 	}
 	for _, opt := range opts {
 		opt(out)
 	}
-	for name := range out.resources {
-		out.versionVector[name] = out.getVersion()
+	for name, resource := range out.resources {
+		resource.cacheVersion = out.getVersion()
+		out.resources[name] = resource
```
Arkelenia, line 171: nit: This should be unnecessary since `resource` is a pointer.

pkg/cache/v3/linear.go
```diff
@@ -360,55 +484,48 @@ func (cache *LinearCache) UpdateResources(toUpdate map[string]types.Resource, to
...
-
-	return nil
+	return cache.notifyAll(modified)
 }
 
 // SetResources replaces current resources with a new set of resources.
-// This function is useful for wildcard xDS subscriptions.
-// This way watches that are subscribed to all resources are triggered only once regardless of how many resources are changed.
+// If only some resources are to be updated, UpdateResources is more efficient.
 func (cache *LinearCache) SetResources(resources map[string]types.Resource) {
```
Arkelenia, line 503: Not something to necessarily change in this PR but it is weird that this specific method doesn't return an error.

Arkelenia (approved):
LGTM 👍

pkg/cache/v3/linear.go
valerian-roche: Good remark, will cleanup in next PR

pkg/cache/v3/linear.go
valerian-roche: Yes sadly that's part of the main API of the cache, so I don't want to touch it

PR #1322, opened 2025-10-21
[Linear cache] Increase mutualization between delta and sotw in linear cache to have behavior driven logic instead

valerian-roche:
Following previous rework of sotw and delta handling, the remaining differences of behavior between the two are:
 - on first request, check the returned version in sotw to validate version prefix. Also compute the version differently
 - in some cases, return full state in sotw. In other cases don't return if the only change is deletion
 - compute the version differently

In this context, the code can now be mostly merged, apart from the type differences on requests and responses (which I think can be reduced to none soon, but requires more rework on simple side and would be somewhat backward incompatible). In this commit, the linear cache no longer treats sotw and delta watches differently. They are tracked in the same maps and the changes and responses are computed in a common method. Differences in behavior (e.g. full state or which version to use) is now a functional attribute of the watch itself. There are two main parts not yet fully merged:
 - changing the version in sotw when using stable versions. The requirement to prepend cache version prefix makes it harder to abstract
 - defining whether a change only removing resources should trigger a response. This could be made a functional flag but given its intersection with `fullStateResponses` seemed fragile

### DIFF
--- a/pkg/cache/v3/cache.go
+++ b/pkg/cache/v3/cache.go
@@ -105,41 +105,67 @@ type Cache interface {
 
 // Response is a wrapper around Envoy's DiscoveryResponse.
 type Response interface {
-	// Get the Constructed DiscoveryResponse
+	// GetDiscoveryResponse returns the Constructed DiscoveryResponse.
 	GetDiscoveryResponse() (*discovery.DiscoveryResponse, error)
 
-	// Get the original Request for the Response.
+	// GetRequest returns the request that created the watch that we're now responding to.
+	// This is provided to allow the caller to correlate the response with a request.
+	// Generally this will be the latest request seen on the stream for the specific type.
 	GetRequest() *discovery.DiscoveryRequest
 
-	// Get the version in the Response.
+	// GetVersion returns the version in the Response.
+	// The version can be a property of the resources, allowing for optimizations in subsequent calls,
+	// or simply an internal property of the cache which can be used for debugging.
+	// The cache implementation should be able to determine if it can provide such optimization.
+	// Deprecated: use GetResponseVersion instead
 	GetVersion() (string, error)
+
+	// GetResponseVersion returns the version in the Response.
+	// The version can be a property of the resources, allowing for optimizations in subsequent calls,
+	// or simply an internal property of the cache which can be used for debugging.
+	// The cache implementation should be able to determine if it can provide such optimization.
+	GetResponseVersion() string
 
 	// GetReturnedResources returns the map of resources and their versions returned in the subscription.
 	// It may include more resources than directly set in the response to consider the full state of the client.
 	// The caller is expected to provide this unchanged to the next call to CreateWatch as part of the subscription.
 	GetReturnedResources() map[string]string
 
-	// Get the context provided during response creation.
+	// GetContext returns the context provided during response creation.
 	GetContext() context.Context
 }
 
-// DeltaResponse is a wrapper around Envoy's DeltaDiscoveryResponse
+// DeltaResponse is a wrapper around Envoy's DeltaDiscoveryResponse.
 type DeltaResponse interface {
-	// Get the constructed DeltaDiscoveryResponse
+	// GetDeltaDiscoveryResponse returns the constructed DeltaDiscoveryResponse.
 	GetDeltaDiscoveryResponse() (*discovery.DeltaDiscoveryResponse, error)
 
-	// Get the request that created the watch that we're now responding to. This is provided to allow the caller to correlate the
-	// response with a request. Generally this will be the latest request seen on the stream for the specific type.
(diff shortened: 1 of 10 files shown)


valerian-roche:
This PR backports the change already merged as part of https://github.com/DataDog/go-control-plane/pull/14

pkg/cache/v3/cache.go
```diff
@@ -184,12 +210,12 @@ type RawDeltaResponse struct {
...
 
 	// Resources to be included in the response.
-	Resources []types.Resource
+	Resources []types.ResourceWithTTL
 
 	// RemovedResources is a list of resource aliases which should be dropped by the consuming client.
 	RemovedResources []string
 
-	// NextVersionMap consists of updated version mappings after this response is applied
+	// NextVersionMap consists of updated version mappings after this response is applied.
```
Arkelenia, line 218: Should this be renamed to `ReturnedResources` for consistency?

pkg/cache/v3/linear.go
```diff
@@ -277,8 +279,9 @@ func (cache *LinearCache) computeSotwResponse(watch ResponseWatch, alwaysConside
 	switch {
 	// For lds and cds, answers will always include all existing subscribed resources, with no regard to which resource was changed or removed.
 	// For other types, the response only includes updated resources (sotw cannot notify for deletion).
-	case !ResourceRequiresFullStateInSotw(cache.typeURL):
-		if !alwaysConsiderAllResources && len(changedResources) == 0 {
+	case !watch.sendFullStateResponses():
+		// TODO(valerian-roche): remove this leak of delta/sotw behavior here.
+		if !watch.isDelta() && !replyEvenIfEmpty && len(changedResources) == 0 {
```
Arkelenia, line 284: I'm not sure to understand why the delta distinction is important here. If `replyEvenIfEmpty` is false and there are no changed resources, it should still be fine to send no response even for delta.

pkg/cache/v3/linear.go
```diff
@@ -166,31 +175,23 @@ func NewLinearCache(typeURL string, opts ...LinearCacheOption) *LinearCache {
...
 // The useStableVersion argument defines what version type to use for resources:
 //   - if set to false versions are based on when resources were updated in the cache.
 //   - if set to true versions are a stable property of the resource, with no regard to when it was added to the cache.
-func (cache *LinearCache) computeResourceChange(sub Subscription, alwaysConsiderAllResources, useStableVersion bool) (updated, removed []string, err error) {
+func (cache *LinearCache) computeResourceChange(sub Subscription, useStableVersion bool) (updated, removed []string, err error) {
 	var changedResources []string
 	var removedResources []string
 
 	knownVersions := sub.ReturnedResources()
-	if alwaysConsiderAllResources {
```
Arkelenia, line 190: I'm curious about this. Is this safe because `alwaysConsiderAllResources` is only true when the returned resources map in the subscription is already empty?

pkg/cache/v3/linear.go
```diff
@@ -634,71 +524,92 @@ func (cache *LinearCache) CreateDeltaWatch(request *DeltaRequest, sub Subscripti
...
+
 	watchID := cache.nextWatchID()
-	// Create open watches since versions are up to date.
+	sub := watch.getSubscription()
+
 	if sub.IsWildcard() {
-		cache.log.Infof("[linear cache] open delta watch %d for all %s resources, system version %q", watchID, cache.typeURL, cache.getVersion())
-		cache.wildcardWatches.delta[watchID] = watch
+		cache.log.Infof("[linear cache] open watch %d for %s all resources", watchID, cache.typeURL)
+		cache.log.Debugf("[linear cache] subscription details for watch %d: known versions %v, system version %q", watchID, sub.ReturnedResources(), cache.getVersion())
```
Arkelenia, line 565: would it be worth adding whether the watch was delta or not here? I know the goal is to completely abstract this away from the caches but I think it might help debug potential bugs in the short term considering the amount of changes that have been recently introduced.

pkg/cache/v3/cache.go
valerian-roche: This is public API here sadly, so I did not want to touch it. I added new methods to normalize, but did not remove the current ones

pkg/cache/v3/linear.go
valerian-roche: Yes exactly. The idea is that the subscription should be erased in such case

pkg/cache/v3/linear.go
valerian-roche: It's okay but an unneeded response. It also creates a risk of spin when returning responses with nothing in them

pkg/cache/v3/linear.go
valerian-roche: My bad missed this one. Will add in next PR

PR #1323, opened 2025-10-23
options to deactivate legacy wildcard

zhiyanfoo:
This PR add two new public facing xds options `DeactivateLegacyWildcard` and `DeactivateLegacyWildcardForTypes`.

This will allow users to disable legacy wildcard mode. This allows us to avoid returning all resources when envoy makes a VHDS request with an empty subscription list. It'll also allows us to ensure that we never return all resources to grpc-xds clients.

### DIFF
--- a/pkg/server/config/config.go
+++ b/pkg/server/config/config.go
@@ -49,22 +49,12 @@ func (o Opts) IsLegacyWildcardActive(typeURL string) bool {
 // this allows for easy inference as to which opt applies to what implementation.
 type XDSOption func(*Opts)
 
-// DeactivateLegacyWildcard deactivates legacy wildcard mode for all resource types.
-// In legacy wildcard mode, empty requests to a stream, are treated as wildcard requests as long
-// as there is no request made with resources or explicit wildcard requests on the same stream.
-// When deactivated, empty requests are treated as a request with no subscriptions to any resource.
-// This is recommended for when you are using the go-control-plane to serve grpc-xds clients.
-// These clients never want to treat an empty request as a wildcard subscription.
 func DeactivateLegacyWildcard() XDSOption {
 	return func(o *Opts) {
 		o.legacyWildcardDeactivated = true
 	}
 }
 
-// DeactivateLegacyWildcardForTypes deactivates legacy wildcard mode for specific resource types.
-// In legacy wildcard mode, empty requests to a stream, are treated as wildcard requests as long
-// as there is no request made with resources or explicit wildcard requests on the same stream.
-// When deactivated, empty requests are treated as a request with no subscriptions to any resource.
 func DeactivateLegacyWildcardForTypes(types []string) XDSOption {
 	return func(o *Opts) {
 		typeMap := make(map[string]struct{}, len(types))
--- a/pkg/server/delta/v3/server.go
+++ b/pkg/server/delta/v3/server.go
@@ -57,6 +57,24 @@ func WithLogger(logger log.Logger) config.XDSOption {
 	}
 }
 
+// DeactivateLegacyWildcard deactivates legacy wildcard mode for all resource types.
+// In legacy wildcard mode, empty requests to a stream, are treated as wildcard requests as long
+// as there is no request made with resources or explicit wildcard requests on the same stream.
+// When deactivated, empty requests are treated as a request with no subscriptions to any resource.
+// This is recommended for when you are using the go-control-plane to serve grpc-xds clients.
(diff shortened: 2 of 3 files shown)


pkg/server/stream/v3/subscription_test.go
```diff
@@ -157,3 +159,78 @@ func TestDeltaSubscriptions(t *testing.T) {
...
+		assert.False(t, sub.IsWildcard())
+
+		// Can still explicitly subscribe to wildcard
+		sub.SetResourceSubscription([]string{"*"})
+		assert.True(t, sub.IsWildcard())
+	})
+}
+
+func TestSotwSubscriptionsWithDeactivatedLegacyWildcardForTypes(t *testing.T) {
+	t.Run("deactivate for specific type", func(t *testing.T) {
```
tnawathe21: Can you also add explicitly subscribing to wildcard in this test?

zhiyanfoo: I removed this test because it's already covered by  the logic of the multiple types test. I added explict wildcard test to all of them, and added delta versions of the test.

pkg/server/stream/v3/subscription.go
```diff
@@ -24,10 +26,27 @@ type Subscription struct {
 }
 
 // newSubscription initializes a subscription state.
-func newSubscription(wildcard bool, initialResourceVersions map[string]string) Subscription {
+func newSubscription(emptyRequest bool, initialResourceVersions map[string]string, opts config.Opts, typeURL string) Subscription {
```
valerian-roche: I don't think passing Opts here is needed. Directly passing if legacy is deactivated is sufficient

pkg/server/config/config.go
```diff
@@ -18,6 +24,16 @@ func NewOpts() Opts {
 	}
 }
 
+// LegacyWildcardDeactivated returns whether legacy wildcard mode is deactivated for all resource types
+func (o Opts) LegacyWildcardDeactivated() bool {
```
valerian-roche: I would remove those two methods and replace them with `IsLegacyWildcardActive(typeURL)`
This would allow extending/evolving the option interface without impacting code in the rest of the modules

pkg/server/stream/v3/subscription.go
```diff
@@ -9,10 +11,10 @@ type Subscription struct {
 	// wildcard indicates if the subscription currently has a wildcard watch.
 	wildcard bool
 
-	// allowLegacyWildcard indicates that the stream never provided any resource
-	// and is de facto wildcard.
-	// As soon as a resource or an explicit subscription to wildcard is provided,
-	// this flag will be set to false
+	// allowLegacyWildcard indicates that an empty request should be treated as a wildcard request.
+	// If in the stream at some point subscribes explicitly a resource or a explicitly makes wildcard
```
valerian-roche: Comment to rewrite. The sentence is inconsistent

pkg/server/stream/v3/subscription.go
```diff
@@ -39,8 +48,8 @@ func newSubscription(wildcard bool, initialResourceVersions map[string]string) S
 	return state
 }
 
-func NewSotwSubscription(subscribed []string) Subscription {
-	sub := newSubscription(len(subscribed) == 0, nil)
+func NewSotwSubscription(subscribed []string, opts config.Opts, typeURL string) Subscription {
```
valerian-roche: Why is this passed in this interface? This is still leaking parameters here that have no value on the subscription

pkg/server/config/config.go
```diff
@@ -28,3 +48,29 @@ func NewOpts() Opts {
...
 // this allows for easy inference as to which opt applies to what implementation.
 type XDSOption func(*Opts)
+
+// DeactivateLegacyWildcard deactivates legacy wildcard mode for all resource types.
+// In legacy wildcard mode, empty requests to a stream, are treated as wildcard requests as long
+// as there is no request made with resources or explicit wildcard requests on the same stream.
+// When deactivated, empty requests are treated as a request with no subscriptions to any resource.
+// This is recommended for when you are using the go-control-plane to serve grpc-xds clients.
+// These clients never want to treat an empty request as a wildcard subscription.
+func DeactivateLegacyWildcard() XDSOption {
```
valerian-roche, line 52: Minor: can you map them in `sotw` and `delta` to allow cleaner import as reported at line 46

zhiyanfoo: done

PR #1324, opened 2025-10-23
[Linear] Ensure a resource is only serialized/hashed at most once to reply it

valerian-roche:
Currently the go-control-plane caches (both linear and snapshots) will serialize the resource as many times as there are clients receiving it. This is an issue with control-planes watched by a lot of clients, especially with large resources (e.g. endpoints)

This commit ensures that the serialization occurs at most once per resource, in all cases (sotw/delta watches and linear/snapshot cache). A resource will still only be serialized if:
  - it is returned to at least one client.
  - its version had to be considered to be returned (i.e. the resource was added again with the same stable version).

This does not yet remove the use of the version map in snapshot mode.
In linear mode a major gain already occurred in https://github.com/envoyproxy/go-control-plane/pull/1320 removing the hashing of resources never considered to be returned (which could be significant for endpoints depending on how the cache is managed).

### DIFF
--- a/pkg/cache/v3/cache.go
+++ b/pkg/cache/v3/cache.go
@@ -18,6 +18,7 @@ package cache
 import (
 	"context"
 	"errors"
+	"fmt"
 	"sync/atomic"
 
 	"google.golang.org/protobuf/proto"
@@ -179,14 +180,14 @@ type RawResponse struct {
 	// Proxy responds with this version as an acknowledgement.
 	Version string
 
-	// Resources to be included in the response.
-	Resources []types.ResourceWithTTL
+	// resources to be included in the response.
+	resources []*cachedResource
 
-	// ReturnedResources tracks the resources returned for the subscription and the version when it was last returned,
+	// returnedResources tracks the resources returned for the subscription and the version when it was last returned,
 	// including previously returned ones when using non-full state resources.
 	// It allows the cache to know what the client knows. The server will transparently forward this
 	// across requests, and the cache is responsible for its interpretation.
-	ReturnedResources map[string]string
+	returnedResources map[string]string
 
 	// Whether this is a heartbeat response. For xDS versions that support TTL, this
 	// will be converted into a response that doesn't contain the actual resource protobuf.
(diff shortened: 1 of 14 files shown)


pkg/cache/v3/cached_resource.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/envoyproxy/go-control-plane/pkg/cache/types"
+)
+
+// cachedResource is used to track resources added by the user in the cache.
+// It contains the resource itself and its associated version (currently in two different modes).
+type cachedResource struct {
+	name string
+
+	// resource must not be modified once the cachedResource is created.
+	resource types.Resource
```
Arkelenia, line 17: Do we even need to keep it as an attribute? In the end, this is already encapsulated in `marshalFunc`. Do we ever access it directly?

Arkelenia: Nevermind, it's used in the linear cache to provide `GetResources`

pkg/cache/v3/cache.go
```diff
@@ -198,7 +199,7 @@ type RawResponse struct {
 	Ctx context.Context
 
 	// marshaledResponse holds an atomic reference to the serialized discovery response.
-	marshaledResponse atomic.Value
+	marshaledResponse atomic.Pointer[discovery.DiscoveryResponse]
```
Arkelenia, line 202: I'm wondering if it wouldn't be worth using the same pattern as the `cachedResource` with `sync.onceValues`. I'm not sure whether `GetDiscoveryResponse` is called concurrently on the same `RawResponse` or not. If it is, then we may be marshaling multiple times the same response.

Arkelenia (approved):
LGTM 👍

pkg/cache/v3/cached_resource.go
valerian-roche: Yeah `GetResources` is a plague for evolution. But a major part of the current API I expect users to rely on

pkg/cache/v3/cache.go
valerian-roche: I was wondering. I'm also not sure it's even worth it now that resources will be marshaled at most once. I did not want to overcomplexify the change for it now, but something for later potentially

valerian-roche:
This PR was originally implemented as https://github.com/DataDog/go-control-plane/pull/21

PR #1325, opened 2025-10-24
Add server fuzzer

AdamKorcz:
This PR adds a fuzz test for the servers `ServeHTTP` method. 

The fuzzer controls several of the responses to also test the servers handling of odd responses. 

The fuzzer creates a random http request and passes it to the servers `ServeHTTP` method. It does not check the error from the server and therefore primarily checks for bugs that will crash the server.

The fuzzer then proceeds to do the same for the Delta servers `DeltaStreamHandler`.

### DIFF
--- a/pkg/server/v3/fuzz_test.go
+++ b/pkg/server/v3/fuzz_test.go
@@ -1,0 +1,249 @@
+// Copyright 2025 Envoyproxy Authors
+//
+//   Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


valerian-roche:
Hey, thank you for your contribution

The Fetch API on the rest server is a rarely used piece of code of `go-control-plane` (as envoy and grpc usually use the sotw or delta gRPC APIs). Can you provide the motivation to add this on the REST API instead?

PR #1326, opened 2025-10-27
Implements `Fetch` in `MuxCache`

davinci26:
`MuxCache` doesn't implement the `Fetch` method and it returns an error. However `MuxCache` doesn't have any logic instead it delegates to the appropriate cache based on the request. This means that we can just use the same logic as `CreateWatch` to delegate to the appropriate cache's `Fetch` method.

`Linear` cache doesnt implement `Fetch` either so this part of the call will fail but the rest of the caches will work as expected.

Additionally it was a bit confusing to debug the issue since the error message resembles quite a bit a `gRPC Unimplemented` error. So I updated the error message in `Linear` cache to be more specific.

### DIFF
--- a/pkg/cache/v3/linear.go
+++ b/pkg/cache/v3/linear.go
@@ -570,7 +570,7 @@ func (cache *LinearCache) getVersion() string {
 }
 
 func (cache *LinearCache) Fetch(context.Context, *Request) (Response, error) {
-	return nil, errors.New("Fetch is not implemented by LinearCache")
+	return nil, errors.New("fetch is not implemented by LinearCache")
 }
 
 // NumResources returns the number of resources currently in the cache.


davinci26:
@valerian-roche I see you are pretty active in the repo so I wonder if I can pester for you a code review 🙏

pkg/cache/v3/linear.go
```diff
@@ -570,7 +570,7 @@ func (cache *LinearCache) getVersion() string {
 }
 
 func (cache *LinearCache) Fetch(context.Context, *Request) (Response, error) {
-	return nil, errors.New("not implemented")
+	return nil, errors.New("Fetch is not implemented by LinearCache")
```
valerian-roche: Minor: most linters will trigger on capitalized error messages. Not sure if this goes through here as matching the method name though

valerian-roche (dismissed):

pkg/cache/v3/linear.go
davinci26: yeah not sure, how this slipped. My IDE also missed it

PR #1329, opened 2025-10-30
Add MCP filter name to wellknown

chuangw6:
Signed-off-by: Chuang Wang <chuangw@google.com>

### DIFF
--- a/pkg/wellknown/wellknown.go
+++ b/pkg/wellknown/wellknown.go
@@ -58,6 +58,8 @@ const (
 	HTTPGRPCStats = "envoy.filters.http.grpc_stats"
 	// HTTP WASM filter
 	HTTPWasm = "envoy.extensions.filters.http.wasm.v3.Wasm"
+	// MCP HTTP filter
+	MCP = "envoy.filters.http.mcp"
 )
 
 // Network filter names


botengyao:
Need to fix DCO.

chuangw6:
> DCO

~what's DCO?~ Oh I see. Done

botengyao (approved):

valerian-roche:
`wellknown` is [deprecated](https://github.com/envoyproxy/go-control-plane/pull/1329/files#diff-8f3f7baa2dfb1aa337ecf3b7a43b95a2c62e31a16ad754e7812a0573917413d7R17). It does require manual tracking of the very large number of typed extensions in envoy, and has the bad side effect of creating dependency on the main `go-control-plane` module instead of the `envoy` submodule. 
If users believe this is really valuable this should be moved to a new pkg within the `envoy` submodule and ideally be autogenerated from the protobufs

kyessenov:
+1, we don't recommend using filter names in xDS. Envoy just looks up by the type_url.

chuangw6:
aha I didn't notice the deprecation comment. Cool, so that means, `name` field doesn't matter (even if it's empty) as long as `@type` field is correct, right?

```yaml
            - name: envoy.filters.http.rbac
              typed_config:
                "@type": type.googleapis.com/envoy.extensions.filters.http.rbac.v3.RBAC
```

valerian-roche:
> aha I didn't notice the deprecation comment. Cool, so that means, `name` field doesn't matter (even if it's empty) as long as `@type` field is correct, right?
> 
> ```yaml
>             - name: envoy.filters.http.rbac
>               typed_config:
>                 "@type": type.googleapis.com/envoy.extensions.filters.http.rbac.v3.RBAC
> ```

The name field is used in some context iirc (e.g. if using route level override), but it does not need to be the filter type

PR #1356, opened 2025-12-07
[Snapshot Cache] Simplify Snapshot cache responses.

valerian-roche:
The snapshot cache reply logic for sotw and delta watches has grown complex and very distinct from the linear cache. This has created issues recently, mostly related to the logic in `CreateWatch` having to perfectly match the one in `respond` to avoid major issues. This commit simplifies the code in the same model as the linear cache: a single method creates the response, and the watch is created if no response is returned.

There is some behavior change: if the type is not Listeners or Clusters, the control-plane will now only return the modified resources, per the xDS protocol. Please create an issue if this creates problems.

Further work will build on this to:
 - support a new snapshot model, allowing per type updates and versions, as well as better delta handling
 - use per-resource version even in sotw to avoid resending resources when not needed.
 - provide support for partial wildcard responses, to provide a new support model for wildcard + resources subscriptions (e.g. for OdCDS).


pkg/cache/v3/simple.go
```diff
@@ -482,59 +485,164 @@ func (cache *snapshotCache) cancelWatch(nodeID string, watchID int64) func() {
...
+		if missing := difference(resources, watch.subscription.SubscribedResources()); len(missing) > 0 {
+			return nil, missingRequestResource{missing}
 		}
 	}
 
-	resp := createResponse(ctx, request, resources, version, heartbeat)
-	cache.log.Debugf("respond %s (requested %v) version %q with version %q and resources %v", request.GetTypeUrl(), request.GetResourceNames(), request.GetVersionInfo(), version, slices.Collect(maps.Keys(resp.GetReturnedResources())))
+	// This implementation can seem more complex than needed, as it does not blindly rely on the request version.
+	// This allows for a more generic implemenentation when considering wildcard + subscribed, or partial replies.
+	// In other context a lot could be simplified as
```
tnawathe21: Is this comment missing something?

pkg/cache/v3/snapshot.go
```diff
@@ -111,8 +111,8 @@ func (s *Snapshot) Consistent() error {
 			}
 
 			// Check superset.
```
tnawathe21:
```suggestion
			// Check difference.
```

pkg/cache/v3/simple.go
valerian-roche: Good catch, updated

pkg/cache/v3/snapshot.go
valerian-roche: Updated

PR #1467, opened 2026-05-11
[cache/linear] Expose cache version

nezdolik:
# Description
Expose version in linear cache as public getter. 

#<issue>

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How has this been tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* Go Toolchain:

# Checklist:

- [ ] I have made corresponding changes to the changelog
- [x] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published in downstream modules
- [x] I've added sufficient test coverage for my change. If not, please explain why.

### DIFF
--- a/pkg/cache/v3/linear.go
+++ b/pkg/cache/v3/linear.go
@@ -565,6 +565,10 @@ func (cache *LinearCache) CreateDeltaWatch(request *DeltaRequest, sub Subscripti
 	return cache.trackWatch(watch), nil
 }
 
+func (cache *LinearCache) GetVersion() uint64 {
+	return cache.version
+}
+
 func (cache *LinearCache) nextWatchID() uint64 {
 	cache.currentWatchID++
 	if cache.currentWatchID == 0 {
--- a/pkg/cache/v3/linear_test.go
+++ b/pkg/cache/v3/linear_test.go
@@ -394,6 +394,26 @@ func TestLinearGetResources(t *testing.T) {
 	assert.Truef(t, reflect.DeepEqual(expectedResources, resources), "resources are not equal. got: %v want: %v", resources, expectedResources)
 }
 
+func TestLinearGetVersion(t *testing.T) {
+	c := NewLinearCache(testType)
+	assert.Equal(t, uint64(0), c.GetVersion())
+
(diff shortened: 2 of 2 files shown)


pkg/cache/v3/linear.go
```diff
@@ -565,6 +565,10 @@ func (cache *LinearCache) CreateDeltaWatch(request *DeltaRequest, sub Subscripti
 	return cache.trackWatch(watch), nil
 }
 
+func (cache *LinearCache) GetVersion() uint64 {
+	return cache.version
```
valerian-roche, line 569: This cannot be returned without concurrency protection. `version` is only touched under lock today.
Also the version integer is an internal detail. The version from a public API point-of-view is a string with the provided version prefix, not this integer

valerian-roche (changes requested):
Can you provide a rationale to increase the public interface?
I am hoping to fully remove this "version" from the linear cache as it's irrelevant in delta mode and is arbitrary in sotw.

nezdolik:
The need for this change occurred while migrating cilium custom xds control plane to use go control plane in ADS SOTW mode. In cilium control plane we send core Envoy resource and custom resources (like network policy). Since Snapshot cache (which exposes xds resource version) does not support custom resource types (we try to avoid to use TypedExtensionConfig for now), I have to use linear cache for custom resource types (and then combine linear caches with snapshot cache into a mux cache). Cilium control plane registers "await" objects (based on xds node Id, resource version set by cache and type url) and performs certain actions based on when await objects have been acked or if they timed out. In our case we care about resource version which is being set by cache and being sent on the stream it does not need to be user friendly or human readable. Since snapshot cache exposes resource version it would be meaningful to also expose resource version in linear cache to be able to use those cache types in mux cache. String resource version would work as well, we will just parse it on out side to extract the integer value (we care about what is being actually sent on stream).

